PR #3, opened 2015-12-29
Load template from fs if directory exists.

zigomir:
I used this to prepare my custom template https://github.com/zigomir/vue-starter-kit which is fork of official `webpack` with `vuex`.


yyx990803:
:+1:

PR #14, opened 2016-01-13
Use download-git-repo to support bitbucket also

### LINKED ISSUE #9: Ability to download from git url or other sources
Within our company, we would like to fork some of the main templates and make changes to them, but we would like to host it privately in Bitbucket. Our use case is not so much building templates for everyone to use, but building templates for us to use internally that contains our api hooks, layouts etc. Is this something that may be supported?

flipxfx:
Resolves #9

### DIFF
--- a/package.json
+++ b/package.json
@@ -28,7 +28,7 @@
   "dependencies": {
     "chalk": "^1.1.1",
     "commander": "^2.9.0",
-    "download-github-repo": "^0.1.3",
+    "download-git-repo": "^0.0.1",
     "khaos-patched": "^0.9.3",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vue-cli",
-  "version": "1.1.3",
+  "version": "1.2.3",
```
yyx990803: Please do not change the version here, I will do it separately.

flipxfx:
Sorry about that. Removed the version change.

yyx990803:
:+1: thanks!

PR #17, opened 2016-01-14
Add --clone flag to download from private repos

### LINKED ISSUE #16: Ability to download from private repos
I'm submitting the ability to download private repos as a separate issue from #14 because I'm not sure how or if you would like to use it. I made a minor release (non-breaking) change to [flipxfx/download-git-repo](https://github.com/flipxfx/download-git-repo) that allows for private repos.

You can still doing the following for a normal http download:
```javascript
download('owner/repo', 'tmp', function(err) {})
```

But the new version allows you to use [git-clone](https://github.com/jaz303/git-clone) to download both public and private repos.
 ```javascript
download('owner/repo', 'tmp', { clone: true }, function(err) {})
```

The problem is that using `git clone` can be a bit slower (at least for Bitbucket). Due to this I'm not sure if the better way would be to use clone for all template downloads or if there should be an option to specify clone in the cli itself. It's not a ton slower but I didn't want to assume you were ok with the additional time to enable private repo downloads.

These are my stats from the tests in [flipxfx/download-git-repo](https://github.com/flipxfx/download-git-repo) showing http download vs clone.
```
    via github
      ✓ downloads the master branch by default (881ms)
      ✓ download branches too (428ms)
      ✓ downloads from github by default (734ms)
      ✓ clones the master branch by default (592ms)
      ✓ clones branches too (651ms)
    via bitbucket
      ✓ downloads the master branch by default (189ms)
      ✓ download branches too (175ms)
      ✓ clones the master branch by default (1370ms)
      ✓ clones branches too (2040ms)
```

flipxfx:
Resolves #16.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,6 +49,8 @@ Where `username/repo` is the GitHub repo shorthand for your fork.
 
 The shorthand repo notation is passed to [download-git-repo](https://github.com/flipxfx/download-git-repo) so you can also use things like `bitbucket:username/repo` for a Bitbucket repo and `username/repo#branch` for tags or branches.
 
+If you would like to download from a private repository use the `--clone` flag and the cli will use `git clone` so your SHH keys are used.
+
 You can also create your own template from scratch:
 
 - A template repo **must** have a `template` directory that holds the template files.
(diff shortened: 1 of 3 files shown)


yyx990803:
:+1:

ishowman:
@flipxfx `vue init <owner/repo project> <project-name> -c` is only for private repos, it can not work well for public repos?I use the command below, and there're some errors happened...
image

flipxfx:
The clone option is needed for private repos and optional for public.

PR #18, opened 2016-01-16
Only prompt for schema found in the template

flipxfx:
When prompting the user for template options, I think it makes sense only to prompt for answers needed.  So if a template doesn't have `{{ author }}` anywhere, then they will not be bothered with the prompt as it may cause confusion. And the same goes for options that appear in the `meta.json` but not in the template.


yyx990803:
In the future versions we are going to use `schema` in `meta.json` as the only data source for prompt - so if you don't want a prompt to show up, simply delete it from your schema.

PR #65, opened 2016-03-30
Use inquirer instead of prompt-for

### LINKED ISSUE #61: Use inquirer instead of forked prompt-for
`prompt-for` is no longer maintained. Using a custom fork doesn't seem like a good idea.

[inquirer](https://github.com/SBoudrias/Inquirer.js/) looks much better and offers multiple choices which should solve #5 as well.

zigomir:
Fixes #61.

I removed test from scripts because it seemed redundant. Will fix it if I'm wrong.


zigomir:
@yyx990803 any feedback for the PR? I started it because #61 has `help wanted` label.

yyx990803:
Sorry, the issues are a bit misleading. The two issues actually need to be worked on together. `khaos-forked` is using `prompt-for-forked` under the hood, in a rewrite we would get rid of both and directly use `metalsmith` + `inquirer`.

zigomir:
Thanks, I'll see what I can do.

zigomir:
Closing in favor of #67.

Issue #62, opened 2016-03-23
Use raw metalsmith instead of khaos


yyx990803:
Khaos is no longer maintained, but it is based on [metalsmith](https://github.com/metalsmith/metalsmith) which seems to still be maintained.

Using metalsmith directly also allows us to implement custom logic such as conditional files based on prompt results.

PR #67, opened 2016-04-02
Metalsmith and inquirer

### LINKED ISSUE #61: Use inquirer instead of forked prompt-for

zigomir:
Fixes #61 and #62.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,7 +59,7 @@ You can also create your own template from scratch:
 
 - All template files will be piped through Handlebars for simple templating - `vue-cli` will automatically infer the prompts based on `{{}}` interpolations found in the files.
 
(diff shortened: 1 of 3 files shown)


yyx990803:
Wow nice job! Thanks a lot. One small thing I ran into is that it's attempting to render binaries that are not supposed to be parsable by handlebars. Solved by simply adding a mustache presence check.

zigomir:
No problem, thanks for finishing. We can now also close #62.

PR #71, opened 2016-04-05
Tests

### LINKED ISSUE #69: Tests
Basic CLI tests confirming prompts and file generation are working correctly:

```
- generate correct file with text prompt
- generate correct file with boolean prompt
- generate correct file with multi choice
- copy binary files without error
```

zigomir:
Fixes #69

I'm not really sure about failing example for binary file skip. Test `avoid rendering files that do not have mustaches` works even if I comment out regex for testing mustache braces.

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,9 @@
+{
+  "env": {
+    "node": true,
+    "mocha": true,
+    "development": {
+      "presets": ["es2015", "stage-2"]
+    }
+  }
+}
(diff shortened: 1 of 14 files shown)


yyx990803:
Awesome, I like the refactoring... one thing is that `mocha` and `chai` are missing from package.json.

zigomir:
How come I left those out :blush:
Adding right away.

yyx990803:
👍

PR #73, opened 2016-04-08
Add validation for package name in vue init	

watilde:
Hi :wave: 

I try to add a `validation` to check package name is [invalid or not](https://docs.npmjs.com/files/package.json#name) in `vue init`.
screenschot

### DIFF
--- a/lib/ask.js
+++ b/lib/ask.js
@@ -41,7 +41,8 @@ function prompt (data, key, prompt, done) {
     name: key,
     message: prompt.message || prompt.label || key,
     default: prompt.default,
-    choices: prompt.choices || []
+    choices: prompt.choices || [],
+    validate: prompt.validate || function () { return true }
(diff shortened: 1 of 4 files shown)


zigomir:
@watilde it would be really nice if you can also provide a test case.
I'm also a bit confused by the image you attached. Why is `Project name (test)` (guessing here the `test` is a project name) not valid?

watilde:
@zigomir Thanks for your comment! I've rebased and added a test case.
I'm sorry for making you confused by a screenshot I attached. Actually this gif image is what exactly I did in the screenshot. Animation can explain it clearly :smiley: 
gif

zigomir:
@watilde nice! Thanks for the update. This looks good to me. @yyx990803 do you agree as well?

watilde:
@yyx990803 @zigomir Thanks!

PR #76, opened 2016-04-11
Support license input with validation

watilde:
Hello :wave:

I supported `license` input in `vue init`, with validation.
gif


yyx990803:
Thanks for the PR, but I don't think we should list license as an option for the CLI... people will likely be using this for private projects where license doesn't make sense.

watilde:
Thanks for your comment! Yeah I should did it. I will care it next time :smiley:

PR #87, opened 2016-04-27
Wrong directory name

### LINKED ISSUE #86: the directory is wrong
image

zigomir:
Fixes #86.

Although I was thinking we might rather do this

```js
var data = extend({}, metalsmith.metadata(), { destDirName: name })
```

and also update https://github.com/vuejs-templates/webpack/blob/dist/meta.json to use `destDirName` inside `completeMessage`.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,5 +1,9 @@
 {
   "extends": "standard",
+  "env": {
+    "mocha": true,
+    "node": true
+  },
   "rules": {
     "arrow-parens": [2, "as-needed"]
   }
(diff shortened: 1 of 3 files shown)


yyx990803:
That makes sense - I've invited you to the templates organization so you can change both places at the same time. LGTM.

zigomir:
@yyx990803 All right, I renamed it here and [in template](https://github.com/vuejs-templates/webpack/blob/dist/meta.json). Now we should merge this and publish new version imo.

yyx990803:
published 2.0.3.

Issue #93, opened 2016-05-08
register handlebars helper


queckezz:
Is there a possibility to register custom handlebars helper to `vue-cli`? Specifically I need the a helper for camelcasing names automatically. 

An use case is for installation usage in the readme as follows:
```js
const {{camelcase name}} = require('{{name}}')
```

Khaos has this [built into core](https://github.com/segmentio/khaos/blob/master/lib/helpers.js#L9-L13), not sure if you want to include it too.

yyx990803:
Sounds useful - what we can do is [in addition to `meta.json`, also accept `meta.js`](https://github.com/vuejs/vue-cli/blob/master/lib/options.js). Then you can do something like:

``` js
// meta.js
module.exports = {
  prompts: { ... },
  helpers: {
    // custom helpers
  }
}
```

PR #95, opened 2016-05-08
register handlebars helpers

queckezz:
I took a stab at #93. This PR allows a `meta.js` to be defined as a metadata file and additionally allows you to define handlebars helpers like this:

```js
// meta.js
module.exports = {
  helpers: {
    lowercase: str => str.toLowerCase()
  }
}
```

I also tried to document `meta.js` but i'm not sure if this is the best way. Let me know!

### DIFF
--- a/README.md
+++ b/README.md
@@ -68,7 +68,7 @@ vue init ~/fs/path/to-custom-template my-project
 
 - A template repo **must** have a `template` directory that holds the template files.
 
-- A template repo **may** have a metadata file for the template which can be both a `meta.js` or a `meta.json` file. It can contain the following fields:
+- A template repo **may** have a metadata file for the template which can be either a `meta.js` or `meta.json` file. It can contain the following fields:
 
   - `prompts`: used to collect user options data;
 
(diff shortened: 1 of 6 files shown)


README.md
```diff
@@ -68,7 +68,7 @@ vue init ~/fs/path/to-custom-template my-project
 
 - A template repo **must** have a `template` directory that holds the template files.
 
-- A template repo **may** have a `meta.json` file that provides metadata for the template. The `meta.json` can contain the following fields:
+- A template repo **may** have a metadata file for the template which can be both a `meta.js` or a `meta.json` file. It can contain the following fields:
```
yyx990803: perhaps should use "which can be **either** a ... or ..."

README.md
```diff
@@ -129,6 +129,24 @@ Two commonly used Handlebars helpers, `if_eq` and `unless_eq` are pre-registered
...
+
+Upon registration, they can be used as follows:
+
+``` handlebars
+{{ lowercase name }}
+```
+
 #### File filters
 
 The `filters` field in `meta.json` should be an object hash containing file filtering rules. For each entry, the key is a [minimatch glob pattern](https://github.com/isaacs/minimatch) and the value is a JavaScript expression evaluated in the context of prompt answers data. Example:
```
yyx990803: should also update `meta.json` here to be more generic?

yyx990803:
Nice work! Just a few small doc tweaks - LGTM otherwise.

queckezz:
Alright, fixed :)

PR #104, opened 2016-05-24
cli: fix local template path for windows

queckezz:
This PR now additionally allows windows like paths with `\`s

```bash
# both works now
vue init D:\dev\create-module test-project
vue init ~/dev/create-module test-project
```

### DIFF
--- a/bin/vue-init
+++ b/bin/vue-init
@@ -7,6 +7,7 @@ var path = require('path')
 var rm = require('rimraf').sync
 var uid = require('uid')
 var ora = require('ora')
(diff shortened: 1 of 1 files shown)


bin/vue-init
```diff
@@ -158,3 +158,7 @@ function downloadAndGenerate (template) {
     })
   })
 }
+
+function isLocal (str) {
```
zigomir: `isLocal` implies more than just that str is string and not an url. In my opinion `exist` should also be called here.
+ you can easily add test for it ;)

PR #120, opened 2016-07-12
Windows fix for local templates.

zigomir:
No need to check for forward slash when checking if path is local.

Fixes #104.


zigomir:
@yyx990803 I think this one is ready to merged. I tested it on Windows 8 VM.

PR #125, opened 2016-07-19
Add an `isInPlace` data to metadata `completeMessage` property

hugohil:
We test if a destination path is provided. If not, then we are scaffolding into the current directory and pass the `isCurrentDir` to `true`, if not, we set it to `false`.


zigomir:
Hi @hugohil!

Sorry for not responding for so long here. As you can see, at this moment your PR has some conflicts with master branch.

My general concern is if we really need another argument just to get a slightly better `completeMessage`. Also, there is no test that would clearly demonstrate what this change is really doing.

What do you think?

hugohil:
@zigomir Thanks for the feedback !

I resolve conflicts and also avoided using this option as a new parameter for the `generate` function.

For the tests, I just spend a some time trying to get one working without any luck (changing the cwd and still dealing with async got a bit tricky). I'll give it another shot later.

bin/vue-init
```diff
@@ -101,11 +101,7 @@ function run () {
       if (!hasSlash) {
         // use official templates
         template = 'vuejs-templates/' + template
-        if (template.indexOf('#') !== -1) {
-          downloadAndGenerate(template)
-        } else {
-          checkDistBranch(template, downloadAndGenerate)
-        }
+        checkDistBranch(template, downloadAndGenerate)
```
zigomir: I think there should be no changes here in `vue-init` file. We shouldn't call `checkDistBranch` for every case because with that user can't specify the branch from which she/he wants to download template for.

bin/vue-init
```diff
@@ -90,7 +90,7 @@ if (exists(to)) {
 
 function run () {
   // check if template is local
-  if (exists(template)) {
+  if (hasSlash && exists(template)) {
```
zigomir: `hasSlash` was removed from master to support Windows.

zigomir:
@hugohil thanks! This looks better now. I left you few comments on `vue-init` file changes.

test/e2e/test.js
```diff
@@ -40,7 +40,8 @@ describe('vue-cli', () => {
       sass: true
     },
     pick: 'no',
-    noEscape: true
+    noEscape: true,
+    isInPlace: false
```
zigomir: `isInPlace` isn't used anywhere else.

test/e2e/mock-template-repo/template/package.json
```diff
@@ -1,13 +1,9 @@
...
+  "name": "vue-cli-test",
+  "description": "vue-cli e2e test",
+  "author": "John Doe <john@doe.com>",
   "devDependencies": {
-    {{#preprocessor.less}}
     "less": "^2.6.1",
-    {{/preprocessor.less}}
-    {{#preprocessor.sass}}
     "node-sass": "^3.4.2"
-    {{/preprocessor.sass}}
```
zigomir: I think we shouldn't change this file for this feature.

bin/vue-init
```diff
@@ -107,7 +107,11 @@ function run () {
           checkDistBranch(template, downloadAndGenerate)
         }
       } else {
-        downloadAndGenerate(template)
+        if (template.indexOf('#') !== -1) {
+          downloadAndGenerate(template)
+        } else {
+          checkDistBranch(template, downloadAndGenerate)
+        }
```
zigomir: I think `vue-init` file shouldn't be changed for this feature. Please merge master into your dev branch and make sure `vue-init` isn't changed.

zigomir: I think `vue-init` file shouldn't be changed for this feature. Please merge master into your dev branch and make sure `vue-init` isn't changed.

lib/generate.js
```diff
@@ -34,7 +34,8 @@ module.exports = function generate (name, src, dest, done) {
   var metalsmith = Metalsmith(path.join(src, 'template'))
   var data = Object.assign(metalsmith.metadata(), {
     destDirName: name,
-    noEscape: true
+    noEscape: true,
+    isCurrentDirectory: (dest === '.' || !dest)
```
zigomir: Does this really work?  I think `dest` here won't be just `.` but whole path.

zigomir:
@hugohil I left you few more comments...but then just went ahead and did it on my own in https://github.com/vuejs/vue-cli/pull/156. I hope you don't mind.

Issue #130, opened 2016-07-26
`vue list` throw error when github api response message


guox191:
`$ vue list`

Cli will throw error when github API rate limit exceeded, the response just like :
```
{
  "message": "API rate limit exceeded for <your IP address>. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)",
  "documentation_url": "https://developer.github.com/v3/#rate-limiting"
}
```
But code expects an array, so get problem:
```
JSON.parse(body).forEach(function (repo) {

TypeError: JSON.parse(...).forEach is not a function
```

LinusBorg:
This should be merged imho

PR #131, opened 2016-07-26
Compatible with github API error message

guox191:
To solve this issue #130

### DIFF
--- a/bin/vue-list
+++ b/bin/vue-list
@@ -35,9 +35,6 @@ request({
         ' - ' + repo.description)
     })
   } else {
-    console.log('  Unavailable official templates:')
-    console.log()
-    console.error('  ' + requestBody.message)
+    console.error(requestBody.message)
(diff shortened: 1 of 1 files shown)


bin/vue-list
```diff
@@ -24,12 +24,20 @@ request({
...
+    console.log('  Available official templates:')
+    console.log()
+    requestBody.forEach(function (repo) {
+      console.log(
+        '  ' + chalk.yellow('★') +
+        '  ' + chalk.blue(repo.name) +
+        ' - ' + repo.description)
+    })
+  } else {
+    console.log('  Unavailable official templates:')
```
yyx990803: We can just remove this line...

guox191: Thanks for your suggestion 😀

LinusBorg:
Great. Should be merged IMHO.

PR #150, opened 2016-08-21
Fix 'Cannot find module' error of meta.js when using local template

### LINKED ISSUE #140: 'Cannot find module' error of meta.js when using local template
I can't seem to use `meta.js` instead of `meta.json`. I renamed the default webpack meta.json to meta.js and put a `module.exports = ` at the start but it's telling me:

**Error: Cannot find module 'vue-webpack-template/meta.js'**

If I rename the file to meta.json and remove the `module.exports =` it works as expected. Am I doing something wrong?

ktsn:
Possibly fix #140

The error happens when we use `meta.js` for a local template because the relative path for `require(...)` is relative to `__dirname`, not `process.cwd()`

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -40,7 +40,7 @@ function getMetadata (dir) {
   if (exists(json)) {
     opts = metadata.sync(json)
   } else if (exists(js)) {
-    var req = require(js)
+    var req = require(path.resolve(js))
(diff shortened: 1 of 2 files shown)


zigomir:
Thanks @ktsn!

I tried this out and it helps to solve #140. I could simulate it on my machine.

PR #159, opened 2016-08-31
I went looking for the "run dev command twice"...

jppope:
I went looking for the "run dev command twice" so I'm guessing there are other people that were wondering about the full usage. Hopefully it helps the documentation. cheers

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,10 +26,15 @@ The above command pulls the template from [vuejs-templates/webpack](https://gith
 
 Full Example:
 ``` npm install -g vue-cli
+
 vue init webpack my-project
+
 <answer prompts>
+
 cd my-project
+
 npm install
+
 npm run dev ```
 
 
--- a/README.md
+++ b/README.md
@@ -25,12 +25,12 @@ $ vue init webpack my-project
 The above command pulls the template from [vuejs-templates/webpack](https://github.com/vuejs-templates/webpack), prompts for some information, and generates the project at `./my-project/`.
 
 Full Example:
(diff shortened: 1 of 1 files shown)


zigomir:
Hi @jppope. Can you please describe better what was your pain? In more of a step by step fashion so I can see it myself.

Each official template has [usage](https://github.com/vuejs-templates/webpack#usage) section in the readme file. I know this might be a pain because it's not sitting right here.

Maybe we need to document better or more clearly that `vue-cli` is a tool that only downloads a template and from there on it's only template's own job to tell you how to use it.

posva:
When you install a project, the cli tells you what to do. How did you miss that?

screen shot 2016-09-01 at 09 41 31

It doesn't cost much adding it to the example in the readme though

zigomir:
@posva Maybe we need to be more explicit about what exactly `vue-cli` job is. Where it starts and where it ends.

But, since most of users have problems with official templates (of course) maybe it wouldn't be such a bad idea to just pull them in under `vue-cli` repo under `official-templates` directory or something. But now it's a bit late for this and would probably made even more confusion...(just thinking out loud here).

posva:
I think the way templates are fetched is nice.
It may be useful to have a way for each template to display commands. Something like a text file that is printed by vue-cil after the `vue init` command is finished. It may not be enough

Anyways, still waiting for @jppope answer to understand where we failed

zigomir:
Closing due to inactivity, please open new issue if you can provide more accurate info about this issue.

PR #162, opened 2016-09-05
Warn user about Vue 2.x template.

zigomir:
Don't yet warn for `webpack` template since 2.0 version is not yet released.

warn for vue 2 version

/cc @chrisvfritz

### DIFF
--- a/bin/vue-init
+++ b/bin/vue-init
@@ -12,7 +12,7 @@ var inquirer = require('inquirer')
 var request = require('request')
 var logger = require('../lib/logger')
 var generate = require('../lib/generate')
-var checkVersion = require('../lib/check-version')
+var { checkVersion, warnForVue2Version } = require('../lib/check-version')
 
 /**
  * Usage.
@@ -100,11 +100,15 @@ function run () {
     checkVersion(function () {
       if (!hasSlash) {
         // use official templates
-        template = 'vuejs-templates/' + template
+        var officialTemplate = 'vuejs-templates/' + template
         if (template.indexOf('#') !== -1) {
-          downloadAndGenerate(template)
+          downloadAndGenerate(officialTemplate)
         } else {
-          checkDistBranch(template, downloadAndGenerate)
+          // until official webpack template for Vue 2.0 is released in master/dist branch
+          if (template !== 'webpack') {
+            warnForVue2Version(template, name)
+          }
+          checkDistBranch(officialTemplate, downloadAndGenerate)
         }
       } else {
         downloadAndGenerate(template)
--- a/bin/vue-init
+++ b/bin/vue-init
@@ -107,7 +107,7 @@ function run () {
         } else {
           // until official webpack template for Vue 2.0 is released in master/dist branch
           if (template !== 'webpack') {
-            vue2VersionWarn(template, name)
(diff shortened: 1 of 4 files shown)


bin/vue-init
```diff
@@ -12,7 +12,7 @@ var inquirer = require('inquirer')
 var request = require('request')
 var logger = require('../lib/logger')
 var generate = require('../lib/generate')
-var checkVersion = require('../lib/check-version')
+var { checkVersion, warnForVue2Version } = require('../lib/check-version')
```
chrisvfritz: I love destructuring too, 😃  but it's unfortunately [only available in Node 6+](http://node.green/#destructuring--declarations) and I believe we're supporting down to v4.

bin/vue-init
```diff
@@ -100,11 +100,15 @@ function run () {
     checkVersion(function () {
       if (!hasSlash) {
         // use official templates
-        template = 'vuejs-templates/' + template
+        var officialTemplate = 'vuejs-templates/' + template
```
chrisvfritz: I like the addition of `officialTemplate`! Makes the code a bit clearer and keeps the value of `template` more predictable as exactly what the user entered. 👍

lib/check-version.js
```diff
@@ -36,3 +36,17 @@ module.exports = function (done) {
...
+  console.log(chalk.red('  This will install Vue 2.x version of template.'))
+  console.log()
+  console.log(chalk.yellow('  For Vue 1.x use: ') + chalk.green(vue1InitCommand))
+  console.log()
+}
+
+module.exports = {
+  checkVersion: checkVersion,
+  warnForVue2Version: warnForVue2Version
+}
```
chrisvfritz: Wondering if it would be better to just create a new file for `warnForVue2Version`. It doesn't have anything to do with checking the node version, so having it in here might make it difficult to find in a quick search. What do you think?

lib/check-version.js
```diff
@@ -36,3 +36,17 @@ module.exports = function (done) {
     done()
   })
 }
+
+function warnForVue2Version (template, name) {
+  var vue1InitCommand = 'vue init ' + template + '#1.0' + ' ' + name
```
chrisvfritz: Should we also pass in `inPlace` and then only show the `name` in the command if it's false?

lib/vue2-version-warn.js
```diff
@@ -0,0 +1,10 @@
+var chalk = require('chalk')
+
+module.exports = function (template, name) {
+  var vue1InitCommand = 'vue init ' + template + '#1.0' + ' ' + name
```
chrisvfritz: Should we also pass in `inPlace` and only add the `name` to this string if it's false?

zigomir: I'm thinking doing this on call site:

```js
vue2VersionWarn(template, inPlace ? '' : name)
```

chrisvfritz: That works. 👍

chrisvfritz:
@yyx990803 We're getting a fair amount of confusion with people not realising they're using Vue 2.0 by default in most templates now, so @zigomir came up with a fix. If possible, I think getting this out ASAP would really help.

PR #164, opened 2016-09-07
Adds detail to the readme on installing a specific template version

chris-schmitz:
I had to do some hunting through packages and source code to figure out how to pull down the older 1.0 template. After figuring it out I thought it would be helpful to add the details to the readme. 

Merge if you see fit and thank you for creating this tool :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -181,6 +181,25 @@ Note that the `dot` option for minimatch is set to `true` so glob patterns would
 }
 ```
 
+### Installing a specific template version
+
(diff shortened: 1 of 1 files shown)


zigomir:
Thanks @chris-schmitz!

We now have [a warning that tells you're about to install a Vue 2.x template](https://github.com/vuejs/vue-cli/pull/162). This is already published, so we expect a bit less confusion about this.

But your addition to the documentation is still beneficial and adds clarity.

PR #176, opened 2016-09-21
Add -2.0 templates deprecation warning.

### LINKED ISSUE #171: Strategy for transitioning away from the -2.0 templates
@zigomir Before we delete [the `-2.0` template repos](https://github.com/vuejs-templates?utf8=%E2%9C%93&query=2.0), I think it'd be good to have a PR ready to either:

1. Alert people that they no longer exist, because the main repos serve 2.0, _or_
2. Just automatically strip `-2.0` from the user input for official template installs so that it Just Works

Not sure which is the best option. What do you think?

zigomir:
Fixes #171


zigomir:
@chrisvfritz please take a look.

lib/warnings.js
```diff
@@ -0,0 +1,20 @@
+var chalk = require('chalk')
+
+module.exports = {
+  deprecation: function (template, name) {
+    var initCommand = 'vue init ' + template.replace('-2.0', '') + ' ' + name
+
+    console.log(chalk.red('  This template is deprecated.'))
```
chrisvfritz: Perhaps it would be good to add a bit more information here. For example:

> This template is deprecated, as the original template now uses Vue 2.0 by default.

Just so that they know the new command will still use 2.0.

lib/warnings.js
```diff
@@ -0,0 +1,20 @@
+var chalk = require('chalk')
+
+module.exports = {
+  deprecation: function (template, name) {
```
chrisvfritz: Since we might want to have other kinds of deprecations in the future, maybe we can rename this warning to `v2SuffixTemplatesDeprecated`. I know it's a bit long, but I think that's fine since we're not using it constantly. It's also much clearer what's being deprecated.

lib/warnings.js
```diff
@@ -0,0 +1,20 @@
...
+    var initCommand = 'vue init ' + template.replace('-2.0', '') + ' ' + name
+
+    console.log(chalk.red('  This template is deprecated.'))
+    console.log()
+    console.log(chalk.yellow('  Please use this command instead: ') + chalk.green(initCommand))
+    console.log()
+  },
+  vue2Version: function (template, name) {
```
chrisvfritz: While we're here, maybe rename this to a more descriptive `v2BranchIsNowDefault`? What do you think?

PR #177, opened 2016-09-23
Download the template repo into system default temp dir, to avoid the…

tweea:
Download the template repo into system default temp dir, to avoid the unnecessary tmp folder auto-creation on Windows.


zigomir:
Thank you @tweea ✌️

PR #180, opened 2016-09-28
Remove checkDistBranch

zigomir:
also `webpack` template now supports Vue 2.0.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "vue-cli",
-  "version": "2.3.2",
+  "version": "2.3.3",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "vue-cli",
-  "version": "2.3.2",
+  "version": "2.3.3",
```
chrisvfritz: I would update to `2.4.0` instead of `2.3.3`, since we're [adding functionality in a backwards-compatible manner](http://semver.org/).

zigomir: You're totally right. I changed it.

Issue #185, opened 2016-10-01
Short version tag not working


Sarke:
# vue -h

	  Usage: vue <command> [options]

	  Commands:

		init        generate a new project from a template
		list        list available official templates
		help [cmd]  display help for [cmd]

	  Options:

		-h, --help     output usage information
		-V, --version  output the version number

	# vue -v

	  error: unknown option `-v'

	# vue --version
	2.1.1

egoist:
It's the uppercase `-V`

Sarke:
Ah... why?

egoist:
I also think we should use `-v` to be consistent with other cli tools.

We can omit the `program.version(pkg.version)` and use `program.option(-v, --version)` as a workaround.

posva:
`-v` is usually reserved for verbose. It's actually quite common to have `-V` for the version

PR #186, opened 2016-10-01
Use -v instead of -V

egoist:
See https://github.com/vuejs/vue-cli/issues/185


posva:
I prefer having the version on `-V`, people are more used by it

egoist:
ok, it's just a matter of taste, many tools use -v for version and -V for verbose too, I can totally accept both 😜

egoist:
``

PR #194, opened 2016-10-12
Better local template handling with Windows compatability

presidenten:
Windows users have trouble using local templates when the the previous solution only checked for slash and not backslash or `path.sep`. With this new approach mac, linux and windows users can also use relative paths to the template:

``` shell
   vue init ./myTemplate projectName
```

or

``` shell
   vue init ../myTemplatesFolder/myTemplate projectName
```

Tested on Mac osx with node 6.2.2 both from local build and globally installed build, as well as Windows 8 with node 6.7.0 from local build.

### DIFF
--- a/bin/vue-init
+++ b/bin/vue-init
@@ -64,7 +64,7 @@ process.on('exit', function () {
 var template = program.args[0]
 var hasSlash = template.indexOf('/') > -1
 var rawName = program.args[1]
-var templatePath = path.normalize(path.join(process.cwd(), template))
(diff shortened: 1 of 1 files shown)


presidenten:
Please wait with the merge, I broke the absolute path local init. Ill fix it and send a new pull request.

presidenten:
There we go! 
Now it also works with absolute paths. Sorry for the trouble.

It has now been tested on mac, linux (ubuntu), and windows.

zigomir:
Thanks for working on this @presidenten, looks good to me.

Issue #230, opened 2016-11-12
`npm install` after `vue init` fails if git user name has double quotes


sayanriju:
```
$ lsb_release -a
No LSB modules are available.
Distributor ID: Ubuntu
Description:    Ubuntu 16.04 LTS
Release:        16.04
Codename:       xenial

$ node -v
v6.4.0

$ npm -v
3.10.3

$ vue --version
2.5.0
```
Command: `vue init webpack-simple` accepting the default options followed by `npm i` in the project directory.

Result: Fails with 
```
npm ERR! file /tmp/foo/package.json
npm ERR! code EJSONPARSE

npm ERR! Failed to parse json
npm ERR! Unexpected token 'i' at 4:22
npm ERR!   "author": "Sayan "Riju" Chakrabarti <my.actual.email@gmail.com>",
npm ERR!                      ^
npm ERR! File: /tmp/foo/package.json
npm ERR! Failed to parse package.json data.
npm ERR! package.json must be actual JSON, not just JavaScript.
```

PR #232, opened 2016-11-12
Escape string in git config user.name

FadyMak:
Issue #230

### DIFF
--- a/lib/git-user.js
+++ b/lib/git-user.js
@@ -9,7 +9,7 @@ module.exports = function () {
     email = exec('git config --get user.email')
   } catch (e) {}
 
-  name = name && name.toString().trim()
+  name = name && JSON.stringify(name.toString().trim()).slice(1, -1)
   email = email && (' <' + email.toString().trim() + '>')
   return (name || '') + (email || '')
 }


lib/git-user.js
```diff
@@ -9,7 +9,7 @@ module.exports = function () {
     email = exec('git config --get user.email')
   } catch (e) {}
 
-  name = name && name.toString().trim()
+  name = name && JSON.stringify(name.toString().trim()).slice(1, -1)
```
zigomir: How come `.slice` is needed here?

sayanriju: The extra quotes seems to be a quirk on the node REPL. It should work fine if you are running it in a script.

FadyMak: @sayanriju `JSON.stringify('hello')` will result in an additional set of quotes around the output:

```javascript
console.log(JSON.stringify('hey'))
// "hey"
```

i.e.: the resultant string is actually `"hey"` as opposed to just `hey`

In our case, we don't want the resultant Git `user.name` to be wrapped in quotes but we would like to escape them:

```javascript
var gitConfigName = 'hello "me" you'
var packageJSON = { author: JSON.stringify('hello "me" you').slice(1, -1) }
console.log(packageJSON)
// { author: 'hello \\"me\\" you' }
packageJSON = { author: JSON.stringify('hello "me" you') }
console.log(packageJSON)
// { author: '"hello \\"me\\" you"' }
```

I hope that helps clear it up a bit.

sayanriju: My bad! I mean slash, not quote (The extra ones in \\"me\\"). :P
Anyway, it did clear things up.

BTW, I guess `JSON.stringify` is preferred over `String.replace` due to the slight performance hit?

zigomir:
Thanks @FadyMak!

PR #259, opened 2016-12-07
Minor improvements for 2.x warning

phanan:
There are two problems (IMHO at least) with the current 2.x "warning":

* It's not an error per se, more like an information, so I changed the color to green instead of red.
* Grammar.

Since both happen in one line, I decided to merge them into one commit.


zigomir:
I agree as well. I'm merging because we're already 3 :)

PR #266, opened 2016-12-17
add bash to terminal commands

ChangJoo-Park:


posva:
Thanks for the PR but this doesn't change anything on the look. The `$` at the beginning is meant to show that visually, and semantically

PR #276, opened 2016-12-26
add the package name custom validate to take effect

likun7981:
I see the #73 ,The add the validation for package name , it is really nice , but it will be overwritten the user custom validate function , so , I fix it.

### DIFF
--- a/lib/options.js
+++ b/lib/options.js
@@ -75,12 +75,15 @@ function setDefault (opts, key, val) {
 }
 
 function setValidateName (opts) {
-  opts.prompts.name.validate = function (name) {
+  var name = opts.prompts.name
+  var customValidate = name.validate
+  name.validate = function (name) {
     var its = validateName(name)
     if (!its.validForNewPackages) {
       var errors = (its.errors || []).concat(its.warnings || [])
       return 'Sorry, ' + errors.join(' and ') + '.'
     }
+    if (typeof customValidate === 'function') return customValidate(name)
     return true
   }
 }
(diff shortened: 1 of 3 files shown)


kazupon:
Thank you for you PR!
However, I could not understand overwritten the user custom validate function.
Can you tell us about the usecase please?

likun7981:
in the meta.js  , if user define a validate option in the name prompt , the validate option will be overitten , it will not work。 
if I config a meta.js the prompts like this:
```javascript
prompts: {
    name: {
      type: 'string',
      required: true,
      label: 'Project name',
      validate: function (input) {
        return input === 'custom' ? 'can not input `custom`' : true
      }
    }
 }
```
 #73 ,  Add a package name validate,  like this:
```javascript
function setValidateName (opts) {
  opts.prompts.name.validate = function (name) {
    var its = validateName(name)
    if (!its.validForNewPackages) {
      var errors = (its.errors || []).concat(its.warnings || [])
      return 'Sorry, ' + errors.join(' and ') + '.'
    }
    return true
  }
}
```
so, the validate function I defined in meta.js  will be overwritten by it

kazupon:
Thanks!
However, I think this essential issue is overwritten the user.
we should be fixed, or warned it.
for example, in [options function](https://github.com/vuejs/vue-cli/blob/master/lib/options.js#L15-L18), check the whether exisit `name`, if exist it, ignore or warn.

likun7981:
Sometimes , we need a unified format for name ,example vue-router, vue-resource and so on. I think you should keep it

likun7981:
Maybe you can give a configuration item, it's value is a String or RegExp.

kazupon:
>we need a unified format for name ,example vue-router, vue-resource and so on.

Indeed, I can now see your use-cases.indeed, I could now see your use-cases.
In that case, Looks good to me.

PR #287, opened 2017-01-07
Add `vue build` command, closed #173

### LINKED ISSUE #173: New command: `vue build`
Ref: https://github.com/egoist/vbuild/issues/9#issuecomment-247859666

@egoist is the author of `vbuild` which is a zero-config tool for prototyping Vue projects. We'd like to fold that functionality into `vue-cli` as `vue build` with a sharper focus on prototyping experience.
### Summary
- if a user wants to build a production app, use `vue init`;
- if a user wants to quickly prototype an idea, use `vue build --dev` (and in most cases can later just copy the `src` files into a production scaffold generated with `vue init`)
### Technical Notes (open to discussion)
- Should support directly building a `*.vue` file, e.g. `vue build App.vue` and BOOM. Great for single-component prototyping. (Maybe even incorporate something like [react-storybook](https://github.com/kadirahq/react-storybook) if running in single-component mode)
- We want this to be clearly focused on prototyping, so the default setup should be minimal and fast. Probably better to use a lighter weight ES2015 transpiler e.g. `buble`.
- Users will still inevitably have their preferences, even for prototyping. But maybe we can make this config global, e.g. have one single `vue.config.js` in the user's home directory, which `vue build` will then use for all prototypes - configure once, run everywhere.
- Is there any smart way to transition a project using `vue build` to a proper scaffolded project created from `vue init`?
### Nice to Have
- auto-infer the npm dependencies in the files and `npm install` them automatically

egoist:
As discussed in https://github.com/vuejs/vue-cli/issues/173

vue-cli

## Summary

- This PR introduces one command `vue build` and three options `--port` `--prod/production` ~~`--local`~~ `--config`
- Outputs to `./dist` by default
- Using buble-loader
- Respect user config, you can populate them at `~/.vue/config.js` (cli options) and `~/.vue/webpack.config.js` (webpack config to merge)

Check out [doc](https://github.com/EGOIST-robot/vue-cli/blob/vue-build/docs/build.md) for more.

I choose `~/.vue` directory instead of file `~/vue.config.js`, in a directory you can install loaders and modules there, so that you don't need to install them again in your project

## ~~What's not ready~~

No `auto install`, it could be buggy IMO

- [x] Still need to add `vue build Counter.vue` magic
- [x] Still need to add some custom css preprocessors for SFC, so that user only need to install relevant loader
- [x] Extract CSS
- [x] Allow to have project-specific config at project directory, use `--config` and `--webpack` to switch, for example `--config config.js`, and `--webpack webpack.config.js`
- [x] Tests

What do you think?


egoist:
cc @kazupon @yyx990803 @posva @zigomir

bin/vue-build
```diff
@@ -0,0 +1,243 @@
...
+  webpackConfig.devtool = 'eval-source-map'
+  webpackConfig.entry.client.push(
+    require.resolve('webpack-dev-server/client'),
+    require.resolve('webpack/hot/dev-server')
+  )
+  webpackConfig.plugins.push(
+    new webpack.HotModuleReplacementPlugin(),
+    new FriendlyErrorsPlugin(),
+    new PostCompilePlugin(() => {
+      console.log(`> Open http://localhost:${options.port}`)
```
kazupon: after listen, to open(e.g. [opn](https://github.com/sindresorhus/opn)) with default browser is useful. 😉

docs/build.md
```diff
@@ -0,0 +1,133 @@
+# vue build
+
+`vue build` command gives you a zero-configuration development setup, install once and build everywhere.
+
+## Features
+
+- **Not a boilerplate**: run a single command to development your app
```
posva: did you mean to develop, to build/bundle?

docs/build.md
```diff
@@ -0,0 +1,133 @@
...
+- **Out of the box**: ES2015, single-file component with hot reloading and custom CSS preprocessors
+- **Customizable**: populate a `~/.vue/webpack.config.js` for custom webpack config
+- **Single-file component mode**: just run `vue build Component.vue` and boom!
```
posva: I think the `boom` should be a bit more explicit about what we get 😆

docs/build.md
```diff
@@ -0,0 +1,133 @@
...
+
+```bash
+$ vue build -h
+```
+
+## Configuration files
+
+By default, we use `~/.vue/config.js` and `~/.vue/webpack.config.js` if they exist. You can use `--config [dir]` to set a custom config directory which is relative to `process.cwd()` instead of using `.vue` directory.
+
+To disable config files, simply `--no-config`.
```
posva: you're missing a verb between simply and the option. Something like add or do

docs/build.md
```diff
@@ -0,0 +1,133 @@
...
+
+### config.js
+
+CLI options will be assign to configs here.
```
posva: > You can define CLI options in this file
Would sound better imo

docs/build.md
```diff
@@ -0,0 +1,133 @@
...
+
+#### port
+
+Type: `number`
+
+Port of dev server.
```
posva: I think we should add the default value too

docs/build.md
```diff
@@ -0,0 +1,133 @@
...
+    require('autoprefixer')(options.autoprefixer)
+  ]
+}
+```
+
+#### template
+
+Type: `Object`
+
+[html-webpack-plugin](https://github.com/ampedandwired/html-webpack-plugin) options, use this option to customize `index.html` output, default value:
```
posva: It may be worth to note that the default value here point to a file of vue-cli

docs/build.md
```diff
@@ -0,0 +1,133 @@
...
+        loader: 'babel-loader', 
+        exclude: [/node_modules/],
+        options: {/* babel presets and plugins */}
+      }
+    ]
+  }
+}
+```
+
+Don't forget to install `babel-core` `babel-loader` and relevant presets in your project or `.vue` directory.
```
posva: a `,` is missing between babel-core and babel-loader

docs/build.md
egoist: yeah develop 😅

posva:
It's looking pretty cool. I haven't tested yet though.
Great job! 👍

docs/build.md
egoist: any suggestion?

docs/build.md
```diff
@@ -0,0 +1,136 @@
+# vue build
+
+`vue build` command gives you a zero-configuration development setup, install once and build everywhere.
+
+## Features
+
+- **Not a boilerplate**: run a single command to develop your app
+- **Out of the box**: ES2015, single-file component with hot reloading and custom CSS preprocessors
+- **Customizable**: populate a `~/.vue/webpack.config.js` for custom webpack config
+- **Single-file component mode**: just run `vue build Component.vue` and boom!
```
posva: well, it depends on what the command do. It could be _and test it out in the browser_ or _and distribute it_

egoist: currently `test it out in the browser` is what it does, so this will conflict with `distribute component`, maybe only enable this behavior by adding `--component` option?

zigomir:
@egoist playing with this in your repo/branch, trying to run what `Get started` is saying: `bin/vue build component.vue` and I get

```
 ERROR  Failed to compile with 1 errors

This dependency was not found in node_modules:

* your-tasteful-component

Did you forget to run npm install --save for it?
> Open http://localhost:4000
```

and I don't know what I should do to make it work. I think `build.md` needs to be more accurate or code isn't doing everything yet.

egoist:
@zigomir you need to create component.vue before running this command, definitely needs to improve the error log too. (`your-tasteful-component` is the alias for your component)

egoist:
@zigomir doc and error log updated 🙌

egoist:
btw, I'm not sure if using `buble-loader` by default is a better option, I personally prefer babel for its extensibility.

zigomir:
@egoist awesome! What about generating a new component file if one does not exist? This is probably "too magical" or not too important at the moment…

bin/vue-build
```diff
@@ -0,0 +1,286 @@
...
+  webpackConfig.entry.client.push(path.join(__dirname, '../lib/default-entry'))
+  webpackConfig.resolve.alias['your-tasteful-component'] = options.entry
+} else {
+  webpackConfig.entry.client.push(options.entry)
+}
+
+// installed by `yarn global add`
+if (isYarn(__dirname)) {
+  // modules in yarn global node_modules
+  // becuase of yarn's flat node_modules structure
```
zigomir: Minor. `becuase` -> `because`.

bin/vue-build
```diff
@@ -0,0 +1,286 @@
...
+ */
+
+program
+  .usage('[entry]')
+  .option('--dist [directory]', 'Output directory for bundled files', 'dist')
+  .option('--port [port]', 'Server port', 4000)
+  .option('--host [host]', 'Server host', 'localhost')
+  .option('--prod, --production', 'Build for production', false)
+  .option('-c, --config [directory]', 'Use custom config directory')
+  .option('--no-config', 'You do not want global and local config')
```
zigomir: I can't find where this config is used... 😊

bin/vue-build
```diff
@@ -0,0 +1,286 @@
...
+  if (hasLocalConfig) {
+    localConfig = require(localConfigPath)
+  }
+  if (hasLocalWebpackConfig) {
+    localWebpackConfig = require(localWebpackConfigPath)
+  }
+}
+
+var options = Object.assign({
+  entry: 'index.js'
```
zigomir: Because of this if you run only `vue build` you'll get `index.js does not exist, did you forget to create one?`. Maybe we should default to help output when no entry is provided?

docs/build.md
```diff
@@ -0,0 +1,144 @@
...
+
+```vue
+<template>
+  <h2>Hello World!</h2>
+</template>
+```
+
+And then just run `vue build component.vue` and go to `http://localhost:4000`
+
+To build for production:
```
zigomir: What does `production` mean in this case? If I understand you'd use `vue build` to prototype or develop a single component and then include/require it in your application with its own build process. Now, it's useful to showcase minified build but maybe this can give a false signal that one can use `--prod` build in their app directly and that might lead to including vue's runtime n times in their app.

This is more of a loud thinking here :)

test/e2e/test-build.js
```diff
@@ -0,0 +1,74 @@
...
+      const cssFile = files.filter(file => file.endsWith('.css'))[0]
+      expect(typeof cssFile).to.equal('string')
+      const cssContent = fs.readFileSync(path.join('dist', cssFile), 'utf8')
+      expect(cssContent).to.contain('-ms-flexbox')
+      done()
+    })
+  })
+
+  describe('build with local config in .vue directory', () => {
+    let result
```
zigomir: This one doesn't seem to be used.

package.json
```diff
@@ -25,28 +26,48 @@
   "homepage": "https://github.com/vuejs/vue-cli#readme",
   "main": "lib/index.js",
   "scripts": {
-    "test": "npm run lint && npm run e2e",
+    "test": "npm run lint && npm run e2e && npm run e2e:build",
     "lint": "eslint test/e2e/test.js lib bin/* --env mocha",
```
zigomir: Please also add `test-build.js` to be linted.

bin/vue-build
egoist: Use `--no-config` to set `config` to `false`

docs/build.md
egoist: both single-file component and regular entry are ok, in fact I feel directly test component doesn't make much sense, because component always needs props 😂

The `build for production` means built as a complete web app in production, maybe add a `distribute component` section to clarify more?

egoist:
It feels like `README.md` needs a rewrite, maybe move all `vue init` stuffs to `docs/init.md` too?

package.json
egoist: ✅

egoist:
@zigomir just added a link to build.md in readme https://github.com/vuejs/vue-cli/pull/287/commits/c96352d19c6bf3982e3ab8cf34d2c8a4a49a367b, you may want to refactor readme though.

zigomir:
@egoist looking good. I'll try out `--mount` as soon as I can.

bin/vue-build
```diff
@@ -29,12 +29,13 @@ program
   .option('--port [port]', 'Server port')
   .option('--host [host]', 'Server host', 'localhost')
   .option('--prod, --production', 'Build for production')
+  .option('--mount', 'Auto-create app instance from given component')
   .option('-c, --config [file]', 'Use custom config file')
   .option('-w, --webpack [file]', 'Use custom webpack config file')
-  .option('--no-config', 'You do not want to use config file')
-  .option('--no-webpack', 'You do not want to use webpack config file')
+  .option('--disable-config', 'You do not want to use config file')
```
egoist: changed `--no-config` to `--disable-config`, if we use `--no-config` the n it will default to `true`, have no idea about how to set the default value to `undefined`

package.json
```diff
@@ -25,28 +26,53 @@
...
     "request": "^2.67.0",
     "rimraf": "^2.5.0",
     "semver": "^5.1.0",
+    "tildify": "^1.2.0",
     "uid": "0.0.2",
-    "validate-npm-package-name": "^2.2.2"
+    "url-loader": "^0.5.7",
+    "user-home": "^2.0.0",
+    "validate-npm-package-name": "^2.2.2",
+    "vue": "2.1.8",
```
zigomir: Did you intentionally fix version to `2.1.8`? I'd say we want to give latest Vue 2 version whenever user installs or updates `vue-cli`. What do you think? (same for `vue-template-compiler` and maybe even `webpack` - I heard 2.2.0 non RC is coming soon).

zigomir:
Looking great! Played with it and it's an awesome experience. Anyone else has any comments before we merge?

yyx990803:
@zigomir thanks for reviewing, I will play with it this week to provide some final feedback

docs/build.md
```diff
@@ -0,0 +1,225 @@
...
+
+And then just run `vue build index.js` and go to `http://localhost:4000`
+
+To build for production (minimized and optimized):
+
+```bash
+$ vue build index.js --prod
+```
+
+If your want to directly test a component without manually create a Vue instance for it, just try:
```
zigomir: What about:

```
If your want to directly test a component without manually creating a Vue instance, try:
```

yyx990803:
Just tried it out - very smooth, nice work!

One thing I agree with: we should probably use `babel` instead of `buble` and include `babel-preset-vue` so that JSX also works out of the box. Thoughts?

yyx990803:
Also - we can probably use a loose version range for `vue` and `vue-template-compiler`.

yyx990803:
One more thing: we can actually detect wether we should use `--mount` mode by looking at the extension of the entry. `--mount` should be used automatically if the entry file is a vue file.

egoist:
@yyx990803 done for the first 2 suggestions, but for:

> --mount should be used automatically if the entry file is a vue file

If we make `--mount` the default behavior, users might be confused when they want to distribute the component instead of demonstrating it, since the entry in `--mount` mode is not the component itself, they have to modify webpack config. The alternative solution might be a `--disable-mount` option, but I would prefer everything being consistent.

Anyways we will need a `--mount` option, the component might also end with a `.js` extenstion (jsx component)

yyx990803:
I see the point about `--mount`, but as of now if I do this:

```
vue build Component.vue --prod
```

The resulting bundle cannot be distributed as a component because the webpack config doesn't have `output.library` and `output.libraryTarget`.

Maybe in this case if the entry is a `*.vue` file, we should automatically infer `output.library` from the filename and set `output.libraryTarget` to umd.

egoist:
> Maybe in this case if the entry is a *.vue file, we should automatically infer output.library from the filename and set output.libraryTarget to umd.

this sounds reasonable.

yyx990803:
Alternatively, this might be simpler and cater to more use cases:

- Always build in "app mode" by default. This means if the entry is a `*.vue` file, `--mount` will implicitly be used;

- For JSX entry components, the user can explicitly add `--mount`;

- Add new option `--lib` for distributing a component as a library:

  ```
  ```

egoist:
@yyx990803 Done.

yyx990803:
Awesome! THIS IS BEAUTIFUL! Thanks for the hard work!!!

PR #303, opened 2017-01-23
Add `--offline` option, closed #280

### LINKED ISSUE #280: Add `--use-cache` option
Maybe we can download the repo to `~/.vue-cli/repos` instead of a random temp path so that we can use cached version later. 🤔 

How do you think?

egoist:
For example:

```bash
vue init webpack project --offline
# use template at ~/.vue-templates/webpack
```


posva:
That was short 😄 
Does it work under Windows?

egoist:
@posva Yep.

PR #305, opened 2017-01-24
Add watch mode

cretueusebiu:
Adds `--watch` to run webpack in watch mode.

### DIFF
--- a/bin/vue-build
+++ b/bin/vue-build
@@ -37,6 +37,7 @@ program
   .option('-o, --open', 'Open browser')
   .option('--proxy [url]', 'Proxy API request')
   .option('--lib [libraryName]', 'Distribute component in UMD format')
+  .option('--watch', 'Run in watch mode')
   .parse(process.argv)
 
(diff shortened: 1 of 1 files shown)


bin/vue-build
```diff
@@ -189,7 +191,8 @@ var webpackConfig = {
         babel: babelOptions
       }
     })
-  ]
+  ],
+  watch: options.watch
```
egoist: no need to set this?

bin/vue-build
```diff
@@ -290,6 +293,12 @@ if (!options.disableWebpackConfig) {
 
 try {
   var compiler = webpack(webpackConfig)
+
+  if (options.watch) {
+    compiler.watch({
```
egoist: should replace the `compiler.run` below with this

egoist:
Thanks for the PR but I decided to make a new one at https://github.com/vuejs/vue-cli/pull/322 due to no response and there're many more changes needed than I expected.

PR #322, opened 2017-02-01
Add watch mode

egoist:
Introducing watch mode:

```bash
$ vue build Counter.vue --watch
```

Like the default development mode but remove HMR support and uses a real file system `compiler.watch`


yyx990803:
Hmm this is technically a breaking change though (`-w` now means something different). Is it possible to add this in a backwards compatible way?

egoist:
sure, but I find using `-w` for webpack config is a bit weird in the first place 😂 especially we have `--watch` now

yyx990803:
Well I think it's actually ok to have breaking changes for CLI stuff, especially when these are just dev features.

PR #328, opened 2017-02-03
Copy static folder

egoist:
It's very common to put a `favicon.ico` in dist folder, adding a `copy-webpack-plugin` would make it easier and work out of the box since the main idea of `vue build` is `zero-config`. Everything in `./static` folder will be copied to `./dist`


BusyHe:
image
when i change build directory error

image

how to do get this output directory

image

PR #339, opened 2017-02-08
Creating a vue file with vue-cli

abdullah:
I wrote a patch code which I use in my Vue projects.

It creates vue files from terminal where the user define specific path or default path `src/components`.

I have implemented code for vue-cli 

https://github.com/abdullah/cliv


posva:
Hi, thanks for sharing!

I don't think this solution is flexible enough. While we may add this feature in the future, it may be a better idea to simply use something like [plop](https://github.com/amwmedia/plop) in the template and add some npm scripts to generate the files

abdullah:
`plop` is a third party tool but the thing I am trying to do in here will be included in Vue Cli itself. I found it useful to create new component files. Also you mentioned that this solution is not flexible enough. I can understand that however if you are willing to accept this kind of thing into Cli I can try to make it more flexible. Let me know.

yyx990803:
Thanks for the PR - however,  the general philosophy of `vue-cli` is that it is only responsible for scaffolding the initial project - project-specific tasks are delegated to the scripts that are contained in the projects themselves. This allows for more flexibility. For example a project-specific scaffold command can eliminate the need for having to specify a folder name, but if it's done in `vue-cli` the folder name always need to be specified because we don't know which template the user used.

Issue #283, opened 2017-01-03
failed to init from github enterprise (with a custom hostname) repository


ETiV:
## what happened

1. command I use

```
vue init -c github:git.company.com:User/webpack-simple
```

2. I expected the vue-cli **clones** the git repo from the specific URL to initiate a new project.

3. but after answering the questions, it returns the following error:

```
vue-cli · Local template "git.company.com:User/webpack-simple" not found.
```
----
## why it happens

I've checked the `bin/vue-init` file,
It seems like it's a conflict with the `local template test`, at here:
https://github.com/vuejs/vue-cli/blob/master/bin/vue-init#L93-L94

```
./ - for Linux CWD,
\w: - for Windows Disk Volume?
```

But it **DO CONFLICT** with the given argument,
because the [download-git-repo doc](https://www.npmjs.com/package/download-git-repo) says:

> ... you can also specify a custom host like gitlab:custom.com:owner/name. Feel free to submit an issue or pull request for additional host options.

----

In my case, I simply modified the file to bypass the local template testing:
```
#L94    if (!clone && /^[./]|(\w:)/.test(template)) {
```

Because I don't know how to exactly test the `local template path`(so many scenarios), so this is an issue instead of PR.

likun7981:
you can change like this 
```js
/^\.{1,2}\/|^\w:|^\//.test(template) 
```

ETiV:
@likun7981 
```
// it will be failed on
/^\.{1,2}\/|^\w:|^\//.test('templates/example')
```

ETiV:
![layer](https://cloud.githubusercontent.com/assets/1926860/21630520/f219252a-d26c-11e6-8ca5-96eacbc9149d.png)

'templates/example' will be `github.com/templates/example`

I think it's better to add a switch to the command line arguments to specify whether to use a local template. Like:
```
vue init --local ${LOCAL_PATH}
```

likun7981:
This is not wrong, `'templates/example'` is not a local path , it should return false

likun7981:
These tests are all right
```js
var localPathRegx = /^\.{1,2}\/|^\w:|^\//  

localPathRegx.test('../')  //true
localPathRegx.test('./')  //true
localPathRegx.test('c:/') //true
localPathRegx.test('template/example')  //false
localPathRegx.test('github:git.company.com:User/webpack-simple')  //false
```

zigomir:
Should be fixed now with #341.

[...]

PR #341, opened 2017-02-09
Extract local path logic and add tests.

zigomir:
Fixes #300 and #283.

Thanks @ajanes93 for opening a PR. Also thanks to @likun7981 for providing regex.
Only thing wrong in regex imho was assumption that windows drive letter can be of any length and thus `bitbucket:username/rep` was treated as local template. I fixed this in this PR and also added some tests.

I'm aware that running these tests on Windows won't pass because of `isAbsolute` used, but CircleCi runes on *nix so... I'm open to any suggestions.

@egoist what do you think?

### DIFF
--- a/lib/local-path.js
+++ b/lib/local-path.js
@@ -1,0 +1,13 @@
+var path = require('path')
+
+module.exports = {
+  isLocalPath: function (templatePath) {
+    return /^[./]|(^[a-z|A-Z]:)/.test(templatePath)
+  },
+
+  getTemplatePath: function (templatePath) {
+    return path.isAbsolute(templatePath)
+      ? templatePath
+      : path.normalize(path.join(process.cwd(), templatePath))
+  }
+}
--- a/lib/local-path.js
+++ b/lib/local-path.js
@@ -2,7 +2,7 @@ var path = require('path')
 
 module.exports = {
   isLocalPath: function (templatePath) {
-    return /^[./]|(^[a-z|A-Z]:)/.test(templatePath)
+    return /^[./]|(^[a-zA-Z]:)/.test(templatePath)
   },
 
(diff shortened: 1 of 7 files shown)


posva:
Maybe we should use appveyor for vue-cli to run tests on a windows machine too

zigomir:
@posva I added appveyor which is finally passing now. I had configuration wrong and it was running against node 4 (which is fine) but there was npm v2 installed which wasn't fine.

I also found we had some race conditions in tests, reading from same directory (thanks Windows) etc. Should be all fixed in this PR now.

posva:
@zigomir 😆 I see that you had some trouble due to the commit history. I hope it wasn't too much of a pain

lib/local-path.js
```diff
@@ -0,0 +1,13 @@
+var path = require('path')
+
+module.exports = {
+  isLocalPath: function (templatePath) {
+    return /^[./]|(^[a-z|A-Z]:)/.test(templatePath)
```
posva: Why is the `|` necessary?

zigomir: Huh, might not be at all :) (regex by trial and error image here 😊)

zigomir:
@posva hehe yeah. No worries, I think it's good that we run tests on Windows too. + I learned a bit how appveyor works.

I changed regex as you pointed out in CR. Let me know if there is anything else we should change.

Also, worth noting are weird race condition issues that were resolved here: https://github.com/vuejs/vue-cli/pull/341/commits/1f15a0184fde38f275b538edc56b8fae91a6262f

Qquanwei:
thank your work <3
I need this feature , but it seems not publish to npm, what should I do?

PR #369, opened 2017-02-22
Use github personal access token if provided in env

### LINKED ISSUE #368: ‘vue list’Command execution（使用vue list出现问题？）
API rate limit exceeded for 116.216.128.235. (But here's the good news: Authenticated requests get a higher rate limit. Check out the documentation for more details.)

The solution?
-----------------
这是我翻译的

API速率限制为116.216.128.235超过。(但这是一个好消息:经过身份验证的请求得到更高的速率限制。查看更多细节的文档)。

请问 有解决方案吗？O(∩_∩)O谢谢

znck:
This PR includes
- Use `GITHUB_PERSONAL_ACCESS_TOKEN` if defined in `env`

Fixes #368

### DIFF
--- a/bin/vue-list
+++ b/bin/vue-list
@@ -18,8 +18,8 @@ process.on('exit', function () {
  */
 var auth
 
-if (process.env.GITHUB_PERSONAL_ACCESS_TOKEN) {
-  auth = { bearer: process.env.GITHUB_PERSONAL_ACCESS_TOKEN }
+if (process.env.GITHUB_TOKEN) {
+  auth = { bearer: process.env.GITHUB_TOKEN }
 }
 
 request({
--- a/bin/vue-list
+++ b/bin/vue-list
@@ -16,12 +16,18 @@ process.on('exit', function () {
 /**
  * List repos.
  */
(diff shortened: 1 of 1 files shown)


posva:
I'd actually call it `GITHUB_TOKEN` because I already have a variable with that name in a `.tokenrc` file I just source. So I think it would be better to pass an option to the cli so users can pass their tokens as they want and detect this error to point them how to use a custom token (there must be some kind of gh blog post about it)

znck:
That would be better.

So,  `--token` would set github personal access token.

Should it load `GITHUB_TOKEN` from env. ( 😑  it cumbersome to add token every time `vue list` is called)

zigomir:
@znck there is not a lot api rate limit exceeded issues reported no more. But, this simple fix is probably still very valuable  to people who gets it. I think we'd need to add a clear documentation on how to specify `GITHUB_TOKEN` to make this work; Link to [GitHub Personal Access Tokens](https://github.com/settings/tokens) and describe every step.

What do you think?

yyx990803:

PR #385, opened 2017-03-08
Make prompts when support function.

likun7981:
Now  we only set prompts __when__ is an attribute from the other prompts , and the prompt must be a  boolean . but sometimes we want set the __when__ to a special value,  when a prompt choice the value , we return a boolean.  
so I resolve this problem; I through this example to illustrate;

previously , we can only set prompt like this
```js
q1: {
  type: 'confirm',
  require: true,
  message: 'this is a question?',
  default: true
},
q2: {
  when: 'q1',
  type: 'list',
  message: 'Which choice ?',
  choices:[] // some choices
}
```
If __q1__ is not a _confirm_ , it is a _list_, like this 
```js
q1: {
  type: 'list',
  message: 'make choice?',
  choices:['c1','c2'] // some choices
},
q2: {
  when: ????,
  type: 'list',
  message: 'Which choice ?',
  choices:[] // some choices
}
```
now  we want to when __q1__ choice _c1_ , then ask __q2__, otherwise do not ask __q2__;

so  when  should like __inquirer__ to support set _function_ value to __when__

we can set like this:
```js
q1: {
  type: 'list',
  message: 'make choice?',
  choices:['c1','c2'] // some choices
},
q2: {
  when: function(answers){
    return answers.q1 === 'c1';
  },
  type: 'list',
  message: 'Which choice ?',
  choices:[] // some choices
}
```

### DIFF
--- a/lib/ask.js
+++ b/lib/ask.js
@@ -32,11 +32,9 @@ module.exports = function ask (prompts, data, done) {
  */
 
 function prompt (data, key, prompt, done) {
-  // skip prompts whose when condition is not met
-  if (prompt.when && !evaluate(prompt.when, data)) {
+  if (prompt.when && !(typeof prompt.when === 'function' ? prompt.when(data) : evaluate(prompt.when, data))) {
     return done()
   }
-
   var promptDefault = prompt.default
   if (typeof prompt.default === 'function') {
     promptDefault = function () {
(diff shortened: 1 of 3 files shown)


posva:
Can you add a test for this, please? 🙂

likun7981:
The test is ok

posva:
ping @egoist @zigomir

zigomir:
Looks good to me! I only maybe miss a test case that exercises previous `when` functionality. I know this is not a concern for this PR, but it might be useful to prevent any regressions.

likun7981:
No one to deal with this PR ?  😳

yyx990803:

PR #388, opened 2017-03-08
Escape double quotes in string prompts

### LINKED ISSUE #382: Escaping for package.json
I am certain if this is a issue with vue-cli or the template, however I think the "setup questions" are asked by the cli?

My problem: I just setup a new project and followed the standard procedure:
See this

As you can see, I used double quotes inside of the Author field (for my nickname). This resulted in the following error:
image

When looking into my generated package.json, you can see that the double quotes just got copied into the `author` field:

```json
....
"author": "Pascal "Arial7" Riesinger <riesinger.pascal@gmail.com>",
  "private": true,
  "scripts": {
....
```

So my suggestion is to escape the user input before writing it to the `package.json` file.

Arial7:
This fixes #382.
It checks if the answer type is `string`, and if it is, `replace`s `"` with `\"`, so it can be parsed as valid JSON.

### DIFF
--- a/test/e2e/test.js
+++ b/test/e2e/test.js
@@ -34,9 +34,6 @@ function monkeyPatchInquirer (answers) {
 }
 
 describe('vue-cli', () => {
-  /* eslint-disable quotes */
-  const escapedAuthor = "John \"The Tester\" Doe <john@doe.com>"
-
   const escapedAnswers = {
     name: 'vue-cli-test',
     author: 'John "The Tester" Doe <john@doe.com>',
@@ -130,7 +127,7 @@ describe('vue-cli', () => {
       const pkg = fs.readFileSync(`${MOCK_TEMPLATE_BUILD_PATH}/package.json`, 'utf8')
       try {
         var validData = JSON.parse(pkg)
-        expect(validData.author).to.equal(escapedAuthor)
+        expect(validData.author).to.equal(escapedAnswers.author)
         done()
       } catch (err) {
         done(err)
--- a/test/e2e/test.js
+++ b/test/e2e/test.js
@@ -34,6 +34,22 @@ function monkeyPatchInquirer (answers) {
 }
 
 describe('vue-cli', () => {
+  /* eslint-disable quotes */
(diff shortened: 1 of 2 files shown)


posva:
Can you add a test case for this?
You can reuse one of the tests and verify that the generated package json is correct by parsing it

Arial7:
I'm not home anymore, but I can do this tomorrow ;)

If I understood you correctly, the test should just verify that the
package.json is valid JSON?

Eduardo San Martin Morote <notifications@github.com> schrieb am Mi., 8.
März 2017, 12:06:

posva:
Yes, so that the double quotes are escaped correctly

Arial7:
Okay, I think that I should be able to do this :)

Eduardo San Martin Morote <notifications@github.com> schrieb am Mi., 8.
März 2017, 12:12:

Arial7:
I just added a test to check the escaping. I tired to leave the other tests alone and add it seperately... May not be the best test, but it should do the trick 😄

test/e2e/test.js
```diff
@@ -106,6 +122,22 @@ describe('vue-cli', () => {
...
 
+  it('generate a vaild package.json with escaped author', done => {
+    monkeyPatchInquirer(escapedAnswers)
+    generate('test', MOCK_TEMPLATE_REPO_PATH, MOCK_TEMPLATE_BUILD_PATH, err => {
+      if (err) done(err)
+
+      const pkg = fs.readFileSync(`${MOCK_TEMPLATE_BUILD_PATH}/package.json`, 'utf8')
+      try {
+        var validData = JSON.parse(pkg)
+        expect(validData.author).to.equal(escapedAuthor)
```
posva: Can't you just do `expect(validData.author).to.equal(escapedAnswers.author)` ?

Arial7: Jup, could definitively do that.... My mistake 😂

Arial7: Just woke up an hour ago, still not very awake

zigomir:
LGTM 👍

PR #390, opened 2017-03-09
If no offline specified. We should remove the tmp dir if it is existed.

yxwu25:


bin/vue-init
```diff
@@ -62,9 +63,11 @@ var to = path.resolve(rawName || '.')
 var clone = program.clone || false
 
 var tmp = path.join(home, '.vue-templates', template.replace(/\//g, '-'))
-if (program.offline) {
+if (program.offline && exists(tmp)) {
   console.log(`> Use cached template at ${chalk.yellow(tildify(tmp))}`)
   template = tmp
+} else {
+  rm(tmp)
```
znck: IMO it should overwrite existing file after successful download.

yxwu25: Yes, that's best. Needs another dir to cache the downloads. Otherwise the git will throw an error.

znck: Native `/tmp` (😝  And whatever alternative for windows) directory can be used for that.

yyx990803:

Issue #401, opened 2017-03-16
Parse error on line 125


OmgImAlexis:
After running `vue init ~/code/simon twistly-frontend` I get this error. 
Template repo: https://github.com/OmgImAlexis/simon
```
? Project name twistly-frontend
? Project description Frontend for Twistly
? Author Alexis Tyler <xo@wvvw.me>
? Need a client? Yes
? Need a server? No
   vue-cli · Parse error on line 125:
...== 1) {      // x{{a,b}}y ==> x{a}y x{b
----------------------^
Expecting 'ID', 'STRING', 'NUMBER', 'BOOLEAN', 'UNDEFINED', 'NULL', 'DATA', got 'INVALID'
```

posva:
Handlebars is trying to parse that but it fails because it's not a syntax it knows. You should be able to escape it with `\{{}}`.

OmgImAlexis:
@posva would it be possible to have something more descriptive added here as this doesn't really point back to the file with the issue instead of states there is one. Going through my code I can't actually find anything with that line in it as I'm assuming that's compiled code it's dealing with?

calvernaz:
@posva @OmgImAlexis The patch definitely gives a better error description for what's going on, but soon you will realize the issue is handlebars trying to interpolate a file you don't care about like a file inside node_modules, in order to avoid these interpolation errors use the property **skipInterpolation**.

`"skipInterpolation": ["src/**/*.vue", "node_modules/**"]`

PR #407, opened 2017-03-17
Add filename when in handlebars errors

posva:
When handlebars gives an error, it doesn't contain the filename that is actually producing the error.
Should help solve problems like #401 

Instead of 

```
Parse error on line 21
```

We now have

```
[readme.md] Parse error on line 21
```

@egoist @zigomir Feel free to modify how the filename is included in the error message


zigomir:
@posva nice!

PR #426, opened 2017-04-01
Add an alias from '@' to src

### LINKED ISSUE #422: vue build src/main.js --prod cant resolve @/components/Hello
vue build src/main.js --prod doesnt build because it cant resolve the include path '@/component/Hello'

Using npm 4, latest vue-cli and it happens on fresh install of vue init webpack my-project

npm run dev works just fine

posva:
Closes #422 
This makes it easier to combine with the official webpack template

### DIFF
--- a/bin/vue-build
+++ b/bin/vue-build
@@ -150,7 +150,9 @@ var webpackConfig = {
       cwd('node_modules'), // modules in cwd's node_modules
       ownDir('node_modules') // modules in package's node_modules
     ],
-    alias: {}
+    alias: {
+      '@': 'src'
+    }
   },
   resolveLoader: {
     modules: [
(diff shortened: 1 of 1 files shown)


bin/vue-build
```diff
@@ -150,7 +150,9 @@ var webpackConfig = {
       cwd('node_modules'), // modules in cwd's node_modules
       ownDir('node_modules') // modules in package's node_modules
     ],
-    alias: {}
+    alias: {
+      '@': 'src'
```
egoist: should use absolute path? like `cwd('src')`

posva: Is it better? I also wondered but since this worked I didn't add anything

LinusBorg: We use an absolute path for this alias in the webpack template as well:

https://github.com/vuejs-templates/webpack/blob/master/template/build/webpack.base.conf.js#L27

/ping @posva

posva: Forgot about this!

PR #428, opened 2017-04-01
Add support for metalsmith builder hooks

### LINKED ISSUE #427: Custom metalsmith plugin
I need to use a custom metalsmith plugin in after default vue-cli ones. (see [generate.js#L47-L49](https://github.com/vuejs/vue-cli/blob/439aece38f54c32f7e0de29f483f24f6523e7a80/lib/generate.js#L47-L49))

In my case, I need to move some files to another destination based on answers data, but I think it would be great to let writers of custom templates allow adding any metalsmith plugin in the chain.

Toilal:
Close #427

### DIFF
--- a/test/e2e/mock-metalsmith-custom/meta.js
+++ b/test/e2e/mock-metalsmith-custom/meta.js
@@ -1,0 +1,15 @@
+module.exports = {
+  "metalsmith": function (metalsmith, opts, helpers) {
+    function customMetalsmithPlugin (files, metalsmith, done) {
+      // Implement something really custom here.
+
+      var readme = files['readme.md']
+      delete files['readme.md']
+      files['custom/readme.md'] = readme
+
+      done(null, files)
+    }
+
+    metalsmith.use(customMetalsmithPlugin)
+  }
+}
--- a/test/e2e/mock-metalsmith-custom/meta.js
+++ b/test/e2e/mock-metalsmith-custom/meta.js
@@ -1,5 +1,6 @@
 module.exports = {
   "metalsmith": function (metalsmith, opts, helpers) {
+    metalsmith.metadata().custom = "Custom";
     function customMetalsmithPlugin (files, metalsmith, done) {
       // Implement something really custom here.
 
--- a/test/e2e/test.js
+++ b/test/e2e/test.js
@@ -13,6 +13,7 @@ const metadata = require('../../lib/options')
 const { isLocalPath, getTemplatePath } = require('../../lib/local-path')
 
 const MOCK_META_JSON_PATH = path.resolve('./test/e2e/mock-meta-json')
+const MOCK_METALSMITH_CUSTOM = path.resolve('./test/e2e/mock-metalsmith-custom')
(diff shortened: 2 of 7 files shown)


zigomir:
I think this would be even better with tests.

Toilal:
I'll write one test that writes target files to another destination, as I think it would be a common use case.

Toilal:
Test written. I have also removed `it.only` in `test.js` that was causing only one test to run.

test/e2e/test.js
```diff
@@ -254,7 +265,7 @@ describe('vue-cli', () => {
     expect(getTemplatePath('../template')).to.equal(path.join(__dirname, '/../../../template'))
   })
 
-  it.only('points out the file in the error', done => {
```
zigomir: 👍

test/e2e/mock-metalsmith-custom/meta.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  "metalsmith": function (metalsmith, opts, helpers) {
+    function customMetalsmithPlugin (files, metalsmith, done) {
+      // Implement something really custom here.
+
+      var readme = files['readme.md']
+      delete files['readme.md']
+      files['custom/readme.md'] = readme
+
+      console.log(files)
```
zigomir: I believe this could be removed.

test/e2e/mock-metalsmith-custom/meta.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  "metalsmith": function (metalsmith, opts, helpers) {
+    function customMetalsmithPlugin (files, metalsmith, done) {
+      // Implement something really custom here.
+
+      var readme = files['readme.md']
+      delete files['readme.md']
+      files['custom/readme.md'] = readme
```
zigomir: Just to be sure. What you're doing here in custom metalsmith plugin is you're moving file from root template directory into `custom` directory, right?

test/e2e/mock-metalsmith-custom/meta.js
```diff
@@ -0,0 +1,17 @@
+module.exports = {
+  "metalsmith": function (metalsmith, opts, helpers) {
```
zigomir: Besides this one you can also test `before` and `after`, right?

test/e2e/mock-metalsmith-custom/meta.js
Toilal: oups yes :)

test/e2e/mock-metalsmith-custom/meta.js
Toilal: Yes that's it.

test/e2e/mock-metalsmith-custom/meta.js
Toilal: ok

Toilal: I just push a new commit with an additional test, and better assertions

zigomir:
@Toilal looks good, thank you!

PR #430, opened 2017-04-05
Add filter support for optional questions

### LINKED ISSUE #408: Apply a filter on undefined response data
I'm currently writing a custom template based on webpack template, and I ask user the JavaScript compiler he wants.

Based on the answer, some other questions are removed using `when`

```js
"compiler": {
      "type": "list",
      "message": "Which JavaScript compiler do you want to use?",
      "choices": [
        {
          "name": "Babel (ES2015)",
          "value": "babel",
          "short": "babel"
        },
        {
          "name": "TypeScript (Awesome Typescript Loader)",
          "value": "typescript",
          "short": "typescript"
        },
        {
          "name": "No compiler (Vanilla JS)",
          "value": "none",
          "short": "none"
        }
      ]
    },
    "eslint": {
      "when": "compiler == 'babel'",
      "type": "confirm",
      "message": "Use ESLint to lint your code?"
    },
```

The eslint question is displayed only when compiler babel is choosen, all is ok here.

But then, if user choose something else than babel as compiler, an error message is displayed about filters.

```
Error when evaluating filter condition: eslint
```

It seems to be caused by the filters handling because eslint is undefined in data.

```
  "filters": {
    ".eslintrc.js": "eslint",
    ".eslintignore": "eslint",
    "config/test.env.js": "unit || e2e",
    "test/unit/**/*": "unit",
    "build/webpack.test.conf.js": "unit",
    "test/e2e/**/*": "e2e",
    "src/router/**/*": "router"
  }
```

Toilal:
undefined value is now defined in metalsmith metadata when meta.js "when" property returns false. This makes the filter evaluation works for filters using data that have been skipped in the prompt workflow by using the default value of the field.

Close #408

### DIFF
--- a/lib/ask.js
+++ b/lib/ask.js
@@ -34,6 +34,8 @@ module.exports = function ask (prompts, data, done) {
 function prompt (data, key, prompt, done) {
   // skip prompts whose when condition is not met
   if (prompt.when && !evaluate(prompt.when, data)) {
+    // set undefined value to avoid failure on filter evaluations
+    data[key] = undefined
     return done()
   }
 


zigomir:
@Toilal do you have time or will to write a test for this case?

yyx990803:

PR #438, opened 2017-04-14
Fixes: cloning with colons

mattixittam:
This fixes two issues:

- When using a repo like _bitbucket:local.url.com:org/name_ in vue init, the temp directory made would have multiple colons in it, this caused git to fatal.


yyx990803:
This has included in #565 but thanks for the PR.

PR #450, opened 2017-05-05
clean cached template in downloadandgenerate

### LINKED ISSUE #403: Issue with using vue-cli with private bitbucket repo
I've got a customized version of the webpack template that I'm using on my projects. Hosting it on a public github account works fine, but my organization has decided to go with private bitbucket repositories. I copied my github repository to bitbucket and went to test creating a new vue app with the cli and my bitbucket template with the following command:

`vue init -c bitbucket:rwb7041/vuetemplate myproject`

I get the following error:
`vue-cli · Failed to download repo vuejs-templates/myproject: 'git clone' failed with status 128`

So next I tried creating the myproject directory and running the following inside that directory:
`vue init -c bitbucket:rwb7041/vuetemplate`

And I got the following error (after responding 'Y' to the question of generating project in the current directory:
`vue-cli · Local template "bitbucket:rwb7041/vuetemplate" not found.`

I am using version 2.8.1 of the vue-cli. What am I doing incorrectly?

### LINKED ISSUE #409: Consider removing template dirs in ~/.vue-templates/ when downloading updated templates
Hello, I maintain a `vue-cli` template and I recently made some updates to make use of Vue's new SSR features. Part of stealing Evan You's code from the hackernews example involved renaming `client-entry.js` and `server-entry.js` to `entry-client.js` and `entry-server.js` respectively.

Upon testing out the newly pushed code, I discovered that `vue-cli` was still scaffolding `client-entry.js` and `server-entry.js` as well as `entry-client.js` and `entry-server.js`.

This problem was resolved by deleting the `.vue-templates` folder in my home dir.

I suspect this will become a bigger issue when any official/popular template renames (or possibly deletes) a file. And it will affect any user that used the template prior to the file changes.

Because of this, I think that at a minimum, a note about the `.vue-templates` dir should be added to the README - Though I believe this is a subpar solution as the user may not even be aware that there are problems/dead code.

I think, ideally, `vue-cli` should remove template dirs when downloading updated templates.

Thank vue very much for your time!

steven-klein:
Remove the currently cached template when we are doing a download and generate.  

This helps to resolve #403 where using the --clone flag with private repos fails when the cached template directory already exists.

It also helps resolve #409 where init results in a template that is outdated.

### DIFF
--- a/bin/vue-init
+++ b/bin/vue-init
@@ -142,7 +142,7 @@ function run () {
 function downloadAndGenerate (template) {
   var spinner = ora('downloading template')
   spinner.start()
-  // check if template is local
+  // Remove if local template exists
   if (exists(tmp)) rm(tmp)
   download(template, tmp, { clone: clone }, function (err) {
     spinner.stop()
--- a/bin/vue-init
+++ b/bin/vue-init
@@ -9,6 +9,7 @@ var home = require('user-home')
 var tildify = require('tildify')
 var chalk = require('chalk')
 var inquirer = require('inquirer')
+var rm = require('rimraf').sync
 var logger = require('../lib/logger')
 var generate = require('../lib/generate')
 var checkVersion = require('../lib/check-version')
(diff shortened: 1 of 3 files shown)


bin/vue-init
```diff
@@ -141,6 +142,8 @@ function run () {
 function downloadAndGenerate (template) {
   var spinner = ora('downloading template')
   spinner.start()
+  // check if template is local
```
egoist: Should use a proper comment, eg: `Remove if local template exists`

PR #518, opened 2017-07-04
fix vue-init fail from gitlab in Windows OS

winnieBear:
the path can NOT contain `:` in Windows OS,  which result in failure of initialization by `vue-init -c gitlab:gitlab.host:user/repo project-name` in Windows OS. when `downloadAndGenerate`, the tmp path name may look like ` C:\Users\username\.vue-templates\gitlab:githost:repo-name`, the character : in path lead `git clone` fail in Window OS.

### DIFF
--- a/bin/vue-init
+++ b/bin/vue-init
@@ -65,7 +65,8 @@ var name = inPlace ? path.relative('../', process.cwd()) : rawName
 var to = path.resolve(rawName || '.')
 var clone = program.clone || false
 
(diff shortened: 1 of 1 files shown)


kospot:
Hi, when i use the http to clone package, I found the symbol ':' is can't be in the file name( Window OS)
temp should be this
`var tmp = path.join(home, '.vue-templates', template.replace(/[\/|:]/g, '-'))`

yyx990803:
Closing as this has been included in #565, thanks.

PR #519, opened 2017-07-05
add local package's node_modules directory to resolve{Loader}.modules if it exists

UncleBill:
Once I run `vue build foo.vue` in a **sub-directory** of my project, which contains `less` code, I got Error:

``` shell
Module not found: Error: Can't resolve 'less-loader' in 'path\to\proj\src\components'
```
I already had `less-loader` installed in `path\to\proj`, root directory of the project.

I don't want to create `path\to\proj\src\components\node_modules` with lots of packages (`less-loader` and its dependencies ) installing.

### DIFF
--- a/bin/vue-build
+++ b/bin/vue-build
@@ -130,6 +130,7 @@ if (useBabelRc) {
 }
 
 var filenames = getFilenames(options)
+var resolveModules = getModules()
 
 var webpackConfig = {
   entry: {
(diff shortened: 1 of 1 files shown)


UncleBill:
Any comments? @egoist @posva

egoist:
hmm why not just run it in the root dir with `npm run`?

posva:
I second the question but I think this is worth adding. We could use some existing packages to get the project's folder though

UncleBill:
>hmm why not just run it in the root dir with npm run?
   
Isn't that convenient to run wherever inside your project? Make `vue build` "intelligent", it will know you are in a subdirectory, and know where to find exact dependencies for your component.

UncleBill:
>run wherever inside your project?

Came up with an idea: why can't it be "outside" too?It will be "any where"! We can implement this by searching the root of the project from `__dirname` of the component file  instead of `cwd`. 💯

UncleBill:
@egoist @posva 
Like `npm`'s behavior, when you execute `npm` inside a node project, `npm` will search `package.json` to location root directory.So that the `npm scripts` will run from root directory.

yyx990803:

PR #550, opened 2017-08-06
hot reload not working - solutions

shamane:


posva:
Thanks a lot for your interest in helping the project but this doesn't belong to the vue-cli project.
It's better to file an issue if there's a bug first though 🙂

PR #551, opened 2017-08-07
Add support for custom template in gitlab, such as `vue init gitlab:c…

qbcbyb:
Add support for custom template in gitlab, such as `vue init gitlab:custom.com:owner/name Test -c`


posva:
Thanks for the feature (even though you didn't add a new test case and tests aren't passing 😛 ) but it's easier to keep the git syntax for gitlab and have only one exception for github, it's also easier to copy paste the git link. I think it's worth to add a note about using git for gitlab on the readme though. Searching for gitlab yields no results at the moment. A PR for that will be welcome 🙂

PR #558, opened 2017-08-11
The current working directory must be exist

zce:
The current working directory must be exist, there is no need to check whether the directory exists again


posva:
That check doesn't look necessary, what is the case you want to handle?

zce:
The `exist(to)` always return true when `inPlace` is true, I just think it's possible to skip a call here

PR #561, opened 2017-08-14
using es6 features

TokaLazy:
I'll modify other files but i prefer moving forward step by step and have your opinion


TokaLazy:
ok... i failed :D

PR #564, opened 2017-08-17
Add option to run vue-cli through npx

josephrexme:


josephrexme:
I just realized this isn't working as expected right now so it's great that it isn't merged yet. I created an issue for this at zkat/npx#112

README.md
```diff
@@ -16,6 +16,12 @@ $ npm install -g vue-cli
 $ vue init <template-name> <project-name>
 ```
 
+Or you may use npx and avoid global install if you are on npm >= 5
+
+```bash
+npx vue init <template-name> <project-name>
```
posva: You're right, we have to provide `vue-cli` to `npx`

posva:
@josephrexme It does work for me though...

josephrexme:
I just changed it to use `vue-cli` instead

README.md
```diff
@@ -16,7 +16,11 @@ $ npm install -g vue-cli
 $ vue init <template-name> <project-name>
 ```
 
-> You can also avoid installing the CLI and run **npx vue init <template-name> <project-name>**
+Or you may use npx and avoid global install if you are on npm >= 5
```
notyeti: Note: npx and the workflow are also supported on npm 4.3.

yyx990803:
Thanks for the PR, but I don't really think there's a point in installing `vue-cli` locally in your project. It's used to create a project in the first place and itself is not updated that often.

shahilm:
I installed vue not globally as I need permission from my system admin , so I wantto run with npx

rashnk:
I didnot install vue-cli globally,` npx vue create hello-world` worked for me

Issue #525, opened 2017-07-13
clone from private repository


yangblink:
i use private gitlab,  the vue-cli version is `2.8.2`

if i clone with project-name like this
```
vue init -c gitlab.com:username/my-vue-template test
```
it show err:
```
vue-cli · Failed to download repo vuejs-templates/test: 'git clone' failed with status 128
```
it seems  that the order of argument is wrong

use this cammnd is work fine
```
vue init -c test gitlab.com:username/my-vue-template 
```
and this
```
vue-init  gitlab.com:username/my-vue-template test
```

posva:
From readme:

> If you would like to download from a private repository use the `--clone` flag and the cli will use `git clone` so your SSH keys are used.

yangblink:
@posva   thanks for quick prompt
I mean it's should 
```
# yes
vue init --clone my-project username/repo   # it's right on my mac

# no 
vue init --clone username/repo my-project  # vue will treate `my-project` as the repository addr
```

or it's just the bug of my `commander` module

posva:
I see the bug, I'll fix it

posva:
I gave it a look but I don't understand why the order is being messed up when loading the args. Can you take a quick look @egoist , please?

posva:
BTW, using the option at the end works as expected

mrcsmcln:
Not sure if this helps, but this appears to work (just replace `vue init` with `vue-init`):

```
vue-init -c gitlab:user/repo my-project
```

bajras:
Im having a similar kind of issue with the custom template and its hosted in my work's private bitbucket repo and get the following error:

`vue-cli · Failed to download repo ssh://git@bitbucket.moo.com:7999/mac/vue-moo-webpack-boilerplate.git: 'git clone' failed with status 128`

I tried all the mentioned steps: 
`vue init -c ssh://git@bitbucket.moo.com:7999/mac/vue-moo-webpack-boilerplate.git yay
vue init ssh://git@bitbucket.moo.com:7999/mac/vue-moo-webpack-boilerplate.git --clone yay`

When I put in on a public Github repo it works fine but not sure why its not working with the private bitbucket repo? 

When I do git clone bitbucket of the exact repo it works fine as so I can tell its not the connection issue with SSH.

`git clone ssh://git@bitbucket.moo.com:7999/mac/vue-moo-webpack-boilerplate.git`

Any help would be highly appreciated. Thanks.

mrcsmcln:
[...]

PR #565, opened 2017-08-18
fix the  vue init error from private repository  in window OS

kospot:
if the file name has symbol ':', will make this error like :
'git clone' failed with status 128


bin/vue-init
```diff
@@ -65,7 +65,7 @@ var name = inPlace ? path.relative('../', process.cwd()) : rawName
 var to = path.resolve(rawName || '.')
 var clone = program.clone || false
 
-var tmp = path.join(home, '.vue-templates', template.replace(/\//g, '-'))
+var tmp = path.join(home, '.vue-templates', template.replace(/[\/|:]/g, '-'))
```
posva: when do you need to replace the `|` character?

kospot: the | character means or

and is that you planing publish the npm package ?

posva: using `[]` will already do an _or_
I don't get your question, can you rephrase it, please?

rwb7041:
Is there a timeline of when this change will be added to the cli and published to NPM? This change will help me out a lot as I have a custom template I've been using out of a private repo and I can use that template on my mac just fine, but I cant on my windows machine due to the filename error.

weizhenye:
@posva @yyx990803 

Could please you modify this PR yourself and merge it? There are hundreds issues and PRs ( #438 #518 #525 #567 ) trying to fix this bug which already exists 8 months. Or just reopen and merge #567

posva:
This PR is not solving those issues... It's still not clear to me what is this solving that doesn't work on current vue-cli version

rwb7041:
@posva It's solving a windows error that prevents the vue-cli from being able to use template that are not on github, such as templates in private bitbucket repositories. The cli fails because it tries to make a folder with a colon in the name, and colons are not allowed in file/folder names on windows, just like the \ character. (I found the full list of reserved characters for Windows here: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247(v=vs.85).aspx)

PR #567, opened 2017-08-18
fix issues#525

nodejser:
fix issues [clone from private repository bug](https://github.com/vuejs/vue-cli/issues/525)


posva:
Thanks a lot for your PR, but we already got #565

PR #616, opened 2017-10-18
Auto correct default names

qw3n:
Names containing spaces and capitals will not pass validation so correct them for default name suggestion. Alternatively this could be changed to convert the name to kebab case.


yyx990803:

Issue #655, opened 2017-11-30
Error installing vue-cli because coffeescript moved


fellz:
Hi
System: Windows 10, node version 8.3.0
`npm i -g vue-cli` gives this error:

```
> coffee-script@1.12.8 postinstall C:\Users\Dave\AppData\Roaming\npm\node_modules\vue-cli\node_modules\coffee-script
> node --eval 'if (require("./package.json").name === "coffee-script") { var red, yellow, cyan, reset; red = yellow = cyan = reset = ""; if (!process.env.NODE_DISABLE_COLORS) { red = "\x1b[31m"; yellow = "\x1b[33m"; cyan = "\x1b[36m"; reset = "\x1b[0m"; } console.warn(red + "CoffeeScript has moved!" + reset + " Please update references to " + yellow + "\"coffee-script\"" + reset + " to use " + yellow + "\"coffeescript\"" + reset + " (no hyphen) instead."); console.warn("Also, a new major version has been released under the " + yellow + "coffeescript" + reset + " name on NPM. This new release targets modern JavaScript, with minimal breaking changes. Learn more at " + cyan + "http://coffeescript.org" + reset + "."); console.warn(""); }'

[eval]:1
'if
^^^

SyntaxError: Invalid or unexpected token
    at createScript (vm.js:74:10)
    at Object.runInThisContext (vm.js:116:10)
    at Object.<anonymous> ([eval]-wrapper:6:22)
    at Module._compile (module.js:573:30)
    at evalScript (bootstrap_node.js:435:27)
    at startup (bootstrap_node.js:139:9)
    at bootstrap_node.js:578:3
```
i guess it's because of this 
https://www.npmjs.com/package/coffee-script
Now coffee-script is coffeescript

timfcsm:
same error with node 6.11.0

a062670:
CoffeeScript has moved!

https://www.npmjs.com/package/coffee-script

agogex:
+1 The same problem. How to fix it?

lihua123569:
how to fix??

posva:
It's something that has to be fixed in a dependency of one of our dependencies. Someone already created a PR for it at https://github.com/jonschlinkert/gray-matter/pull/56

So we'll wait for it to be released and then reinstalling vue-cli would reinstall the updated version of gray-matter

Ideally, we should also ask https://github.com/segmentio/metalsmith/ to update to that version and finally, update the version in the package.json file here

posva:
[...]

PR #667, opened 2017-12-05
Changed coffee-script to coffeescript in the package.json

tessaSAC:
I recently installed vue-cli on a new machine and noticed a warning about the `coffee-script` package name being deprecated in favor of `coffeescript`.

(It's also in `gray-matter` but I left that one alone).


posva:
Thanks for the PR but this is intentional see #655

tessaSAC:
Ahh oops! Thanks @posva!!

PR #672, opened 2017-12-07
merge mod to master

laohonglan:
1. last fix is to make keras.min.js a local dependency instead of node_module
2. keep only mnist basic conv


laohonglan:
this pull is not on purpose,  i was going to diff the forked master, but somehow diff to the original author's master,  sorry...

PR #673, opened 2017-12-07
check in packages.json for windows

laohonglan:


posva:
Hey thanks for your interest in contributing to Vue!
We cannot accept a PR like this because it's modifying too many things 😄 
Next time make sure to also describe a bit better what you're trying to achieve so we can understand your intentions and help you out with your PR

PR #679, opened 2017-12-11
Rename coffee-script to coffeescript

Tabrizian:
More information about this rename in here:
https://github.com/jashkenas/coffeescript/issues/4714

This will remove the warning that appears upon installation to rename coffee-script to coffeescript


erickwilder:
I don't know what am I missing here but do `vue-cli`still uses `coffeescript`? I couldn't find a single reference in the code base besides the `package.json` dependency. A simple search on the repo also didn't help:

https://github.com/vuejs/vue-cli/search?utf8=%E2%9C%93&q=coffee&type=

Maybe some of the core maintainers could enlighten me. Otherwise, I truly think it's better to just remove this dependency.

yyx990803:
This is from a **nested dependency**. We pinned it because old coffeescript had a behavior that breaks all installs. The warning does not affect any usage. The next version of `vue-cli` will not have this problem.

PR #700, opened 2017-12-31
Update coffeescript package name

corburn:
```
npm WARN deprecated coffee-script@1.12.7: CoffeeScript on NPM has moved to "coffeescript" (no hyphen)
```


rigor789:
This has been already done in #679 ?

mclassdesigns:
FYI: I am still getting the deprecation warning when I installed today.

yyx990803:
This is from the nested dependency so there's no way for us to get rid of that warning. It doesn't cause any real problem though.

PR #701, opened 2018-01-03
Update Node.js support

smalllong:
There is a 'let' expression outside of the strict mode in the bin/vue-init file, which makes Node.js(4.x) unsupported.


posva:
Duplicate of #634

Issue #708, opened 2018-01-08
Add '--force' flag to 'vue init' command


codetheorist:
**Problem:**

When testing templates, it's a frequent cause of headaches when testing. 

Currently in the template test.sh scripts, they use the `yes` command piped to the `vue init`, which doesn't work.  

**Solution**

To workaround this, we could add a `--force` flag that will send a new line character to each inquirer prompt.

We could use the `mock-stdin` package for sending the input to the prompt.

There is a PR coming shortly, fully tested with the Vue Webpack template using both the `--force` and the `-f` flag.

LinusBorg:
Imho we should rather provide a way to skip the inquirer prompt completely and load some predefined answers from some .json file or something. Kind of what I implemented in https://github.com/vuejs-templates/webpack/pull/1218

Shiyinq:
when i type vue init webpack shi . i get error like this 

`vue init webpack shi
C:\Users\shiyinq\AppData\Roaming\npm\node_modules\vue-cli\node_modules\is\index.js:1
(function (exports, require, module, __filename, __dirname) { ��Y�6�P��h�����Ѐ���W�:��V�p��������؀
                                                              ^

SyntaxError: Invalid or unexpected token
    at createScript (vm.js:80:10)
    at Object.runInThisContext (vm.js:139:10)
    at Module._compile (module.js:599:28)
    at Object.Module._extensions..js (module.js:646:10)
    at Module.load (module.js:554:32)
    at tryModuleLoad (module.js:497:12)
    at Function.Module._load (module.js:489:3)
    at Module.require (module.js:579:17)
    at require (internal/module.js:11:18)
    at Object.<anonymous> (C:\Users\shiyinq\AppData\Roaming\npm\node_modules\vue-cli\node_modules\metalsmith\lib\index.js:6:10)`

yyx990803:
The new version in `next` branch supports that:

```
$ vue create --help

  Usage: create [options] <app-name>

  create a new project powered by vue-cli-service


  Options:

    -s, --saved                      Skip prompts and use saved config
    -d, --default                    Skip prompts and use default config
    -c, --config <json>              Skip prompts and use inline JSON string as config
    -r, --registry <url>             Use specified NPM registry when installing dependencies
    -m, --package-manager <command>  Use specified NPM client when installing dependencies
    -f, --force                      Overwrite target directory if it exists
    -h, --help                       output usage information
```

deniszavadskiy:
[...]

PR #709, opened 2018-01-08
Issue #708: Add '--force' flag to 'vue init' command

codetheorist:
* Add `mock-stdin` package as a dependency
* Add `mock-stdin` to `ask.js` file
* Add `--force` flag in `bin/vue-init`
* Add conditional to send newline character at each prompt of `--force` flag is set

### DIFF
--- a/bin/vue-init
+++ b/bin/vue-init
@@ -27,12 +27,12 @@ program
   .usage('<template-name> [project-name]')
   .option('-c, --clone', 'use git clone')
   .option('--offline', 'use cached template')
-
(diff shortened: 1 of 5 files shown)


posva:
`force` is probably not the right naming choice. Something like `--default` (next version) or `--yes` seems more appropriate

codetheorist:
Now you've mentioned it, it does seem more logical @posva for it to be called `default` or `yes`.

yyx990803:
We will no longer be releasing new features to the 2.x branch, but thanks for contribution.

Issue #717, opened 2018-01-10
RFC: PWA Plugin


yyx990803:
/cc @addyosmani @jeffposnick @sudo-suhas

So, I've been working pretty hard on the next version of `vue-cli`, which uses a completely plugin-based architecture, and the PWA functionalities are now fully isolated in its own module: https://github.com/vuejs/vue-cli/tree/next/packages/%40vue/cli-plugin-pwa

A brief intro for the plugin API: `index.js` is the runtime plugin that will be loaded when users run `npm run serve` or `npm run build`. `generator` contains a module that will be invoked when the user scaffolds the project with the global `vue` command. The API should be pretty explanatory, and seems to cover most use cases. But any feedback is greatly appreciated.

Our previous PWA template is a fork of the main webpack template and has trouble keeping up with the upstream changes. This new setup should greatly reduce the maintenance scope for PWA-related features in the future.

After the launch of `vue-cli` 3.0, we should deprecate the pwa template and move all collaborations under the plugin instead.

sudo-suhas:
Hey @yyx990803, I wanted to test this out so I cloned the repo and tried to follow the steps in **"Development Setup"** from the readme but `yarn link` did not work correctly for me and the `vue` binary did not get installed globally. But I worked around that by doing this:

```
λ cd packages\test\

λ node ..\@vue\cli\bin\vue create test-app
```

However, after I manually selected the features, it failed to scaffold the project:

```
Vue CLI v3.0.0-alpha.1
✨  Creating project in E:\Projects\repos\vue-cli\packages\test\test-app.
🗃  Initializing git repository...
 ERROR  command failed: yarn add --dev @vue/cli-plugin-babel @vue/cli-plugin-eslint @vue/cli-plugin-unit-jest @vue/cli-plugin-pwa @vue/cli-service
error An unexpected error occurred: "http://registry.npmjs.org/@vue%2fcli-plugin-babel: Not found".
```

What do I need to do to make this work? 

Another question I had was, what about the documentation? Would each plugin have it's own documentation? Right now, the documentation for the PWA template is mostly similar to the webpack template but there is some additional documentation which touches upon things like service worker.

Is there anything else that you'd like me to help with?

addyosmani:
I'm excited about this direction of encapsulating PWA functionality into a vue-cli plugin. I've handed maintainace of the PWA template to Jeff but am supportive of deprecating it in favor of the plugin once we can confirm feature parity. 

[...]

PR #769, opened 2018-02-03
feat: Use the Workbox webpack plugin in pwa template

jeffposnick:
R: @yyx990803 & co.

See #717 for context.

Open question: should I also add something to the `README.md` for the template explaining which configuration options are supported, and how they map to the Workbox options? If so, is there an example of another template that has this written up in the format you're looking for?

### DIFF
--- a/packages/@vue/cli-plugin-pwa/index.js
+++ b/packages/@vue/cli-plugin-pwa/index.js
@@ -2,25 +2,39 @@ module.exports = (api, options) => {
   api.chainWebpack(webpackConfig => {
     const name = api.service.pkg.name
 
+    const userOptions = options.pwa || {}
+
     // the pwa plugin hooks on to html-webpack-plugin
     // and injects icons, manifest links & other PWA related tags into <head>
     webpackConfig
       .plugin('pwa')
         .use(require('./lib/HtmlPwaPlugin'), [Object.assign({
           name
-        }, options.pwa)])
+        }, userOptions)])
 
     // generate /service-worker.js in production mode
     if (process.env.NODE_ENV === 'production') {
-      webpackConfig
-        .plugin('sw-precache')
-          .use(require('sw-precache-webpack-plugin'), [{
-            cacheId: name,
-            filename: 'service-worker.js',
-            staticFileGlobs: [`${options.outputDir}/**/*.{js,html,css}`],
-            minify: true,
-            stripPrefix: `${options.outputDir}/`
-          }])
+      // Default to GenerateSW mode, though InjectManifest also might be used.
+      const workboxPluginMode = userOptions.workboxPluginMode || 'GenerateSW'
+      const workboxWebpackModule = require('workbox-webpack-plugin')
+      if (workboxPluginMode in workboxWebpackModule) {
+        const workBoxConfig = Object.assign({
+          cacheId: name,
+          exclude: [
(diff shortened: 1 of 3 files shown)


packages/@vue/cli-plugin-pwa/index.js
```diff
@@ -2,25 +2,39 @@ module.exports = (api, options) => {
...
+            new RegExp('favicon\.ico$'),
+            new RegExp('manifest\.json$')
+          ]
+        }, userOptions.workboxOptions)
+
+        webpackConfig
+          .plugin('workbox')
+          .use(workboxWebpackModule[workboxPluginMode], [workBoxConfig])
+      } else {
+        throw new Error(`${workboxPluginMode} is not a supported Workbox webpack plugin mode. ` +
```
jeffposnick: I wasn't sure what the philosophy was around misconfiguration. If there's something that should be done here other than throwing an `Error`, I'm happy to change.

yyx990803: Since this is a built-in plugin, you can add it the user options validation schema here: https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-service/lib/options.js#L3-L35

But throwing an error is also fine.

yyx990803:
Thanks @jeffposnick !

Yes, we will be documenting each plugin in its own README, but we haven't started finalizing the format yet. For now, feel free to provide any information you feel necessary, I'll be working on docs next week and revise it to make sure it's consistent.

packages/@vue/cli-plugin-pwa/index.js
```diff
@@ -2,25 +2,39 @@ module.exports = (api, options) => {
...
+      const workboxPluginMode = userOptions.workboxPluginMode || 'GenerateSW'
+      const workboxWebpackModule = require('workbox-webpack-plugin')
+      if (workboxPluginMode in workboxWebpackModule) {
+        const workBoxConfig = Object.assign({
+          cacheId: name,
+          exclude: [
+            new RegExp('\.map$'),
+            new RegExp('img/icons/'),
+            new RegExp('favicon\.ico$'),
+            new RegExp('manifest\.json$')
```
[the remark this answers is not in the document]
jeffposnick: The request that a browser makes for either a registered favicon or for the home screen icon bypasses the service worker, so there's no benefit from caching them.

jeffposnick:
Thanks! I'll add some docs to the README and push an additional commit.

packages/@vue/cli-plugin-pwa/index.js
jeffposnick: I ❤️ `joi`, but have found the default error messages to be a little hard to read, so I'll just stick with a custom `Error` for this specific failure.

jeffposnick:
PTAL.

yyx990803:
Thanks! 🎉

yyx990803:
@jeffposnick I noticed that even in the production build, the service worker pulls a dev version of workbox from the CDN. Is there anyway to configure which CDN link is used?

yyx990803:
Aha! I noticed it fetches the dev/prod version based on whether it's on localhost or not. Very nice!

jeffposnick:
(The dev/prod dynamic loading is all thanks to @gauntface!)

PR #775, opened 2018-02-04
Jest template: fix the name of the tested file 

dunglas:
Replace `Hello.vue` by `HelloWorld.vue`


posva:
I think you did the same PR twice 😄

dunglas:
Almost, I noticed later that there is the same "problem" for the Mocha template.

posva:
I mean it appears to be the exact same change on the same file (mocha plugin)

yyx990803:
This PR is changing the mocha plugin :)

I think you can just change both plugins in the same PR.

dunglas:
Done

PR #778, opened 2018-02-04
Swiper cause iOS 10 safari blank screen

fengerzh:
[Swiper](https://github.com/nolimits4web/swiper) is a widely used component. As iOS 10 safari does not support some ES6 syntax, such as a = b ** c, per https://github.com/nolimits4web/Swiper/issues/2263, we need to include swiper into babel, otherwise it will cause blank screen on iOS 10 safari.


yyx990803:
We don't want to whitelist things in core. This is configurable in `vue.config.js`:

``` js
module.exports = {
  chainWebpack: config => {
    config.rule('js').include.add(/node_modules\/(dom7|swiper)\/.*/)
  }
}
```

cdm1225:
add this config, but has an error
```
⠋  Building for production... ERROR  TypeError: config.rule is not a function
TypeError: config.rule is not a function
    at chainWebpack (/Users/wecash/work/official-abakus/vue.config.js:38:12)
    at webpackChainFns.forEach.fn (/Users/wecash/work/official-abakus/node_modules/@vue/cli-service/lib/Service.js:209:40)
```

change to 
```javascript
config.module.rule('js').include.add(/node_modules\/(dom7|swiper)\/.*/)
```

It works !

PR #779, opened 2018-02-05
Update README.md

jonlemmon:
Add warning about development branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,10 @@
 
 ## Install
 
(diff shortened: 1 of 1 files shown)


yyx990803:
Well, the branch does not really matter, since the old CLI is `vue-cli` and the new one is `@vue/cli` - they are separate packages, and the new one is also compatible with the old version when using `vue init`.

The new version also states its in alpha state - but we can probably be more specific on that.

Issue #787, opened 2018-02-05
Environment Variables


brunoseco:
How can I configure my environment variables in this new cli? For example, BASE_URL, ENDPOINT_API?

Thanks.

yyx990803:
Use `.env` files. Variations like `.env.[.mode][.local]` are also supported. Also, only variables starting in `VUE_APP_` will be available in the app.

There will be more documentation on this soon.

brunoseco:
Thanks dude,

But how can I replace BASE_URL?

LinusBorg:
> There will be more documentation on this soon.

As Evan just told you: Use a `.env` file. If you don't know what that is - it's a file that contains env variables just like you write them in your terminal:
```
BASE_URL=http://google.com/
```

yyx990803:
`BASE_URL` should be configured in `vue.config.js`, since it's also needed by the cli-service:

``` js
module.exports = {
  baseUrl: '/foo'
}
```

brunoseco:
> As Evan just told you: Use a .env file. If you don't know what that is - it's a file named .env, in your project's root directory, that contains env variables just like you write them in your terminal:

I know what is that.

Actually, .env file just allow variables that starts with VUE_APP_ ;)

sylvaincaillot:
Hello I just tried  something similar by setting up a .env.development and a .env.production files in my app. My .env.development file looks like:

```
NODE_ENV: '"development"'
VUE_APP_API: '"http://localhost:2096/api/"'
VUE_APP_DEBUG=true
```
As specified in the docs, I have used variables starting with `VUE_APP_`

However, if i do
```
console.log(process.env.VUE_APP_API)
```
it returns undefined  while it should return http://localhost:2096/api/

Any thoughts?

antonelepfl:
Hello, 
Regarding to the ENV I would like to use in the app some variables from **Jenkins** for example.

When in `vue.config.js` I do `console.log('Commit:', process.env.GIT_COMMIT)` I get it **correctly**.

In `.vue` files `process.env` is **only** replaced by 
```javascript
Object({
   NODE_ENV: "production"
})
```
I read about the `.env` but I don't know how to put _variables_ there. What is the correct way to do it?

phoet:
just FYI

antonelepfl:
Thanks @phoet in the end I made it work with the change below in `vue.config.js`

``` javascript
module.exports = {
  configureWebpack: {
    ...,
    plugins: [
[...]

PR #796, opened 2018-02-07
Add example of baseUrl in vue.config.js

antonelepfl:
It took me time until I found https://github.com/vuejs/vue-cli/issues/787
So probably add the extra params in the readme could be and easy way.
I perfectly understand if you don't want to included. It's just a suggestion.
If you find a better way I will be happy to contribute.

Thanks for the hard work.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,6 +93,9 @@ module.exports = {
   configureWebpack: {
     // ...
   }
+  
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks, the README for now is just a placeholder. We will cover everything in details in the final docs.

PR #798, opened 2018-02-07
feat: allow access to dev server through reverse proxy

Toilal:
`disableHostCheck` should be set to `true` to allow access to dev server through a reverse proxy.


yyx990803:
You can already configure this via the `devServer` object in `vue.config.js`. The options will be passed on to webpack-dev-server.

PR #829, opened 2018-02-12
fix(request): Change axios lib for request

### LINKED ISSUE #785: Hangs behind HTTP Proxy
### Version
3.0.0-alpha.8

### Steps to reproduce
Run vue create on a computer configured with a corporate HTTP Proxy

(`http_proxy` and `https_proxy` environment variable are defined, everything else is working properly)

### What is expected?
It should run as usual

vue create command hangs forever with no prompt at all.

---
Tested on Windows and Ubuntu.

wadetandy:
Closes #785

Axios has a [known issue](https://github.com/axios/axios/issues/658)
that causes requests to hang when accessing an HTTPS resource via a
proxy served over HTTP.  This changes out the axios dependency for the
[request](https://github.com/request/request) library. In order to keep
`async/await` conventions easy, I also dropped in the
[request-promise-native](https://github.com/request/request-promise-native)
dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,7 +42,6 @@
     ]
   },
   "devDependencies": {
-    "axios": "^0.17.1",
     "babel-core": "^7.0.0-0",
     "debug": "^3.1.0",
     "eslint": "^4.16.0",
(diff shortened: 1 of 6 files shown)


wadetandy:
Whoops a typo after running tests broke them. Fixing.

PR #830, opened 2018-02-12
Add missing "EOT" (upper case) font extension.

pbastowski:
I have EOT (upper cased) fonts in my project, which currently cause the compilation process to fail with vue-cli 3 alpha 11. This PR fixes the problem.

### DIFF
--- a/packages/@vue/cli-service/lib/config/base.js
+++ b/packages/@vue/cli-service/lib/config/base.js
@@ -79,7 +79,7 @@ module.exports = (api, options) => {
 
     webpackConfig.module
       .rule('fonts')
-        .test(/\.(woff2?|EOT|eot|ttf|otf)(\?.*)?$/)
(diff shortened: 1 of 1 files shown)


yyx990803:
Can't you change the extension to lowercase? I think that's the more common practice...

pbastowski:
Sure, that's possible. I just didn't realise that it's common practice with EOT font files, and I didn't find any advice with google about that. Also, my previous webpack config specifically mentioned both eot and EOT. 
But, yeah, it makes sense :) I'll rename them.
Thanks for your help.

pbastowski:
I'm fine if you close this PR. But if you decide to merge it, it may save someone an hour or two of searching for why one of their fonts just doesn't load and breaks compilation when they move to vue-cli 3.

pbastowski:
Thank you.

PR #835, opened 2018-02-14
Fixed update info's formatting.

RatanShreshtha:
Before the prompt for new available update looked like this

┌─────────────────┐
│ ✨  Update available: 3.0.0-alpha.13 ✨  │
└─────────────────┘

After 

┌──────────────────┐
│ ✨  Update available: 3.0.0-alpha.13 ✨  │
└──────────────────┘

### DIFF
--- a/packages/@vue/cli/lib/util/clearConsole.js
+++ b/packages/@vue/cli/lib/util/clearConsole.js
@@ -16,9 +16,9 @@ module.exports = async function clearConsoleWithTitle (checkUpdate) {
   }
   if (checkUpdate && semver.gt(latest, current)) {
     title += chalk.green(`
-┌─────────────────────────${`─`.repeat(latest.length)}─┐
+┌─────────────────────────${`─`.repeat(latest.length + 2)}─┐
 │ ✨  Update available: ${latest} ✨  │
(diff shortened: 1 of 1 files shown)


yyx990803:
What OS / shell / terminal are you using? It looks fine in macOS/bash...

RatanShreshtha:
I am using fish on fedora 27.

RatanShreshtha:
Can anyone tell me how can I test this and make sure that it's not a shell / terminal issue.

chrisvfritz:
@RatanShreshtha Emoji can sometimes do strange things with length counts, such as:

``` js
'🙏🏻'.length //=> 4
```

In the above scenario, it's because the emoji is actually a ligature. I wonder if something similar is happening, like maybe `✨` isn't fully supported on the system so not included in the length? Total stab in the dark, but you could try `'✨'.length` to see what you get.

yyx990803:
Or we can just remove the emoji to avoid the headache ;)

RatanShreshtha:
I have tried `'✨'.length` in node repel and its giving me `1` as the output.

chrisvfritz:
Hm, would you mind trying to remove the emoji locally anyway and seeing if that fixes it?

PR #843, opened 2018-02-16
Allow setting `useBuiltIns` to be `false`.

zigomir:
Because if you set `useBuiltIns` to false we'll always fallback to `'usage'`.


LinusBorg:
@zigomir @yyx990803 Where do I set this option? It's not in the schema for `vue.config.js` as pluginOptions for `babel`?

zigomir:
👍 I'm using this in `.babelrc` file yes, and you're not stupid :)

PR #857, opened 2018-02-18
Output help information on unknown CLI commands

### LINKED ISSUE #849: Default output for CLI with unknown commands
If we invoke CLI with unknown commands from terminal, the CLI just ignore and exit.
For example:

```bash
vue something
```

It will output nothing.

How about output the help information and tell users that the entered `something` is not a correct command or is an unknown command? It's just a small enhancement for user experience though.

hsxfjames:
resolves #849.
I read quick through the dependency `commander.js`'s README and source code, finding that it hasn't implemented this yet. Here is a workaround, but I am not sure if it has a side effect.

### DIFF
--- a/packages/@vue/cli/bin/vue.js
+++ b/packages/@vue/cli/bin/vue.js
@@ -87,6 +87,15 @@ program
     loadCommand('init', '@vue/cli-init')
   })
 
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks!

PR #886, opened 2018-02-26
Don't swallow vue.config.js errors (fixes #874)

### LINKED ISSUE #874: Errors in vue.config.js are swallowed
### Version
3.0.0-alpha.13

### Reproduction link
[https://github.com/chrisvfritz/vue-enterprise-boilerplate](https://github.com/chrisvfritz/vue-enterprise-boilerplate)

### Steps to reproduce
Add a syntax error or throw an error from `vue.config.js`. 

### What is expected?
I would expect Vue CLI commands that require `vue.config.js` to fail loudly when it contains an error, making me aware of the source of the problem (or even that there _is_ a problem).

Vue CLI will swallow it and continue as if everything is normal. I noticed this behavior when Webpack was acting as if my aliases were not defined. It took a little while to figure out that there was a syntax error in a file required from `vue.config.js`.

chrisvfritz:
I assumed the try/catch was only to avoid an error if `vue.config.js` didn't exist. This should accomplish the same, but still provide useful error output in cases of a syntax or runtime error in this file.

### DIFF
--- a/packages/@vue/cli-service/lib/Service.js
+++ b/packages/@vue/cli-service/lib/Service.js
@@ -160,7 +160,7 @@ module.exports = class Service {
       process.env.VUE_CLI_SERVICE_CONFIG_PATH ||
       path.resolve(this.context, 'vue.config.js')
     )
-    try {
+    if (fs.existsSync(configPath)) {
       fileConfig = require(configPath)
       if (!fileConfig || typeof fileConfig !== 'object') {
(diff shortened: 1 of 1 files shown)


yyx990803:
There are some test related quirks for this change so I ended up doing it differently in a few commits starting at 14a2dc725977f8cda7c2068390ab6955a0646dd8 😅

PR #903, opened 2018-02-28
Fixed table of contents

johnpyp:
Building as a Library or Web Component was singular in the table of contents and  plural in its title (#### Building as Library or Web Components). Changed both to be singular


yyx990803:
Thanks for the PR. Technically you can build a single project into multiple web components, so that was intentional.

johnpyp:
Sure, only noticed it because the table of contents navigation wasn't directing to it. Would it be better to change both to plural?

PR #915, opened 2018-03-01
feat(cli-service): Add CSS Module localIdentName to vue config

n-zeplo:
Changed the vue config to take a user input option for CSS module localIdentName and default back to initial localIdentName ([name]_[local]__[hash:base64:5])


yyx990803:
Thanks, great job!

PR #925, opened 2018-03-03
fix: fix vue-cli-service build

koen-chen:
add --silent option for ignore printing status log


yyx990803:
Thanks, this is actually an internal flag used only during testing. It's not meant to be a public option.

PR #950, opened 2018-03-07
Make experimentalCompileTsWithBabel compatible with vue.config.js

elevatebart:
The schema of `vue.config.js` is very specific in what it can accept.
For instance, it can accept custom options only in the `pluginOptions` param.
I put the `experimentalCompileTsWithBabel` back in it's place

### DIFF
--- a/packages/@vue/cli-plugin-typescript/__tests__/tsPluginExpDecorators.spec.js
+++ b/packages/@vue/cli-plugin-typescript/__tests__/tsPluginExpDecorators.spec.js
@@ -1,0 +1,14 @@
+jest.setTimeout(30000)
+
+const { assertServe, assertBuild } = require('./tsPlugin.helper')
+
+const options = {
+  plugins: {
+    '@vue/cli-plugin-typescript': {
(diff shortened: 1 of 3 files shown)


yyx990803:
Thanks for the PR - however, it is currently intentionally omitted from the main schema because this is by no means stable yet. We are waiting on the Babel side to stabilize before exposing this.

elevatebart:
For now, if one chooses to use the experimental decorators, she will get an error.
Please try choosing options `TypeScript > experimental decorators` and then try to run `yarn serve`.
This PR makes it work without changing the `vue.config.js`.

yyx990803:
I don't understand. The experimental flag has nothing to do with decorators and it's not exposed in CLI options. If you are talking about the class-syntax components, that seems to work as intended. Either way, if there is a bug please open an issue with reproduction.

elevatebart:
Ooooops! I am so sorry, I guess writing pull request at 10 PM is not good for me.
I completely missed the `process.env.VUE_CLI_EXPERIMENTAL?`. 
Brain fart. 
Thank you for your time.

Issue #992, opened 2018-03-16
DLL mode bundle sizes


alexdilley:
### Version
3.0.0-beta.6

### Reproduction link
[https://codesandbox.io/s/ppo0y403lj](https://codesandbox.io/s/ppo0y403lj)

### Steps to reproduce
Enable DLL mode by setting the `dll` option in `vue.config.js` to `true` (as per [docs](https://github.com/vuejs/vue-cli/blob/dev/docs/cli-service.md#dll-mode)):

```js
// vue.config.js
module.exports = {
  dll: true
}
```

### What is expected?
Vendor libs not to be included in `app` bundle; bundle sizes to be reduced.

### What is actually happening?
With the option set to `true`, some `node_modules` are packaged into the `app` bundle – the opposite of a build with the option disabled, and hence the opposite of the option's function described by the documentation.

Comparison of build sizes given the following base project set-up:

```sh
vue create -i '{"plugins": { "@vue/cli-plugin-babel": {} }, "router": true, "vuex": true}' my-project
```

`dll: false` (the default):

```
  File                          Size                   Gzipped

  dist/js/vendor.456df753.js    95.01 kb               32.36 kb
  dist/js/app.be870e84.js       12.51 kb               8.05 kb
  dist/css/app.8b290d57.css     0.42 kb                0.26 kb

  Images and other types of assets omitted.
```

`dll: true`:

```
  File                          Size                   Gzipped

  dist/js/vendor.3884c493.js    95.12 kb               32.29 kb
  dist/js/app.d8cf8ae4.js       13.89 kb               8.71 kb
  dist/css/app.8b290d57.css     0.42 kb                0.26 kb

  Images and other types of assets omitted.
```

The difference is exacerbated as projects grow.

---
Is [this](https://github.com/vuejs/vue-cli/blob/dev/packages/@vue/cli-service/lib/config/app.js#L61) condition the opposite of what it should be?

Unless I'm mistaken in how this feature should work, the consequence of changing/fixing this will be that all vue-cli generated apps' bundle sizes will subsequently increase, except those where the option has already been set to `true` in the anticipation it would reduce bundle sizes (which it isn't currently), in which case future builds _will_ be smaller :)

LinusBorg:
codesandbox does not use vue-cli 3 yet, so adding a vue.config.js file wo't have any effect, so this is not a reproduction we can work with. I assume you created it to have a link to share?

[...]

PR #1002, opened 2018-03-17
docs: clarify function of DLL mode

alexdilley:
#992


yyx990803:
ref https://github.com/vuejs/vue-cli/issues/992#issuecomment-373958362

yyx990803:
Closing as we are planning to upgrade to webpack 4, and due to the perf improvements in webpack 4, we will remove built-in DLL support as it's providing less value and no longer worth the extra complexity/maintenance cost.

PR #1004, opened 2018-03-19
exclude `node_modules` .

yibuyisheng:


Akryum:
Hi, thanks for your PR. Why did you need to make this change?

yibuyisheng:
@Akryum 

```js
const minimatch = require('minimatch');

console.log(minimatch('use/node_modules/aaa/bbb/a.js', '**node_modules**'));     // false
console.log(minimatch('use/node_modules/aaa/bbb/a.js', '**/node_modules/**'));   // true
```

`**node_module**` does not match `use/node_modules/aaa/bbb/a.js`.

PR #1016, opened 2018-03-20
feat(e2e nightwatch): allow setting cli `config` option.

aitboudad:
Hi,
This allow to run e2e with a custom config file `vue-cli-service e2e --config nightwatch.config.js` instead of using the internal Nightwatch config.

My use-cases is using e2e nightwatch with a custom integration like [nightwatch-cucumber](https://github.com/mucsi96/nightwatch-cucumber) which seems not working with the default behavior.

```js
// nightwatch.config.js
require('nightwatch-cucumber')({}) // is ignored when using the internal Nightwatch config.
module.exports = {
  ...
}
```

Thanks!

### DIFF
--- a/packages/@vue/cli-plugin-e2e-nightwatch/index.js
+++ b/packages/@vue/cli-plugin-e2e-nightwatch/index.js
@@ -31,17 +31,20 @@ module.exports = (api, options) => {
     return serverPromise.then(({ server, url }) => {
       // expose dev server url to tests
       process.env.VUE_DEV_SERVER_URL = url
-      // expose user options to config file
-      const fs = require('fs')
-      let userOptionsPath, userOptions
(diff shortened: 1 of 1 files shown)


ArielleA:
Hi, 

Having the --config file option is great, but doesn't it cause that you eject? I have the same usecase for cucumber but would prefer not to keep the whole nightwatch config in my tree, only the cucumber snippet. 

i.e. nightwatch.config.js
`
require('nightwatch-cucumber')({
  cucumberArgs: [
    '--require', 'test/e2e/step_definitions',
    '--require', 'test/e2e/steps',
    'features',
  ],
});
`

aitboudad:
@ArielleA to keep the whole nightwatch config use:
```js
const deepmerge = require('deepmerge');
...

const config = require('@vue/cli-plugin-e2e-nightwatch/nightwatch.config.js');
module.exports = deepmerge(config, {
  ...
});
```

PR #1025, opened 2018-03-22
Minor doc change:  Show example in Advanced Webpack config section

gnulnx:
Even though this should have been obvious I had a bit of trouble figuring this out today so I figured I'd update the docs to help others.  

I wasn't sure if it made sense to leave the original generic example in place or remove it... so I left it in place and added example below.   I'm more than happy to update as necessary based on feedback.

Thanks

### DIFF
--- a/docs/webpack.md
+++ b/docs/webpack.md
@@ -93,7 +93,9 @@ module.exports = {
 }
 ```
 
-For example say you want to change the default location of index.html from */Users/username/proj/public/index.html* to */Users/username/proj/app/templates/index.html*.  By referencing [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin#options) you can see a list of options you can pass in. To change our template path we simply pass in a new template path like:
+You will need to familiarize yourself with [webpack-chain's API](https://github.com/mozilla-neutrino/webpack-chain#getting-started) and [read some source code](https://github.com/vuejs/vue-cli/tree/dev/packages/%40vue/cli-service/lib/config) in order to understand how to leverage the full power of this option, but it gives you a more expressive and safer way to modify the webpack config than directly mutation values.
+
+For example, say you want to change the default location of index.html from */Users/username/proj/public/index.html* to */Users/username/proj/app/templates/index.html*.  By referencing [html-webpack-plugin](https://github.com/jantimon/html-webpack-plugin#options) you can see a list of options you can pass in. To change our template path we can pass in a new template path with the following config:
 
 ``` js
 // vue.config.js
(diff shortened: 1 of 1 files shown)


docs/webpack.md
```diff
@@ -93,6 +93,28 @@ module.exports = {
...
+
+``` js
+// vue.config.js
+module.exports = {
+  chainWebpack: config => {
+    config
+      .plugin('html')
+      .tap(args => {
+        return [{
+          template: '/Users/username/proj/app/templates/index.html'
```
yyx990803: The change should extend the existing options (`args[0]`) so that it doesn't discard other possible options that were already defined.

gnulnx: Hi Evan I think I'm being a bit dense about this.  I've tried both of these.

```
module.exports = {
  chainWebpack: config => {
    config
      .plugin('html')
      .tap(args => {
        args[0] = {
            template: '/Users/jfurr/zenperf/public2/index.html'
        }
        return args
      })
  }
}
```

And 

```
module.exports = {
  chainWebpack: config => {
    config
      .plugin('html')
      .tap(args => {
        return [{template: '/Users/jfurr/zenperf/public2/index.html'}]
      })
  }
}
```

In both cases the output results in the HtmlWebackPlugin options dropping ```env```, though the rest of the options seem to be unmodified.

Before modification
```
/* HtmlWebpackPlugin */ {
      options: {
        template: '/Users/jfurr/zenperf/public/index.html',
        filename: 'index.html',
        hash: false,
        inject: true,
        compile: true,
        favicon: false,
        minify: false,
        cache: true,
        showErrors: true,
        chunks: 'all',
        excludeChunks: [],
        title: 'Webpack App',
        xhtml: false,
        env: {
          NODE_ENV: 'development',
          BASE_URL: '/'
        }
      }
    },
```

After mod the template is correct, but ```env``` is missing..

```
/* HtmlWebpackPlugin */ {
      options: {
        template: '/Users/jfurr/zenperf/public2/index.html',
        filename: 'index.html',
        hash: false,
        inject: true,
        compile: true,
        favicon: false,
        minify: false,
        cache: true,
        showErrors: true,
        chunks: 'all',
        excludeChunks: [],
        title: 'Webpack App',
        xhtml: false
      }
    },
```

Am I missing some obvious?  Do I need to explicitly set the env back to the default?

abruere: @yyx990803 probably refers to the following way to extend:

```
module.exports = {
  chainWebpack: config => {
    config
      .plugin('html')
      .tap(args => {
        return [args[0], {template: '/Users/jfurr/zenperf/public2/index.html'}]
      })
  }
}
```

This should preserve `env` and just update `template` option. See https://github.com/mozilla-neutrino/webpack-chain#config-plugins-modify-arguments

gnulnx:
@abruere Thanks and that makes total sense now.   PR updated.

PR #1031, opened 2018-03-24
test: fix "Protocol error" error when running test in vscode terminal

xsbchen:
fix "Protocol error (Page.navigate): Cannot navigate to invalid URL undefined" error when running test in vscode terminal(zsh)


elevatebart:
I get this issue too  but your branch does not seem to pass appveyor ;)

elevatebart:
This failure looks like some false positives due to a timeout in a test and a too short delay in between first load and hmr refresh. `sleep(1000)` does not seem to be wide enough.

Issue #1048, opened 2018-03-29
vue invoke could be able to rename files


elevatebart:
### What problem does this feature solve?
When using a plugin immediately at create, files can be renamed or deleted. 
When using `vue invoke` or `vue add` plugins cannot. 
This makes creation of plugin this much harder.

### What does the proposed API look like?
A couple of changes on the Generator.js files to be able to compare file trees before and after plugin would allow the `@vue/cli/lib/util/writeFileTree.js` to delete the file that have been renamed.

I made a quick PR to propose a solution. You can find an illustration of the problem I solve [here](https://github.com/elevatebart/vuecliinvoke)

yyx990803:
Closed via #1049

PR #1049, opened 2018-03-29
feat(invoke): allow invoke to replace/rename/delete files

elevatebart:
If a plugin is used within the vue create command,
It will be able to delete and rename files.
This is enabling the same tools for invoking 3rd party plugins.

#1048 is solved in this PR


yyx990803:
Thanks! I consider this a fix since that should be the intended behavior.

PR #1066, opened 2018-04-03
css.extract is enable when  no Prod

lusess123:
css.extract is enable when  isDev


yyx990803:
The user should be able to explicitly disable extraction.

PR #1083, opened 2018-04-06
Pass all parameters to onProxyReq in proxy configuration

davidhemphill:
In the current version of vue-cli, one is able to use the `onProxyReq` callback to modify the request. This is useful for adding things such as client IDs and secrets for OAuth when one can't keep them secure client-side.

```js
module.exports = {
  target: process.env.PROXY_URL,
  onProxyReq(proxyReq, req, res) {
    console.log(proxyReq, req, res)
    if (req.path == '/oauth/token') {
      req.body.client_id = process.env.CLIENT_ID
      req.body.client_secret = process.env.CLIENT_SECRET
    }
  },
}
```

The currently tagged beta 6 doesn't use a user-provided `onProxyReq` callback. The dev branch does, but it does not pass all the parameters to it: 

```js
onProxyReq (proxyReq) {
  if (usersOnProxyReq) {
    usersOnProxyReq(proxyReq)
  }
  // ...
},
```

It would be ideal if we could keep this functionality. :-)

### DIFF
--- a/packages/@vue/cli-service/lib/util/prepareProxy.js
+++ b/packages/@vue/cli-service/lib/util/prepareProxy.js
@@ -79,7 +79,7 @@ module.exports = function prepareProxy (proxy, appPublicFolder) {
           )
         }
       },
-      onProxyReq(proxyReq, req, res) {
+      onProxyReq (proxyReq, req, res) {
         if (usersOnProxyReq) {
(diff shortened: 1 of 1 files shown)


davidhemphill:
Is there anything I can do to help you kind folks regarding this?

PR #1095, opened 2018-04-09
chore(UI): use computedKeywords to filter the search

Haroenv:
Both `keywords` and `computedKeywords` work, but `computedKeywords` will make sure this only matches for those with the correct name, not all with that keyword. 

Thanks!

cc @Akryum


Akryum:
Thanks for your help! 👍

PR #1116, opened 2018-04-11
Support custom initial commit message

dhensche:
I was not sure how to write a test for this feature, but I have tested it manually. Any input on how to write the test would be appreciated


yyx990803:
I think it's better to avoid yet another prompt when creating a new project when a CLI flag is available, so I've removed the prompt.

yyx990803:
FYI in c2aac9e I've simplified it to `-g, --git`

PR #1117, opened 2018-04-12
Add `camelCase` option for CSS Modules

sirlancelot:
Documentation: https://github.com/webpack-contrib/css-loader#camelcase


yyx990803:
Thanks, however we will likely remove the `css` config option altogether as we upgrade to vue-loader 15. In the future, just tap and modify `css-loader` options directly.

Issue #1076, opened 2018-04-05
Allow 3rd party plugins to extend configTransforms


ascendancyy:
### What problem does this feature solve?
Currently only official plugins can respect the `useConfigFiles` option when creating a project when defined in a preset (or manual mode). Right now 3rd party plugins have the option of making use of `extendPackage` or using `render` to write out the file manually. This would mostly be for 3rd party plugins to use `extendPackage` and not worry about where the config ends up.  

### What does the proposed API look like?
Adding a method like `addConfigTransform(key, transform)` to `GeneratorAPI` that adds transforms to an object on the `Generator` instance of the `GeneratorAPI`. Then when `extractConfigFiles` is eventually called the transforms can be resolved to `Object.assign({}, configTransforms, this.configTransforms)`. It'd be simple enough to prevent adding transforms for something like the vue config (or maybe all of the defaults).  

Would this be viable?

Issue #994, opened 2018-03-16
.eslintrc is deprecated


nicocevallos:
### Version
3.0.0-beta.6

### Reproduction link
*not-required*

### Steps to reproduce
Create a new Vue app activating ESLint.

### What is expected?
Vue CLI should create a `.eslintrc.json` or `.eslint.js` file, due `.eslintrc` file is deprecated according to the ESLint documentation: https://eslint.org/docs/user-guide/configuring#configuration-file-formats

### What is actually happening?
Vue CLI creates a `.eslintrc` file

---
Maybe the option to create a `.eslintrc.js` file will be better to place comments and solve conditional rules according to the `NODE_ENV` value

piehei:
.eslintrc.js would also be my pick for the stated reasons.

yangmingshan:
And Babel 7 support `.babelrc.js`, PostCSS support `postcss.config.js`

LinusBorg:
currently, the cli creates JSON, not Javascript code. JSON is also easier to augment for 3rd paty plugins, since you can simply require the file, change the updated object and save the JSON back to disc, so I would vote for `.eslintrc.json`

ascendancyy:
Just to note right now if someone were to do this the typescript plugin will change files that end in `.js` and don't contain `config` to a `.ts` file. If the config was a `.eslintrc.js` file it would end up as .eslintrc.ts which means the config file won't be found.  

So .eslintrc.json would probably be a better default in any case.

cj:
@ascendancyy the jest plugin is generating `jest.config.js` just fine with typescript https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-unit-jest/generator/index.js#L64.  My vote would be for `.eslintrc.js` for this reason https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-eslint/eslintOptions.js#L6-L7

PR #1130, opened 2018-04-14
feat(generator): allow plugins to modify how configs are extracted

ascendancyy:
This would help plugin developers that are working on plugins for libraries/framerworks/etc that can be configured in multiple places (a field in package.json, multiple files). Or users that want to create a personal plugin that they can add to a preset to quickly setup a project in the case that they would prefer a different configuration file type over the defaults (eg. `.babelrc.js` over `.babelrc`).  

Related: #994  

Closes: #1076


packages/@vue/cli/lib/GeneratorAPI.js
```diff
@@ -77,6 +77,24 @@ class GeneratorAPI {
...
 
+  /**
+   * Configure how config files are extracted.
+   *
+   * @param {string} key - Config key in package.json
+   * @param {ConfigTransform} transformer - A function that receives the
+   *   config object, whether to check for an existing config on disk, and the
+   *   current context.
+   */
+  addConfigTransform (key, configTransform) {
```
yyx990803: I think the current API would be a bit difficult to use as the plugin author would have to reimplement `makeXXXTransform` or require vue-cli internals (risks breaking).

The second argument could also accept an object in the form of

```
{
  file: '.baberc.js',
  type: 'js|json|multi'
}
```

pksunkara:
This PR looks good. @yyx990803 Any update on this? The author has addressed your review comments.

Akryum:
I have merged dev and refactored config transforms to a more flexible system similar to the one used by the ui. (Will eventually merge the two systems.)

PR #1135, opened 2018-04-15
Add .swp files to default .gitignore

robcresswell:
.swp files are temporary files generated when a file is open in VIM.
Let's ignore them like the other editors :)


ycmjason:
We should really use `*.sw*` as vim will use `.swo` or other ending chars if `.swp` already existed. I have submited the same request on the same day.

https://github.com/vuejs/vue-cli/pull/1133

Issue #1162, opened 2018-04-23
npm run serve error: You must pass the `"decoratorsLegacy": true` option to @babel/preset-stage-2


scliuyang:
### Version
3.0.0-beta.6

### Reproduction link
[https://codepen.io/](https://codepen.io/)

### Steps to reproduce
- Vue CLI v3.0.0-beta.6
- ? Please pick a preset: Manually select features
- ? Check the features needed for your project: TS, Router, Vuex, CSS Pre-processors, Linter
- ? Use class-style component syntax? Yes
- ? Use Babel alongside TypeScript for auto-detected polyfills? Yes
- ? Pick a CSS pre-processor (PostCSS, Autoprefixer and CSS Modules are supported by default): LESS
- ? Pick a linter / formatter config: Airbnb
- ? Pick additional lint features:
- ? Where do you prefer placing config for Babel, PostCSS, ESLint, etc.? In dedicated config files
- ? Save this as a preset for future projects? No

vue-cli will install @babel/preset-stage-2@7.0.0-beta.45 and npm run serve will throw error:

`The new decorators proposal is not supported yet. You must pass the `"decoratorsLegacy": true` option to @babel/preset-stage-2`

see:  https://github.com/babel/babel/blob/master/packages/babel-preset-stage-2/src/index.js#L29

forgive my poor English

### What is expected?
server success

### What is actually happening?
The new decorators proposal is not supported yet. You must pass the `"decoratorsLegacy": true` option to @babel/preset-stage-2

dzza:
Same. On a fresh install:

```
>yarn serve
yarn run v1.5.1
$ vue-cli-service serve --open
 INFO  Starting development server...
 94% asset optimization

 ERROR  Failed to compile with 1 errors                                                                                                         23:59:26

 error  in ./src/main.js

Module build failed: Error: [BABEL] /Users/dzza/multiselect/src/main.js: The new decorators proposal is not supported yet. You must pass the `"decoratorsLegacy": true` option to @babel/preset-stage-2 (While processing: "/Users/dzza/multiselect/node_modules/@vue/babel-preset-app/index.js$1")
    at /Users/dzza/multiselect/node_modules/@babel/preset-stage-2/lib/index.js:107:11
    at /Users/dzza/multiselect/node_modules/@babel/helper-plugin-utils/lib/index.js:18:12
    at /Users/dzza/multiselect/node_modules/@babel/core/lib/config/full.js:172:14
[...]

PR #1163, opened 2018-04-23
Fixing "The new decorators proposal is not supported yet."

tamatamvan:
Adding option `decoratorsLegacy: true` in `@babel/preset-stage-2` for fixing this error :
```
Module build failed: Error: [BABEL] /path/to/project/src/main.js: The new decorators proposal is not supported yet. You must pass the `"decoratorsLegacy": true` option to @babel/preset-stage-2 (While processing: "/path/to/node_modules/@vue/babel-preset-app/index.js$1")
```

### DIFF
--- a/packages/@vue/babel-preset-app/index.js
+++ b/packages/@vue/babel-preset-app/index.js
@@ -37,7 +37,8 @@ module.exports = (context, options = {}) => {
   // stage 2. This includes some important transforms, e.g. dynamic import
   // and rest object spread.
   presets.push([require('@babel/preset-stage-2'), {
-    useBuiltIns: true
+    useBuiltIns: true,
+    decoratorsLegacy: true
   }])
 
   // transform runtime, but only for helpers
--- a/packages/@vue/babel-preset-app/index.js
+++ b/packages/@vue/babel-preset-app/index.js
@@ -17,8 +17,7 @@ module.exports = (context, options = {}) => {
   const envOptions = {
     modules: options.modules || false,
     targets: options.targets,
-    useBuiltIns: typeof options.useBuiltIns === 'undefined' ? 'usage' : options.useBuiltIns,
-    decoratorsLegacy: options.decoratorsLegacy || false
+    useBuiltIns: typeof options.useBuiltIns === 'undefined' ? 'usage' : options.useBuiltIns
   }
   delete envOptions.jsx
   // target running node version (this is set by unit testing plugins)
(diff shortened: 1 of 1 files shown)


silkentrance:
@tamatamvan please let the user pass in the decoratorsLegacy option so that he/she can decide whether to use that option. See the proposal by @soft903 https://github.com/vuejs/vue-cli/issues/1162#issuecomment-383483712

tamatamvan:
hi, I'd made another change, so users can decide to use the option or not

packages/@vue/babel-preset-app/index.js
```diff
@@ -17,7 +17,8 @@ module.exports = (context, options = {}) => {
   const envOptions = {
     modules: options.modules || false,
     targets: options.targets,
-    useBuiltIns: typeof options.useBuiltIns === 'undefined' ? 'usage' : options.useBuiltIns
+    useBuiltIns: typeof options.useBuiltIns === 'undefined' ? 'usage' : options.useBuiltIns,
+    decoratorsLegacy: options.decoratorsLegacy || false
```
jkzing: Is this needed for `preset-env`? 
It seems `decoratorsLegacy` only affects stage-0/1/2.

hzoo: yeah it's only for stage 0,1,2. It's one of the unfortunate issues related to having stage presets at all since you might not even use decorators. This is why after a lot of discussion may just want to remove them since it's just an array of plugins that will constantly changing https://github.com/babel/babel/issues/7770.

light24bulbs:
We should also lock the version of babel so this doesn't happen again.  I was just talking to the Babel maintainers and they said that nobody should be using babel beta with a `^` because they WILL be breaking it more.  They don't use it internally with a `^`

QuantumInformation:
Just came across this in a brand new cli project

yyx990803:
Just a note... vue-cli itself is in beta too.

light24bulbs:
So it's good to be finding issues!

We can't use `^` with Babel beta, they said it will continue to break and
they aren't using it internally with a hat.  Semver doesn't work with betas.

BenoitAverty:
Will this PR alone fix the problem ? How would I pass the option needed to the file that was just modified ?

ryakstis:
I have a local config which looks like this:

```json
{
  "presets": [
    [
      "@vue/app",
      {
        "decoratorsLegacy": true
      }
    ]
  ]
}
```

This worked for me with the above patch.

QuantumInformation:
The above didn't work for me +     "@babel/preset-stage-2": "7.0.0-beta.44"
```
 ERROR  Failed to compile with 1 errors                                                                                            08:29:08

 error  in ./src/main.js

Module build failed: Error: [BABEL] /Users/nikos/WebstormProjects/quantumjs2/src/main.js: The new decorators proposal is not supported yet. You must pass the `"decoratorsLegacy": true` option to @babel/preset-stage-2 (While processing: "/Users/nikos/WebstormProjects/quantumjs2/node_modules/@vue/babel-preset-app/index.js$1")
```

QuantumInformation:
I upgraded to beta 7, deleted and reinstalled node mods, and it worked

Issue #1167, opened 2018-04-24
Cannot specify output chunk names


sanchitgn:
### Version
3.0.0-beta.6

### Reproduction link
[https://codesandbox.io/s/oxm4l6857y](https://codesandbox.io/s/oxm4l6857y)

### Steps to reproduce
1. Specify a chunk using dynamic import syntax 
2. Add a chunk name `import(/* webpackChunkName: "profile" */ "./views/Profile/Profile.vue")`
3. Build `yarn run build`

### What is expected?
```  
   File                                      Size             Gzipped

  dist/js/profile.e9349658.js               31.66 kb         5.56 kb
```

### What is actually happening?
``` 
 File                                      Size             Gzipped

  dist/js/0.e9349658.js                     31.66 kb         5.56 kb
```

---
Sorry if this is has already been posted. Could not find a similar issue.
The other webpack comments (`webpackMode`) work fine

Possible duplicate [#1097](https://github.com/vuejs/vue-cli/issues/1097)

LinusBorg:
You probably have to set `"comments": true` in the Babel config.

sanchitgn:
Setting `"comments": true` in `.babelrc` doesn't seem to work

LinusBorg:
...which is why I alredy removed my comment. I misread your OP.

sanchitgn:
@LinusBorg Check [this repo](https://github.com/sanchitgn/chunk). After specifying `webpackChunkName` in [router.js](https://github.com/sanchitgn/chunk/blob/master/src/router.js#L12:53), the output is still `0.[hash].js`. The project was created with the latest vue-cli. The only change it the dynamic `Home` component

LinusBorg:
I think this line is the problematic one:

https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-service/lib/config/prod.js#L8

Please try this in vue.config.js (can't do that myelf right now):
```javascript
module.exports {
  chainWebpack: config => {
    config.output.chunkFilename(`js/[name].[chunkhash:8].js`)

  }
}
```

If that solves the problem, we should change ths with a PR.

sanchitgn:
Yup, that solves it. It makes sense to add [name] in the chunk file name

zigomir:
Looks like this one can be closed.

doublemarked:
This doesn't work if you're building as `lib` or `wc`, because the lib/wc config is applied after the vue.config.js.

[...]

PR #1173, opened 2018-04-25
fix: Add name to chunk file output (#1167)

sanchitgn:
Add [name] to file name build output. This makes chunk name configurable with 
`/* webpackChunkName: "xyz" */`


LinusBorg:
Naming chunks is a plus, has my vote.

Issue #1177, opened 2018-04-26
Cannot add URLs as dependencies in custom cli plugin


joelearn:
### Version
3.0.0-beta.6

### Steps to reproduce
In a new vue-cli-plugin, add to generator.js:

```javascript
module.exports = (api, options) => {
  api.extendPackage({
    dependencies: { 'my-lib': 'https://bitbucket.org/user/my-lib.git#semver:^1.0.0' }
  })
}
```

Then invoke the plugin on a vue-cli 3 project.

### What is expected?
Dependencies should be added to package.json, whether they are defined as semver, URL, or anything valid (see: https://docs.npmjs.com/files/package.json#dependencies)

### What is actually happening?
A warning "invalid version range for dependency" is displayed in the console but the dependency merge is skipped.

dhensche:
@joelearn, based on the npm documentation it looks like the example you provided isn't actually valid? Your URL looks like a git repo (`my-lib.git#semver:^1.0.0`) but the documentation specifically lists only the `git`, `git+ssh`, `git+http`, `git+https`, or `git+file` protocols for git URLs while your example is `https`. Right now the plugin API does not support URL, Git URL, local file, or GitHub URLs because of a strict semver check for some "smart" dependency merging. I was going to look into coming up with a solution, but I wanted clarification on your example

joelearn:
@dhensche Yep, it was just an example (and perhaps a bad one at that!). I thought it might've been valid since if you ```npm install bitbucket:user/my-lib``` it saves it in that format to package.json.

My actual use-case was with the git+file protocol - that's when I first encountered the problem.

dhensche:
it may just be missing from the npm documentation. I'll take note of that. In coming up with a solution, I was assuming that the core Vue guys (I'm just doing this in some free time) would like to try and maintain the smart dependency merging that tries to use the latest version of a dependency. According to the npm documentation you can specify a semver version when using Git URLs so I was going to take that into account

PR #1184, opened 2018-04-27
feat(plugin-api): allow non-semver versioned dependencies (#1177)

dhensche:
npm supports multiple "version" strings when declaring deps in your
`package.json`. This PR is an attempt to support some of these other
formats in a plugin specification, while retaining the smart version
merging. If a semver range can be extracted, it will be used for
version conflict resolution. If not, a warning will be displayed to
the developer

### DIFF
--- a/packages/@vue/cli/lib/util/mergeDeps.js
+++ b/packages/@vue/cli/lib/util/mergeDeps.js
@@ -7,8 +7,13 @@ module.exports = function resolveDeps (generatorId, to, from, sources) {
     const r1 = to[name]
     const r2 = from[name]
     const sourceGeneratorId = sources[name]
+    const isValidURI = r2.match(/^(?:file|git|git\+ssh|git\+http|git\+https|git\+file|https?):/) != null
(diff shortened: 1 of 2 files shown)


packages/@vue/cli/lib/util/mergeDeps.js
```diff
@@ -45,6 +52,8 @@ module.exports = function resolveDeps (generatorId, to, from, sources) {
 
 const leadRE = /^(~|\^|>=?)/
 const rangeToVersion = r => r.replace(leadRE, '').replace(/x/g, '0')
+const extractSemver = r => r.replace(/^.+#semver:/, '')
```
yyx990803: shouldn't `semver:` be optional here?

dhensche: according to the npm docs (which do not seem to be 100% accurate),  if it is just `#...` and not `#semver:...` then it is referring to a commit hash, not a semver

Issue #1068, opened 2018-04-03
Allow --preset to accept a file path


JohnnyCrazy:
### What problem does this feature solve?
I want to test my local preset-repository and it would be a shame if I had to push everytime. So, for testing and general purposes, it would be nice to also load presets by specifying a path to a local file.

### What does the proposed API look like?
so we could either fallback if the preset name is not found, then we could go for:

`vue create my_app --preset preset.json`

or a more specialized way:

`vue create my_app --preset file:preset.json` or even `vue create my_app --preset-file preset.json`

The implementation it self is rather short. When there is an API decision, I could go for a PR.

yyx990803:
I think we can simply assume it's a file if the value ends in `.json`. Open to a PR.

PR #1201, opened 2018-04-30
feat(cli): allow local .json files for presets (#1068)

dhensche:
I wasn't sure how to write a test for this (didn't see any for testing the remote preset fetching either), so I didn't write one...

### DIFF
--- a/packages/@vue/cli/lib/Creator.js
+++ b/packages/@vue/cli/lib/Creator.js
@@ -1,3 +1,4 @@
+const fs = require('fs-extra')
 const chalk = require('chalk')
 const debug = require('debug')
 const execa = require('execa')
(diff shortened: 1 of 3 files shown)


packages/@vue/cli/lib/Creator.js
```diff
@@ -238,6 +239,8 @@ module.exports = class Creator {
         error(`Failed fetching remote preset ${chalk.cyan(name)}:`)
         throw e
       }
+    } else if (name.endsWith('.json')) {
+      preset = await readLocalPreset(this.context, name)
```
yyx990803: Can directly use `fs-extra` here so it's just `JSON.parse(await fs.readFile())` here, I don't think it's necessary to split into a separate file.

Akryum: It could even just be `preset = await fs.readJson(path.join(this.context, name))`

Issue #1202, opened 2018-04-30
Allow router/vuex to be late-added


jmcooper:
### Version
3.0.0-beta.6

### Reproduction link
CLI issue, no code necessary

### Steps to reproduce
1. Install vue cli 3 beta 6
2. `vue create my-project`
  2a. Do NOT install router
3. `vue add @vue/router`

### What is expected?
That it will add the router configuration/dependencies that would have been added if I selected the router from the initial install.

### What is actually happening?
No vue router plugin is found so the `vue add @vue/router` command fails.

p-adams:
I could be mistaken, but I think `vue add@vue/pluginname` command only works with those plugins prefixed with `cli-plugin-`: https://github.com/vuejs/vue-cli/tree/11956acb4d4d9dc287ce3bb7249d92734ae22a43/packages/%40vue

dhensche:
Another point about vuex/vue-router are that those are not vue-cli plugins. They are features/libraries that are just included in the generated project. I believe to add those after project creation you would just want to run `yarn install vue-router` or `npm install --save vue-router`. Because these libraries are so tightly coupled to Vue (and there are a couple files generated when added via the `vue create` CLI call) it may be nice to support adding them via the CLI as well as the standard process of using your preferred node package manager

LinusBorg:
The point is that adding these libraries with the same generated codd after the initial setup will be *very* brittle. 

Assuming that people will hNr made some changes to the initially created files, adding them later is either very complicated to get right or very likely to break tour existing code because the generated code doesn't work. 

And a plugin that *doesnt* add any files does not have any advantage over a simple `npm install`.

Akryum:
@LinusBorg As the progressive framework, we should expect users not to install vue-router or vuex right away and to want to add them later - IMHO.

> The point is that adding these libraries with the same generated code after the initial setup will be very brittle.

I think most third-party plugins are/will do this anyway. Also, if the user changes his code too much, it's OK to assume he will be on is own to integrate those libs into it. You can display a warning message, for example: `api.exitLog('"main.js" not found, follow <link to vue-router usage> instructions to add vue-router to your code.', 'warn')`

> And a plugin that doesnt add any files does not have any advantage over a simple npm install.

Yes it can, for example modifying the webpack config or adding new cli-service commands.

jim-cooper:
[...]

PR #1204, opened 2018-05-01
feat(cli): add ability to `add` vuex and vue-router internal plugins (#1202)

dhensche:
custom handling for `vuex` and `router` when invoking `vue add` to simply add the dependencies to `package.json` and point the user to documentation about integrating the libraries into their app

### DIFF
--- a/packages/@vue/cli/lib/add.js
+++ b/packages/@vue/cli/lib/add.js
@@ -6,20 +6,30 @@ const { resolveModule } = require('./util/module')
 const {
   log,
   error,
+  warn,
   hasYarn,
   stopSpinner,
   resolvePluginId
 } = require('@vue/cli-shared-utils')
 
+// vuex and vue-router are internally added via cli-service so they need to be handled differently
+const internalPluginRE = /^(vuex|router)$/
+
 async function add (pluginName, options = {}, context = process.cwd()) {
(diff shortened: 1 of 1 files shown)


Akryum:
I think it would be better to extract the vue-router and vuex generator logics from `cli-service` into their own builtin plugins.

dhensche:
@Akryum are you thinking they should be separate packages or some sort of new internal plugin structure? It sounded like the desire was to not create new packages for vuex/vue-router https://github.com/vuejs/vue-cli/issues/1202#issuecomment-385740574 and they should be treated differently than other plugins

yyx990803:
So the idea here is:

1. Split out the router/vuex parts inside `cli-service/generator` into separate files. Each of them can be a "sub generator" that exposes the same `(api, options) => {}` signature, and the main generator will just require and call them.

2. Extract the logic of invoking a required generator module into an exported function in `cli/lib/invoke.js`

3. In `cli/lib/add.js` when `router` or `vuex` special cases are encountered, require the sub-generators from `cli-service` and invoke them using the function extracted in (2).

Akryum:
Having builtin plugins would allow reusing the same UI for installing plugins. Or else we will need to work on specific UI for this.

PR #1213, opened 2018-05-02
feat(test): warn developers when they re-use a test project name

dhensche:
This warning would have saved me a fair bit of time during testing because test projects are not cleaned up between individual tests and I kept stepping on previous tests. If this is a dumb idea then you can just reject and ignore


yyx990803:
Hmm this shouldn't be a problem because of the `--force` flag at line 42. Also `yarn test` runs `yarn clean` before starting the tests. Not sure how you run into the overwriting problem?

yyx990803:
Ah ok, yeah it's common to forget changing the test project name after copy pasting when writing tests.

PR #1247, opened 2018-05-09
Fix doc

zhuoqi-chen:
Env Variables in Index HTML can not use <%= BASE_URL %>
should be <%= htmlWebpackPlugin.options.env.BASE_URL %>


yyx990803:
The docs are updated for the upcoming release.

PR #1259, opened 2018-05-11
feat: support style lang postcss

zyy7259:


Akryum:
Hum postcss is already applied to any style block. Not sure this is needed at all.

zyy7259:
I need to specify the lang attribute of the style block to enable syntax highlight.

Without lang being specified, there will be syntax problem.

image

When I specify the lang with 'postcss', the syntax highlight works fine. But webpack won't able to recognize it.

image

image

here is [a test project](https://github.com/zyy7259/test-vue-cli-style-postcss)

```bash
git clone git@github.com:zyy7259/test-vue-cli-style-postcss.git
cd test-vue-cli-style-postcss
yarn
yarn build
```

zyy7259:
or maybe change `createCSSRule('css', /\.css$/)` to `createCSSRule('css', /\.(p(ost)?)?css$/)`?

hsxfjames:
A similar discussion is already in vuejs/vetur#506.
At last @blake-newman says:
> The issue is solved with explicitness with vue-loader. <style lang="postcss/pcss/sugarcss/sss">

I think it is sensible to keep this feature. In the past, all we need is to install `postcss-cssnext` and add the custom config in `.postcssrc.js`.

zyy7259:
Why did the CI fail :(

octref:
Thanks again for supporting `<style lang="postcss">` in vue-loader. As PostCSS allows plugins to extend the syntax, it's rather hard to support its syntax highlighting and give correct language features. Having at least a `lang` attribute allows Vetur to not apply CSS grammar / language features to it and cause frustration. In the future when we have better PostCSS support in VS Code or Vetur, I will gradually turn on more supports for PostCSS in Vetur.

PR #1282, opened 2018-05-13
feat(babel-plugin): use project's babel config to transpile dependencies

Hokid:
If we add a dependency to `transpileDependencies` list then it will not
been transpiled by babel. This PR make it possible.

### DIFF
--- a/packages/@vue/cli-plugin-babel/index.js
+++ b/packages/@vue/cli-plugin-babel/index.js
@@ -3,6 +3,8 @@
 // TODO remove after upgrading Babel to 7.0.0-beta.47
 require('./patchBabel')
 
+const { babelHelpers } = require('@vue/cli-shared-utils')
+
 module.exports = (api, {
   parallel,
   transpileDependencies
@@ -44,8 +46,20 @@ module.exports = (api, {
           .loader('thread-loader')
     }
 
-    jsRule
+    const babelLoader = jsRule
       .use('babel-loader')
         .loader('babel-loader')
+
+    // use project's babel config for dependencies
+    if (transpileDependencies.length !== 0) {
+      // `babelrc` = false
+      const config = babelHelpers.loadUsersConfig(api.resolve('.'))
+
+      // prevent from ignore
+      config.ignore = [babelHelpers.resolveNodeModulesIgnorePattern(transpileDependencies)]
+
+      babelLoader
+        .options(config)
+    }
   })
 }
--- a/packages/@vue/cli-shared-utils/index.js
+++ b/packages/@vue/cli-shared-utils/index.js
@@ -4,7 +4,8 @@
   'spinner',
   'validate',
   'openBrowser',
-  'pluginResolution'
+  'pluginResolution',
+  'babelHelpers'
 ].forEach(m => {
   Object.assign(exports, require(`./lib/${m}`))
 })
--- a/packages/@vue/cli-shared-utils/lib/babelHelpers.js
+++ b/packages/@vue/cli-shared-utils/lib/babelHelpers.js
@@ -1,0 +1,39 @@
+const babel = require('@babel/core')
+
+const lib = exports.babelHelpers = {}
+
+lib.loadUsersConfig = (rootDir) => {
+  if (typeof rootDir !== 'string') {
+    throw new TypeError('`rootDir` must be a string')
+  }
+
+  const options = {
+    cwd: rootDir,
+    filename: 'package.json'
+  }
+
+  const partial = babel.loadPartialConfig(options)
+
+  delete partial.options.filename
+
+  const normalize = item => [item.file.resolved, item.options]
+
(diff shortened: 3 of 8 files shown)


packages/@vue/cli-shared-utils/index.js
```diff
@@ -4,7 +4,8 @@
   'spinner',
   'validate',
   'openBrowser',
-  'pluginResolution'
+  'pluginResolution',
+  'babelHelpers'
```
yyx990803: I don't see why this needs to be placed in `cli-shared-utils`. It's used only by `cli-plugin-babel` and should be placed in that package instead.

packages/@vue/cli-plugin-babel/index.js
```diff
@@ -44,8 +46,20 @@ module.exports = (api, {
...
       .use('babel-loader')
         .loader('babel-loader')
+
+    // use project's babel config for dependencies
+    if (transpileDependencies.length !== 0) {
+      // `babelrc` = false
+      const config = babelHelpers.loadUsersConfig(api.resolve('.'))
+
+      // prevent from ignore
+      config.ignore = [babelHelpers.resolveNodeModulesIgnorePattern(transpileDependencies)]
```
yyx990803: This is not needed because we already have the `exclude` rule above which applies `babel-loader` only to the correct files.

packages/@vue/cli-shared-utils/lib/babelHelpers.js
```diff
@@ -0,0 +1,39 @@
...
+  if (typeof rootDir !== 'string') {
+    throw new TypeError('`rootDir` must be a string')
+  }
+
+  const options = {
+    cwd: rootDir,
+    filename: 'package.json'
+  }
+
+  const partial = babel.loadPartialConfig(options)
```
yyx990803: why use `loadPartialConfig` then manually normalize instead of just using [`loadOptions`](https://github.com/babel/babel/tree/master/packages/babel-core#babelloadoptionsoptions-object)?

packages/@vue/cli-plugin-babel/index.js
Hokid: yes, you are right

packages/@vue/cli-shared-utils/index.js
Hokid: because `cli-plugin-typescript` also use babel-loader, may be in the future this will need there or some where else

packages/@vue/cli-shared-utils/lib/babelHelpers.js
Hokid: `loadPartialConfig` loads only users config. both `loadPartialConfig` and `loadOptions` return not valid config for passing it to the babel-loader(babel.transform), this why there is need manually normilize config.

packages/@vue/cli-shared-utils/lib/babelHelpers.js
```diff
@@ -0,0 +1,29 @@
...
+const lib = exports.babelHelpers = {}
+
+lib.loadUsersConfig = (rootDir) => {
+  if (typeof rootDir !== 'string') {
+    throw new TypeError('`rootDir` must be a string')
+  }
+
+  const options = {
+    cwd: rootDir,
+    filename: 'package.json'
```
loganfsmyth: Could someone clarify what specifically this is meant to do? Babel's configuration can vary per-file, so by doing this you're essentially loading the subset of the configuration that happens to apply to this specific filename.

I'd be happy to help make things work more easily with Babel, but at the moment this all seems to be going against a bunch of Babel's expectations.

packages/@vue/cli-plugin-babel/index.js
```diff
@@ -44,8 +46,17 @@ module.exports = (api, {
...
       .use('babel-loader')
         .loader('babel-loader')
+
+    // use project's babel config for dependencies
+    if (transpileDependencies.length !== 0) {
+      // `babelrc` = false
+      const config = babelHelpers.loadUsersConfig(api.resolve('.'))
+
+      babelLoader
+        .options(config)
```
loganfsmyth: How does `babelLoader` here end up calling Babel exactly? I'd normally expect the loader to do all the work of reading the config, so I'm not totally following why this tries to read the config up front.

Hokid:
@loganfsmyth, @yyx990803 
So many questions. There are suggestions how to do it better?

MartinPeverelli:
@yyx990803 @Hokid I'm sorry but I can't seem to understand how to make this work.
I have an SFC loaded via npm from a repository. Its main entry point is a `wrapper.js` that handles the import/export.
I understand that since the package is inside node_modules, the transpile process that takes place during `npm run test:unit` is ignoring it, and when my module under test tries to import it, it fails.

How can I use this `transpileDependencies` keys in order to force the `npm run test:unit` process to transpile a given package?

Thanks!

PR #1289, opened 2018-05-14
feat(lint): add max-warnings for cli-plugin-eslint, close #1268

### LINKED ISSUE #1268: eslint plugin does not support --max-warnings option
ESLint has a `--max-warnings` command line options that triggers an return code > 0 when eslint detects more than n warnings.

This is really useful for example to allow developpers to work without worrying about eslint warnings, but make the CI build fail if there are some warnings.

Since @vue/cli-plugin-eslint uses eslint's `CLIEngine` instead of the cli object, this option is not supported (it is handled by `cli.js`, not `cli-engine.js`). It would be nice to have this feature in vue cli, either by re-implementing it, or by using the `cli.js` script of eslint.

`vue-cli-service lint --max-warnings=0`

jkzing:
add a `--max-warnings` and `--max-errors` option for cli-plugin-eslint to specify number of warnings or errors to trigger build failed. 

usage:
```bash
vue-cli-service lint --max-warnings 10 --max-errors 2
```

#1268 

https://eslint.org/docs/user-guide/command-line-interface#options

### DIFF
--- a/packages/@vue/cli-plugin-eslint/lint.js
+++ b/packages/@vue/cli-plugin-eslint/lint.js
@@ -19,11 +19,13 @@ module.exports = function lint (args = {}, api) {
   if (config.fix) {
     CLIEngine.outputFixes(report)
   }
+  
+  const maxErrors = argsConfig.maxErrors || 0
+  const maxWarnings = typeof argsConfig.maxWarnings === 'number' ? argsConfig.maxWarnings : Infinity
+  const isErrorsExceeded = report.errorCount > maxErrors
+  const isWarningsExceeded = report.warningCount > maxWarnings
 
-  const isErrorsExceed = report.errorCount > (argsConfig.maxErrors || 0)
-  const isWarningsExceed = report.warningCount > (
-    argsConfig.hasOwnProperty('maxWarnings') ? argsConfig.maxWarnings : Infinity)
-  if (!isErrorsExceed && !isWarningsExceed) {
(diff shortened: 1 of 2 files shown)


BenoitAverty:
The action on the status code is missing I think. The goal is to make the CLI exit with status >0 when the number of warings is greater than the parameter.

packages/@vue/cli-plugin-eslint/lint.js
```diff
@@ -19,7 +20,7 @@ module.exports = function lint (args = {}, api) {
     CLIEngine.outputFixes(report)
   }
 
-  if (!report.errorCount) {
+  if (report.errorCount <= (argsConfig.maxWarnings || 0)) {
```
BenoitAverty: Does report.errorCount includes warnings ?

BenoitAverty:
The expected behaviour is : 
 - When there is one or more error, exit with status 1 (this is already the case)
 - When the number of warnings is greater than the parameter, exit with status 1 (This isn't the case currently, warnings don't make the exit code change to 1)
 - Otherwise, exit with status code 0

Examples: 
```
# With 0 errors and 1 warning
vue-cli-service lint --max-warnings=1 # success.
vue-cli-service lint --max-warnings=0 # failure <-- this doesn't work currently

# With 1 error
vue-cli-service lint --max-warnings=0 # failure
vue-cli-service lint --max-warnings=1 # still failure (max warnings is only about warnings)
```

PR #1292, opened 2018-05-14
Installing lastest nodejs on system

NAAG1999:
Commands to download the latest version of nodejs on ubuntu and ubuntu based destros, using terminal.


yyx990803:
Thanks but I don't think this is Vue CLI's concern - Node.js has its own documentation on installation for various systems.

PR #1293, opened 2018-05-15
docs: remove css.module from default

marocchino:
It is invalid after 1b5bdde1


yyx990803:
Thanks, it will be back in the next release.

PR #1326, opened 2018-05-18
feat(xdg-compliance): rc file location hierarchy

### LINKED ISSUE #1325: Use XDG_CONFIG_HOME for the .vuerc file
Not cluttering my HOME directory

Changing the default directory and using `$HOME` as a backup if `XDG_CONFIG_HOME` does not exist

tristan957:
`vue-cli` now uses the following hierarchy for hosting its rc file:

* `VUE_CLI_CONFIG_PATH`
* `XDG_CONFIG_HOME/vue/.vuerc`
* `HOME/.config/vue/.vuerc`
* `HOME/.vuerc`

Let me know if you need something changed

I would like to point out that the middle 2 paths probably shouldn't be hidden files. Left that up to the devs though.

closes #1325

### DIFF
--- a/packages/@vue/cli/lib/options.js
+++ b/packages/@vue/cli/lib/options.js
@@ -6,8 +6,22 @@ const { error } = require('@vue/cli-shared-utils/lib/logger')
 const { createSchema, validate } = require('@vue/cli-shared-utils/lib/validate')
 const { exit } = require('@vue/cli-shared-utils/lib/exit')
 
+const xdgConfigPath = () => {
+  const xdgConfigHome = process.env.XDG_CONFIG_HOME
+  if (xdgConfigHome) {
+    const rcDir = path.join(xdgConfigHome, 'vue')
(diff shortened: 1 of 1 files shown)


packages/@vue/cli/lib/options.js
```diff
@@ -5,8 +5,32 @@ const cloneDeep = require('lodash.clonedeep')
...
+    return path.join(rcDir, '.vuerc')
+  }
+
+  let xdgConfigHome = process.env.XDG_CONFIG_HOME
+  if (xdgConfigHome) {
+    return createRcDir(xdgConfigHome)
+  }
+
+  // XDG defined backup folder if XDG_CONFIG_HOME is undefined
+  xdgConfigHome = path.join(os.homedir(), '.config')
```
yyx990803: This would be a breaking change. We should only change behavior when `XDG_CONFIG_HOME` is explicitly present.

tristan957:
Should be fixed now

PR #1355, opened 2018-05-23
Add a --copy option to the serve command

morrislaptop:
This feature will copy the local url to the clipboard. 

I use this feature of https://poi.js.org/ quite a lot.

It's not as aggressive as `--open` but just as convenient :)

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -162,7 +162,7 @@ module.exports = (api, options) => {
         let copied = ''
         if (isFirstCompile && args.copy) {
           require('clipboardy').write(urls.localUrlForBrowser)
-          copied = chalk.dim('(copied!)')
+          copied = chalk.dim('(copied to clipboard)')
         }
 
         console.log()
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -159,10 +159,16 @@ module.exports = (api, options) => {
           return
         }
 
+        let copied = ''
+        if (isFirstCompile && (args.copy || projectDevServerOptions.copy)) {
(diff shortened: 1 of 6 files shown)


docs/config.md
```diff
@@ -66,6 +66,7 @@ module.exports = {
   // configure webpack-dev-server behavior
   devServer: {
     open: process.platform === 'darwin',
+    copy: false,
```
jkzing: `devServer` is used to configure `webpack-dev-server`, I don't think adding a non webpack-dev-server option here is a good idea.

packages/@vue/cli-service/lib/commands/serve.js
```diff
@@ -10,12 +10,15 @@ const defaults = {
   https: false
 }
 
+const clipboardy = require('clipboardy')
```
yyx990803: This should be lazy-required only when copy is needed

packages/@vue/cli-service/lib/commands/serve.js
```diff
@@ -149,10 +152,16 @@ module.exports = (api, options) => {
           return
         }
 
+        let copied = ''
+        if (isFirstCompile && (args.copy || projectDevServerOptions.copy)) {
```
yyx990803: As @jkzing suggested I think we should keep it a command line flag only. `--open` is supported in `devServer` because `webpack-dev-server` also supports it.

packages/@vue/cli-service/lib/commands/serve.js
morrislaptop: Fixed.

packages/@vue/cli-service/lib/commands/serve.js
morrislaptop: Fixed.

morrislaptop:
@yyx990803 done :)

PR #1373, opened 2018-05-26
Adding disableHostCheck default value

centerorbit:
adding default for disableHostCheck since there isn't any documentation around this specific feature for vue-cli (I found a react-webpack config StackOverflow answer that worked)

### DIFF
--- a/docs/config.md
+++ b/docs/config.md
@@ -69,6 +69,7 @@ module.exports = {
   // configure webpack-dev-server behavior
   devServer: {
     open: process.platform === 'darwin',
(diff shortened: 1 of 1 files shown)


yyx990803:
Please turn on this option yourself as leaving it on by default is a security vulnerability.

centerorbit:
I had the option default to false, which is the correct default (the secure way). I simply added it to this documentation so that people would know it's there, and could change it if wanted. Please reconsider the PR

centerorbit:
Thank you!

PR #1386, opened 2018-05-29
UI Todos

### LINKED ISSUE #1382: UI Todos
Some notes after trying it out from an actual Yarn global install:

- [x] Missing dep `launch-editor` (fixed in 3bcc511)
- [x] `Error: Failed to resolve plugin: @vue/cli-service` when creating a new project. For some reason, `node_modules` is empty with `yarn.lock` present.
- [x] Some UI default values are not in-sync with the CLI (e.g. when choosing TypeScript, "Use class component" should default to true but it's not in UI, same for "Lint on Save")
- [x] UI should show disconnected state if user terminates the process from the command line

Akryum:
Fix #1382


packages/@vue/cli-ui/package.json
```diff
@@ -16,23 +16,31 @@
     "test": "yarn run build && cd ../cli-ui-addon-webpack && yarn run build && cd ../cli-ui && yarn run test:run"
   },
   "dependencies": {
+    "@vue/cli": "^3.0.0-beta.12",
```
yyx990803: Wouldn't this become a circular dependency?

Issue #1413, opened 2018-05-31
vue-cli 2.9.4 cannot resolve template's name (unknown command init)


VeAmas:
### Version
2.9.3

### Reproduction link
[https://codesandbox.io/s/vue](https://codesandbox.io/s/vue)

### Steps to reproduce
Just init a Vue project.

### What is expected?
success

### What is actually happening?
PS D:\Git> vue init webpack certain-name

  Unknown command init.

   vue-cli · Failed to download repo vuejs-templates/[object Object]: Response code 404 (Not Found)

=> seems to resolve the word "webpack" to [object Object]

actully the version of vue-cli is 2.9.4

VeAmas:
in 2.9.3 it works fine.

VeAmas:
But you don't have vesion 2.9.4 right?
https://www.npmjs.com/package/vue-cli

fenghengzhi:
I have the same problem. 
when i run command available in vue-cli like init/list
it will show Unknow command init/list(vue init will fail,while vue list sucess)

wch0508:
$ npm install vue-cli -g
C:\Users\root\AppData\Roaming\npm\vue-list -> C:\Users\root\AppData\Roaming\npm\node_modules\vue-cli\bin\vue-list
C:\Users\root\AppData\Roaming\npm\vue -> C:\Users\root\AppData\Roaming\npm\node_modules\vue-cli\bin\vue
C:\Users\root\AppData\Roaming\npm\vue-init -> C:\Users\root\AppData\Roaming\npm\node_modules\vue-cli\bin\vue-init
+ vue-cli@2.9.4
added 252 packages from 208 contributors in 12.444s


$ vue init webpack my-project

  Unknown command init.

   vue-cli · Failed to download repo vuejs-templates/[object Object]: Response code 404 (Not Found)

wch0508:
$ npm install vue-cli@2.9.3 -g
npm WARN deprecated coffee-script@1.12.7: CoffeeScript on NPM has moved to "coff                      eescript" (no hyphen)
C:\Users\root\AppData\Roaming\npm\vue -> C:\Users\root\AppData\Roaming\npm\node_modules\vue-cli\bin\vue
C:\Users\root\AppData\Roaming\npm\vue-init -> C:\Users\root\AppData\Roaming\npm\node_modules\vue-cli\bin\vue-init
[...]

PR #1418, opened 2018-05-31
fix: #1413 check command before output help info

F-loat:


bin/vue
```diff
@@ -14,6 +14,7 @@ program
 program
   .arguments('<command>')
   .action((cmd) => {
+    if (['init', 'list', 'build'].includes(cmd)) return
```
jkzing: This just removes the redundant wanring message, and is not enough to solve `Failed to download repo vuejs-templates/[object Object]: Response code 404 (Not Found)`.

`.arguments('<command')` changes parse result of commander option, remove it would work I think.

F-loat: well, but it does work 🤔

F-loat: You are right, I tested with this and a little wrong still exist.

``` js
program
  .arguments('<command> [options]')
  .action((cmd) => {
```

yyx990803:
Thanks for the PR, actually, commander is buggy when we use the catch-all arguments trick here together with its sub-command functionality. Simply having this handler there breaks all the other commands. I've published a more conservative fix in 2.9.5.

Issue #1390, opened 2018-05-29
Laravel Valet like service for Vue CLI projects


morrislaptop:
### What problem does this feature solve?
- Avoid having to manually start `serve` for each project
- Add the ability to share sites using local tunnels
- Having developer friendly test sites using like `my-project.test` instead of `localhost:8080`

### What does the proposed API look like?
A single node service which would handle all requests. This service would spin up the dev servers on demand and then proxy requests to each dev server based on the domain name. 

If the feature is deemed worthwhile by the community, I'm happy to help start putting this together :)

yyx990803:
This has to be a separate project rather than something provided by vue-cli. Maybe @adamwathan will be interested? ;)

morrislaptop:
@yyx990803 what about as part of ui?
@adamwathan what are your thoughts?

morrislaptop:
@yyx990803 @adamwathan 

I've created a Laravel Valet custom driver which does this - https://github.com/morrislaptop/VueServeValetDriver

For each domain name, it spins up a webpack dev server on a unique port according to the domain name provided. 

For example a request to `my-project.test` would spin up a server at `http://localhost:1032` and then proxy other requests to this server. 

Should be good to go once #1421 is released.

yyx990803:
That sounds pretty neat for Laravel users!

morrislaptop:
Not just Laravel users! Valet is pretty general purpose, but it suffers the same fate as Laravel Mix, which is also a very good general purpose tool...

Issue #869, opened 2018-02-22
Allow user to configure the htmlPath (currently public/index.html)


lbjones:
### What problem does this feature solve?
I'd like to combine my vue frontend in the same project as my backend which is already making use of the /public folder.  It would be nice if I could configure the path to serve out of a folder of my chosing, like /hot or something.

### What does the proposed API look like?
I believe it could be added to vue.config.js like this:

```
  // where to serve dev files
  devDir: 'public',
```

lbjones:
Nevermind! 

I just figured out how to do it configuring webpack directly.  In vue.config.js:

```
const path = require('path');

module.exports = {
  lintOnSave: true,
  chainWebpack: (config) => {
    config
      .plugin('html')
      .tap(() => [{
        template: path.resolve('html/index.html'),
      }]);
  },
};
```

andy265:
I think that more prefered way is merge your options with exist options.

    .tap(([options]) => [Object.assign(options, {
        template: path.resolve('html/index.html'),
    })]);

lbjones:
I am also using laravel on the backend.  I changed what I originally had done as seen earlier on this thread.  What I ended up doing is putting laravel in the root folder and the whole frontend in a folder called vue.  Then when it builds, it dumps everything in public/assets.  Then for hot code on dev, you need to use a proxy URL.  Not sure if it's the easiest way to do it, but it works. 

Here is my vue/vue.config.js file if it helps:

```
module.exports = {
  // Project deployment base
  // By default we assume your app will be deployed at the root of a domain,
  // e.g. https://www.my-app.com/
  // If your app is deployed at a sub-path, you will need to specify that
  // sub-path here. For example, if your app is deployed at
  // https://www.foobar.com/my-app/
  // then change this to '/my-app/'
  baseUrl: '/assets/',

  // where to output built files
  outputDir: '../public/assets',

  // use the full build with in-browser compiler?
  // https://vuejs.org/v2/guide/installation.html#Runtime-Compiler-vs-Runtime-only
  compiler: true,

  // whether to use eslint-loader for lint on save.
  // valid values: true | false | 'error'
  // when set to 'error', lint errors will cause compilation to fail.
  lintOnSave: true,

  // tweak internal webpack configuration.
  // see https://github.com/vuejs/vue-cli/blob/dev/docs/webpack.md
  configureWebpack: {},

  chainWebpack: (config) => {
    config
      .entry('app')
      .clear()
      .add('./src/index.js')
[...]

PR #1421, opened 2018-05-31
Only provide baseUrl fix if baseUrl provided

morrislaptop:
In my scenario, the page is being served by PHP and the URL to the bundle is pointed directly to the webpack dev server. 

So, the website is being served at `http://my-project.test` and the script is pointing to `http://localhost:8080/app.js`. 

By specifying `?/sockjs-node` when including the webpack dev server client, it is forcing the sockjs server to be on the same domain as the page. Because of https://github.com/webpack/webpack-dev-server/blob/ab4eeb007283fdf3e8950ff1f3ff8150a4812061/client-src/default/index.js#L30

I need the __resourceQuery to be empty so that it gets the URL for the sockjs server from the `getCurrentScriptSource`

This PR shouldn't break for anyone already using `baseUrl`. 

Or if there is another way to achieve what I need, please let me know!

Also required by #1390 which I am working on!

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -64,7 +64,7 @@ module.exports = (api, options) => {
     if (!isProduction) {
       const devClients = [
         // dev server client
-        require.resolve(`webpack-dev-server/client`) + '?/sockjs-node',
+        require.resolve(`webpack-dev-server/client`) + (options.baseUrl !== '/' ? '?/sockjs-node' : ''),
         // hmr client
         require.resolve(projectDevServerOptions.hotOnly
           ? 'webpack/hot/only-dev-server'


lbjones:
Dang it, this broke my solution over here https://github.com/vuejs/vue-cli/issues/869#issuecomment-378758824

lbjones:
Can you explain how to use the valet driver?  I'm getting errors when I try: 

```
08:29:41 ltrain at in ~/www/vue
$ valet which
PHP Warning:  preg_match(): Compilation failed: nothing to repeat at offset 0 in ~/.valet/Drivers/WebpackDevServerBaseDriver.php on line 107

Warning: preg_match(): Compilation failed: nothing to repeat at offset 0 in ~/.valet/Drivers/WebpackDevServerBaseDriver.php on line 107
This site is served by [BasicValetDriver].
```

Issue #1440, opened 2018-06-03
vue ui - vue/html-indent incorrectly handling tabs


CalebKester:
### Version
3.0.0-beta.15

### Reproduction link
[https://github.com/vuejs/vue-cli/blob/60c3d300285c85acc0046ffcae34176e5848c9d9/packages/%40vue/cli-plugin-eslint/ui.js#L67](https://github.com/vuejs/vue-cli/blob/60c3d300285c85acc0046ffcae34176e5848c9d9/packages/%40vue/cli-plugin-eslint/ui.js#L67)

### Steps to reproduce
1. vue ui
2. ESlint configuration
3. Change Template Indentation to "tabs"
4. Compare eslint config of `'vue/html-indent': ['error', 'tabs'],` with the actual config `https://github.com/vuejs/eslint-plugin-vue/blob/master/docs/rules/html-indent.md`

### What is expected?
The vale of html-indent should be `tab`

### What is actually happening?
The value of html indent is `tabs`

PR #1441, opened 2018-06-03
Changing tabs to tab for vue/html-indent

CalebKester:
See #1440 changing 'vue/html-indent': ['error', 'tab']` to match the vue/html-indent specs

I hope this is where the fix is. If not no harm in closing this.

Thanks for the great work!


Akryum:
Thanks! :+1:

PR #1445, opened 2018-06-03
fix(ui): correct i18n mistake

SevenOutman:
I was writing locale for vue-cli ui and found this mistake in the parameter-prompt in project's `build` task page.

### Addition
Well, then another one.
image


Akryum:
Thanks! :+1:

PR #1448, opened 2018-06-03
make it working on windows again

fprijate:
The last working version on windows was 2.8.3 (before promisses).
2.9.* versions don't work on win , but work in linux.
The problem is inquirer module.


yyx990803:
Please do not increment the package's own version in pull requests. Usually maintainers will decide when to bump before publishing new releases.

PR #1450, opened 2018-06-03
Refactor config & task icons

Akryum:
# Breaking changes

- `file-icon` for the configurations is removed
- Configuration objects `icon` option changed and is now working differently: you can either use a material icon code or a custom image (see Public static files in the UI Plugin docs).
- Task objects have a new `icon` option wich works exactly the same
- By default, if no icon is provided for either the config or the task, the corresponding vue-cli plugin logo will be used instead (if any).


yyx990803:
Make sure to record the breaking changes in the commit message per [commit conventions](https://github.com/vuejs/vue-cli/blob/dev/.github/COMMIT_CONVENTION.md) when merging.

PR #1460, opened 2018-06-04
feat(lint): default to lint tsx files 

jkzing:
LInt tsx files on default as `cli-plugin-typescript` supports `.tsx`.

PS: Do we have a chance to support tslint `--project` option here?

### DIFF
--- a/packages/@vue/cli-plugin-typescript/lib/tslint.js
+++ b/packages/@vue/cli-plugin-typescript/lib/tslint.js
@@ -75,7 +75,7 @@ module.exports = function lint (args = {}, api, silent) {
 
   const files = args._ && args._.length
     ? args._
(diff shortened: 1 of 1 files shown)


packages/@vue/cli-plugin-typescript/lib/tslint.js
```diff
@@ -75,7 +75,7 @@ module.exports = function lint (args = {}, api, silent) {
 
   const files = args._ && args._.length
     ? args._
-    : ['src/**/*.ts', 'src/**/*.vue', 'tests/**/*.ts']
+    : ['src/**/*.ts', 'src/**/*.vue', 'src/**/*.tsx', 'tests/**/*.ts']
```
yyx990803: in this case we need it for `tests` as well

PR #1502, opened 2018-06-07
Added GraphQL Loader

BlackMix:
:)


egoist:
hmm well there's even `.pug` support in the default config, I think it's ok to add `.gql` support since it's used a lot and doesn't introduce any dependencies.

yyx990803:
Hmm yeah I don't think it should be in core.

Akryum:
@egoist pug support is included for the `.vue` files.

egoist:
@Akryum I mean the dependency is not included in `vue-cli` 😄

Akryum:
But not in cli-service either so it won't work without another step from the user.

PR #1508, opened 2018-06-07
More concise example with hot reloading and chaining

Achronius:
For _Passing Options to Pre-Processor Loaders_

Taken from the comment by [hurradieweltgehtunter](https://github.com/hurradieweltgehtunter) from  [here](https://github.com/vuejs/vue-loader/issues/328#issuecomment-395086386)

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -319,13 +319,12 @@ Sometimes you may want to pass options to the pre-processor's webpack loader. Yo
 
 ``` js
 // vue.config.js
-const fs = require('fs')
 
 module.exports = {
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks for the PR! I happen to have included the same during the docs refactor, but thanks still.

Achronius:
@yyx990803 Line 2 `const fs = require('fs')` is not needed anymore, also `@/variables.scss` does not have the same hot-reloading effect, only `src/variables.scss` (after serving again, tested with v3.0.0-rc.1)

Issue #1525, opened 2018-06-08
Allow relative sockjs URL


sickp:
### What problem does this feature solve?
I am using Nginx in development to terminate SSL and proxy to the dev-server (https://nginx/ => http://devserver:8080/) with `devServer.public` set appropriately, so I can use HTTPS/HTTP2 during development.

This had been working fine until beta16 commit:

https://github.com/vuejs/vue-cli/commit/2cbe373

which makes sockjs URL absolute with the wrong host/port, now in this case: `https://10.0.1.6:8080/sockjs-node`. Previously it was simply `/sockjs-node`.

I would like an option to either specify `sockjsUrl` explicitly OR an option to omit the protocol/host/port and just use the pathname `/sockjs-node`.
 

### What does the proposed API look like?
```
module.exports = {
   devServer: {
      disableHostCheck: true,
      public: 'local.mox.cc',
      sockjsRelative: true, // OR
      sockjsUrl: '/sockjs-node'
   }
}
```

JohnCampionJr:
I think this would be useful, not sure why we wouldn't want this to be able to be relative

sokolovstas:
@sickp How to use devServer.public to build relative url?

PR #1526, opened 2018-06-08
Use devServer.public to build sockjsUrl, if defined.

sickp:
See: https://github.com/vuejs/vue-cli/issues/1525


yuanzhi0219:
pretty nice

yuanzhi0219:
I think this needs to be written in the document.

Issue #1482, opened 2018-06-06
Ask for the first commit message or skip git initialization


TotomInc:
### What problem does this feature solve?
According to [this issue](https://github.com/vuejs/vue-cli/issues/1086), sometimes the user wants to change the initial commit message which is "init" once a project is initialized.

### What does the proposed API look like?
It is already implemented in the CLI using `-g --git` argument so it should not be too much work to implement it also in the UI.

PR #1541, opened 2018-06-10
feat: Allow git commit message through UI

voluntadpear:
Implements #1482.

Hi! I went ahead and implemented support on UI to specify a custom initial git commit message and skip git initialization.

I'm not sure if using a single switch for both actions and skip git initialization if the switch is on but no message is specified is the best approach, I can switch to something else like having different switches if it seems a better option.

### DIFF
--- a/packages/@vue/cli-ui/locales/en.json
+++ b/packages/@vue/cli-ui/locales/en.json
@@ -199,7 +199,7 @@
             "options": {
               "label": "Additional options",
               "description": "Overwrite target folder if it exists",
-              "git": "Use custom first initial commit message or skip git initialization",
+              "git": "Initialize git repository (recommended)",
               "gitPlaceholder": "Initial commit message (optional)"
             }
           },
--- a/packages/@vue/cli-ui/locales/en.json
+++ b/packages/@vue/cli-ui/locales/en.json
@@ -198,9 +198,9 @@
             },
             "options": {
               "label": "Additional options",
-              "description": "Overwrite target folder if it exists",
+              "force": "Overwrite target folder if it exists",
               "git": "Initialize git repository (recommended)",
-              "gitPlaceholder": "Initial commit message (optional)"
+              "git-commit-message": "Initial commit message (optional)"
             }
           },
           "buttons": {
--- a/packages/@vue/cli-ui/locales/en.json
+++ b/packages/@vue/cli-ui/locales/en.json
@@ -198,7 +198,9 @@
             },
             "options": {
               "label": "Additional options",
-              "description": "Overwrite target folder if it exists"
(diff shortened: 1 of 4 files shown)


packages/@vue/cli-ui/src/views/ProjectCreate.vue
```diff
@@ -388,6 +399,8 @@ function formDataFactory () {
   return {
     folder: '',
     force: false,
+    git: false,
```
Akryum: This should be `true` by default to be consistent with the cli behavior.

packages/@vue/cli-ui/locales/en.json
```diff
@@ -191,7 +191,9 @@
             },
             "options": {
               "label": "Additional options",
-              "description": "Overwrite target folder if it exists"
+              "description": "Overwrite target folder if it exists",
+              "git": "Use custom first initial commit message or skip git initialization",
```
Akryum: Let's put `"Initialize git repository (recommended)"`.

packages/@vue/cli-ui/src/graphql-api/connectors/projects.js
```diff
@@ -281,6 +281,16 @@ async function create (input, context) {
       answers.useConfigFiles = 'files'
     }
 
+    const cliOptions = {git: true}
+    // Git
+    if (input.skipGit) {
```
Akryum: The name of the variable is wrong, it should be something like `enableGit`.

packages/@vue/cli-ui/src/graphql-api/connectors/projects.js
```diff
@@ -281,6 +281,16 @@ async function create (input, context) {
...
+      if (!input.gitCommit) {
+        cliOptions.git = 'false'
```
Akryum: Should be `false`.

packages/@vue/cli-ui/src/graphql-api/schema/project.js
```diff
@@ -37,6 +37,8 @@ input ProjectCreateInput {
   remote: Boolean
   clone: Boolean
   save: String
+  skipGit: Boolean!
```
Akryum: `enableGit`

packages/@vue/cli-ui/src/views/ProjectCreate.vue
```diff
@@ -512,6 +525,8 @@ export default {
             input: {
               folder: this.formData.folder,
               force: this.formData.force,
+              skipGit: this.formData.git,
```
Akryum: `enableGit`

voluntadpear:
Thanks for the review! I added the requested changes.

PR #1554, opened 2018-06-12
Command: vue config implemented 

beardedpayton:
I've wrapped up the new ```vue config``` command.

**Takeway:**

If there is a better way to do a conditional check to only print the config and resolved path when ```vue config``` is invoked let me know.

Out of the box Commander doesn't support options with multiple values. My work around for the ```vue config -s <path> <value>``` command was to pass an optional argument to the command and do a conditional check if present.

I added the dependency [opn ](https://www.npmjs.com/package/opn) for opening the config file with ```vue config -e``` if there is a shared utility for this I couldn't track it down.

Hope everything looks good!


beardedpayton:
@Akryum how's this look ?

beardedpayton:
@yyx990803 thanks for the feedback. Can you provide me an example of proposed wording for one of the commands?

yyx990803:
Just replacing `preset` with `config` should be good for now.

beardedpayton:
@yyx990803 Changes made and pushed.

Nested path support implemented with dependencies
[get-value](https://www.npmjs.com/package/get-value)
[unset-value](https://www.npmjs.com/package/unset-value)
[set-value](https://www.npmjs.com/package/set-value)

commands/logs updated

sqal:
@beardedpayton 

> Nested path support implemented with dependencies
get-value
unset-value
set-value

Wouldn't be better to just one dependency for this case? I am thinking https://github.com/sindresorhus/dot-prop

beardedpayton:
@yyx990803 Do the changes look good to be merged? Or does this PR need more work?

packages/@vue/cli/bin/vue.js
```diff
@@ -116,6 +116,17 @@ program
     loadCommand('init', '@vue/cli-init')
   })
 
+program
+  .command('config [value]')
+  .description('inspect and modify the config')
+  .option('-g, --get <path>', 'get value from preset')
+  .option('-s, --set <path> <value>', 'set preset value')
+  .option('-d, --delete <path>', 'delete preset from config')
```
pksunkara: You are still using the word `preset` in here and above.

PR #1558, opened 2018-06-13
Update deployment.md

Utwo:


Utwo:
In any CI it is recommanded to use `npm ci` instead of `npm install`. More info can be found here: [https://docs.npmjs.com/cli/ci](https://docs.npmjs.com/cli/ci)

Issue #1551, opened 2018-06-12
Programmatic api for vue-cli-service build/serve that accepts custom webpack config


nklayman:
### What problem does this feature solve?
For a new version of [vue-cli-plugin-electron-builder](github.com/nklayman/vue-cli-plugin-electron-builder) I need to be able to modify the webpack config only if my build command is called, not for the regular build. I need to customize the config to work with electron, but then it won't work in browser. Being able to `api.build(config)` is necessary for my plugin. `api.serve(config)` is not necessary, but would be quite nice. I could possibly add this myself and submit a PR, but I want to make sure this is wanted as a feature.

This kinda relates to [this comment](https://github.com/vuejs/vue-cli/issues/825#issuecomment-380622462).
### What does the proposed API look like?
Some other features I would like:

- events for recompiling on serve command
- being able to get info such as hosted port, compile errors, and other info like that provided by `webpack()`
- bundle data is still shown in UI page

essentially  a api similar to that of `webpack()`, but that includes the extra steps of `vue-cli-service build/serve` (because `webpack(api.resolveWebpackConfig())` throws errors).

nklayman:
If I do that I get a bunch of errors. The [build by vue-cli](vue-cli/packages/@vue/cli-service/lib/commands/build/index.js) includes some extra steps that prevent theses errors from happening.

yyx990803:
I think you will have to better explain the exact API you need. It's pretty vague right now.

nklayman:
I created a pr for the build command. The api.service.run('serve') works just fine for this plugin.

LinusBorg:
Closing as Author closed his own PR.

m4heshd:
@nklayman is there any way I can invoke `serve` and launch an Electron instance after dev server starts without writing any complicated Vue plugin level stuff? I've read through your source but still can't figure out. Even posted a question [here](https://stackoverflow.com/questions/65745878/how-do-i-automatically-run-an-npm-script-after-webpack-dev-server-has-been-start).

nklayman:
@m4heshd See [my comment on StackOverflow](https://stackoverflow.com/questions/65745878/how-do-i-automatically-run-an-npm-script-after-webpack-dev-server-has-been-start#comment116262611_65745878)

m4heshd:
@nklayman thank you for taking time to look through my issue. I actually went through the vue-cli source and ended up doing this which exactly does what I need to.

```javascript
const vueService = require('@vue/cli-service');
[...]

PR #1585, opened 2018-06-15
feat(build/serve): export build/serve command and allow them to take config argument (chainable webpack config)

nklayman:
Solves #1551 by changing the build command to export itself. It can then be imported and ran with a custom config that will be used instead of `api.resolveChainableWebpackConfig()`. This allows plugins to modify the webpack config only if their build command is run, not for normal builds. 

Edit: Now does same process for the serve command.

### DIFF
--- a/packages/@vue/cli-service/lib/commands/build/index.js
+++ b/packages/@vue/cli-service/lib/commands/build/index.js
@@ -69,7 +69,7 @@ const initiateBuild = async (args, api, options, config) => {
     delete process.env.VUE_CLI_MODERN_MODE
     delete process.env.VUE_CLI_MODERN_BUILD
   } else {
-    await build(args, api, options)
+    await build(args, api, options, config)
   }
   delete process.env.VUE_CLI_BUILD_TARGET
 }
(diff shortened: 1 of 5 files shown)


nklayman:
Make sure to hide whitespace changes in the diff viewer. Indentation is updated, so it looks like there are more changes.

nklayman:
For some reason that I can not find, HMR is broken in these cases:

 - serve function is imported and called from a plugin
 - serve function is called using `service.run('serve')` (assuming service = `new Service(api.resolve('.'))`)

Yet it still works when:

 - serve function is imported and called from a plugin, but only if calling package is linked with `yarn link`
 - called with `vue-cli-service serve`

nklayman:
Another method of implementing this feature would be to allow `service.run()` to take a 4th argument. That argument would be an array of arguments that will be passed to the command the service calls. This would likely fix the HMR issue (and other related ones) and make it easier to add this feature to other plugins. I can create a PR for this method if it is preferred.

Akryum:
> Another method of implementing this feature would be to allow service.run() to take a 4th argument. 

Worth a try.

nklayman:
@akryum I created a branch with service.run taking a 4th argument: https://github.com/nklayman/vue-cli/tree/service-run-4th-arg

nklayman:
If that works I can create a PR for that instead.

nklayman:
Closing as I found an alternative and lack of interest.

PR #1619, opened 2018-06-19
feat: pages support title option

posrix:
This option is necessary while developing multi-page website


LinusBorg:
This makes sense to me. 

WDYT @Akryum ?

LinusBorg:
@posrix would you also add the title atttribute to the code sample in the docs about pages [here](https://github.com/vuejs/vue-cli/tree/dev/docs/config#pages)?

PR #1628, opened 2018-06-20
fix: when script has no attributes

snadn:
when use webpack `htmlWebpackIncludeAssetsPlugin` include script. the script has no attributes, and build throw error.

image


snadn:
what about this pr? @Akryum @yyx990803

Issue #1629, opened 2018-06-20
Add end of string modifier to build/serve ui regex


nklayman:
### What problem does this feature solve?
The match regex in `@vue/cli-ui/ui-defaults/tasks.js` doesn't have an end of string modifier at the end of it ($). This means it matches `vue-cli-service build:electron` or something else similar. Since these defaults overwrite plugin ui config, this prevents developers like me for making their plugin work with the ui.

### What does the proposed API look like?
Add `$` to the end of the match regex for the build/serve commands. Created pr #1630 which does exactly this.

PR #1630, opened 2018-06-20
fix(ui): add end of string to build regex to prevent unwanted matches

nklayman:
Fixes issue #1629, pasted below:

The match regex in `@vue/cli-ui/ui-defaults/tasks.js` doesn't have an end of string modifier at the end of it ($). This means it matches `vue-cli-service build:electron` or something else similar. Since these defaults overwrite plugin ui config, this prevents developers like me for making their plugin work with the ui.

Add `$` to the end of the match regex for the build/serve commands. Created pr #1630 which does exactly this.


Akryum:
Thanks! Closing since I fixed the issue differently.

Issue #1656, opened 2018-06-22
Modern mode failing to load module when under HTTP Basic Auth [use credentials]


AlexandreBonaventure:
### Version
3.0.0-rc.3

### Reproduction link
[https://1656-repro-jdyuryvihq.now.sh/](https://1656-repro-jdyuryvihq.now.sh/)
username: john
pwd: secret

### Steps to reproduce
(No repro)
Build an app with vue-cli, and deploy the SPA under an HTTP Basic Authentification.

### What is expected?
Assets are loaded from same-origin with credentials

### What is actually happening?
JS modules assets (with type="module") are not forwarding credentials properly and fail to load with a 401 response from the server.

---
As stated in https://jakearchibald.com/2017/es-modules-in-browsers/ (No credentials section)
```
<!-- Fetched with credentials (cookies etc) -->
<script src="1.js"></script>

<!-- Fetched without credentials -->
<script type="module" src="1.mjs"></script>

<!-- Fetched with credentials -->
<script type="module" crossorigin src="1.mjs?"></script>

<!-- Fetched without credentials -->
<script type="module" crossorigin src="https://other-origin/1.mjs"></script>

<!-- Fetched with credentials-->
<script type="module" crossorigin="use-credentials" src="https://other-origin/1.mjs?"></script>
```
We then have to modify the ModernModePlugin https://github.com/vuejs/vue-cli/blob/0b42f2f77a89602dfcfa54d3bbf57f7bdd31420a/packages/%40vue/cli-service/lib/webpack/ModernModePlugin.js
to include crossorigin attribute, or give the ability to add it via vue.config.js

zhouxinyong:
I had the same issue, and the type="module" js  in network  reponse type is text/html!

AlexandreBonaventure:
Here's a repro for you guys [https://1656-repro-jdyuryvihq.now.sh/](https://1656-repro-jdyuryvihq.now.sh/)
username: john
pwd: secret

AlexandreBonaventure:
reproduction repo: https://github.com/AlexandreBonaventure/1656-repro

yyx990803:
Note: the default has been reverted to anonymous CORS in 30215c28 and you will need to use `corsUseCredentials: true` in `vue.config.js` to enforce `crossorigin="use-credentials"` in the next release (after RC5).

PR #1695, opened 2018-06-27
fix(modern): send credentials when loading script modules.

sickp:
Otherwise, things like BasicAuth won't work.
See: https://github.com/vuejs/vue-cli/issues/1656


yyx990803:
Note: the default has been reverted to anonymous CORS in 30215c28 and you will need to use `corsUseCredentials: true` in `vue.config.js` to enforce `crossorigin="use-credentials"` in the next release (after RC5).

PR #1717, opened 2018-06-30
feature: Generating README.md

### LINKED ISSUE #1675: Generate readme.md
Currently, every new project is generated with no readme.md file. Generating a readme.md with basic deployment information would help rapid prototyping. The end user will have a starting point to expand upon. Example:

```
# my-vue-project

## Build Setup

### dependencies:
npm install # Or yarn install

### spin up server with hot reload (npm run / yarn)
npm run serve

### build for production
npm run build

### review configuration
npm run inspect

## Made with [vue-cli](https://cli.vuejs.org/).

```

```
# client would either generate a generic readme.me or enhanced with provided details (project-name, selected plugins, etc)
vue create my-vue-project
```

arthurdenner:
Closes #1675.

As suggested in the issue, this PR adds a step to the `vue create` process to create a `README.md` based on the project's `package.json`.

I got some descriptions from the `cli-ui` package and created generic ones to the others. Please let me know if I missed any script/description.

Regarding the code, I used `Array.join`s to try to keep the code as clean as possible. I'm not sure if it's the best solution, but using only template strings in this situation is really messy in my opinion.

PS: If tests are necessary to this feature, let me know as well.

### DIFF
--- a/packages/@vue/cli/lib/util/generateReadme.js
+++ b/packages/@vue/cli/lib/util/generateReadme.js
@@ -1,0 +1,35 @@
+const textsForScripts = {
+  build: {
+    command: 'run build',
+    description: 'Build for production'
+  },
+  serve: {
+    command: 'run serve',
+    description: 'Start local server with hot reload'
+  },
+  test: {
+    command: 'run test',
+    description: 'Start local server with hot reload'
+  }
+}
+
+function getScripts (pkg, packageManager) {
+  return Object.keys(pkg.scripts).map(key => {
+    const text = textsForScripts[key]
+
+    if (text) {
+      return `
+### ${text.description}
+${packageManager} ${text.command}
+`
+    }
+  }).join('')
+}
+
+module.exports = function generateReadme (pkg, packageManager) {
+  return `# ${pkg.name}
+
+## Build setup
+${packageManager} install
+${getScripts(pkg, packageManager)}`
+}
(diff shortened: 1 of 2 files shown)


packages/@vue/cli/lib/util/generateReadme.js
```diff
@@ -0,0 +1,26 @@
...
+  lint: 'Lints and fixes files',
+  test: 'Run your tests',
+  'test:e2e': 'Run your end-to-end tests',
+  'test:unit': 'Run your unit tests'
+}
+
+function printScripts (pkg, packageManager) {
+  return Object.keys(pkg.scripts).map(key => {
+    return [
+      `\n### ${descriptions[key]}`,
```
kazupon: about `descriptions`, I think it should be decided dynamically according to features selection at `vue create`.

In that implementation, the description which is not necessary is generated in the `README.md`.
for example, in the features selection, even if unit test feature is not selected, it will be output to `README.md`.

arthurdenner: But the file is generated based on the keys on the `scripts` section of `package.json`. If you don't select unit testing, it won't output the description on `README.md`.

kazupon: That's true.

In that phase, it's works, due to `package.json` contents is resolved with `Generator`.
For #1675, There seems to be no problem with implementation.  ✨ 

As the concern, If third party plugin author want to  extend `README.md` dynamically, I think that it's not that simple implementation, but it need to implement at `Generator` or `cli-service`.

pksunkara: @kazupon That would be even better

PR #1736, opened 2018-07-03
Fix #1735

### LINKED ISSUE #1735: Vue Library-Build contains ES6 language
### Version
3.0.0-rc.3

### Reproduction link
[https://github.com/michaelfeinbier/vue-cli-lib-bug/](https://github.com/michaelfeinbier/vue-cli-lib-bug/)

### Steps to reproduce
Create a new, empty vue-cli project with "default" preset and build as library with `yarn build --target lib`

### What is expected?
Based on the babel-plugin and the default browserlist I expect all ES6 language features transpiled correctly in the output. 

In fact, the file `@vue/cli-service/lib/commands/build/setPublicPath.js` is simply copied into the output but contains a `let i` which throws errors on internet explorers. You can see the generated build in my demo repository.
https://github.com/michaelfeinbier/vue-cli-lib-bug/blob/master/dist/test.umd.js#L355

LinusBorg:
Don't use ES6 in code that we inject into --lib builds
closes #1735

### DIFF
--- a/packages/@vue/cli-service/lib/commands/build/setPublicPath.js
+++ b/packages/@vue/cli-service/lib/commands/build/setPublicPath.js
@@ -1,7 +1,7 @@
 // This file is imported into lib/wc client bundles.
 
 if (typeof window !== 'undefined') {
-  let i
+  var i
   if ((i = window.document.currentScript) && (i = i.src.match(/(.+\/)[^/]+\.js$/))) {
     __webpack_public_path__ = i[1] // eslint-disable-line
   }


antopravin:
Hi, 
Due to this problem, We aren't able to roll out a release to support ios9 devices.
May I know when to expect this fix to be released?

LinusBorg:
Are you maybe on an older version?

antopravin:
hmm, may be, here are the dependencies
    "@vue/cli-plugin-babel": "^3.0.0-beta.6",
    "@vue/cli-plugin-e2e-nightwatch": "^3.0.0-beta.6",
    "@vue/cli-plugin-eslint": "^3.0.0-beta.6",
    "@vue/cli-plugin-unit-jest": "^3.0.0-beta.6",
    "@vue/cli-service": "^3.0.0-beta.6",
    "@vue/eslint-config-standard": "^3.0.0-beta.6",

LinusBorg:
After beta.6 came another 10 beta releases, and now we are on the 3rd Release Candidate.

antopravin:
oh ok. I will try updating them to 3rd release candidate

antopravin:
Thanks a lot that seemed to fix the let on the setPublicPath.js.
But I can still see a let being used on the minified versio, file dist/vendor.db846e77.js 
`let t;(t=window.document.currentScript)`

LinusBorg:
There's not `let` in that file's source code, so that would be very surprising. 

Can you provide a reproduction?

antopravin:
You are right! 
The issue is due to a transitive dependencies of old Vue-cli-service versions in our package.json.
We'll work with the teams to get them updated to RC4 version of Vue-cli-service.
Thanks for the support!

PR #1746, opened 2018-07-04
feat: Better public/index.html template

pksunkara:
Better syntax style for the `public/index.html` file.


yyx990803:
Sorry but this looks quite subjective to me and doesn't add real value IMO.

PR #1747, opened 2018-07-04
feat: Remove unnecessary .vue extensions when importing

pksunkara:
Also use the `@` alias everywhere to keep things consistent


pksunkara:
I would appreciate any help in trying to understand why the tests are failing here.

yyx990803:
I recommend leaving the extensions in there to be explicit.

pksunkara:
Okay, It's just that we removed them from the webpack template, so thought we might want to do that here too.

PR #1758, opened 2018-07-04
fix(babel): set useBuiltins to false during MODERN_BUILD

### LINKED ISSUE #1727: Modern build gets unneeded polyfills detected by Babel
### Version
3.0.0-rc.3

### Reproduction link
[https://github.com/cristianl/repro-vue-cli-usebuiltins](https://github.com/cristianl/repro-vue-cli-usebuiltins)

### Steps to reproduce
In the repro directory, run `node_modules/.bin/vue-cli-service build --modern`. 

The reproduction repo is a vue-cli generated project with the default settings, and only this function definition added to main.js, so Babel detects usage of `Promise`:

```js
function fakePromise() {
  return Promise.all()
}
```

### What is expected?
The vendors bundle generated for modern mode should not get polyfills of ES2015 features already supported by the browsers it targets.

Babel is automatically applying polyfills even for the modern build (in this case, polyfills for `Promise`, detected in *main.js*).

---
It's fixed by setting `useBuiltIns: false` when `process.env.VUE_CLI_MODERN_BUILD` is true.

It may also be a bug in babel-preset-env (when the target is `esmodules`, should it apply ES2015 polyfills?)

LinusBorg:
closes #1727 

when building in modern mode, polyfills still get included by babel-preset-env even though we set `targets: { esmodules: true }` [here](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/babel-preset-app/index.js#L71) and skip applying the default polyfills [here](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/babel-preset-app/index.js#L94)

Appereantly babel-preset-env still respects the `browserslist` field from package.json when `useBuiltIns` is `'usage'` or `'entry'`.

This PR sets `useBuiltIns: `false` when a modern build is running.

I'm think pondering wether this could have any side-effects?


LinusBorg:
So I've been thinking .... When set `useBuiltins: false`, babel-preset-env will not add any polyfills automatically anymore, even though the user might expect it since he maybe even set `useBuiltIns` explicitly.

Right now, I don't see a problem with that because preset-env will only polyfill stuff that's currently standard, and all browsers that support `type="module"` support the current standard 100%, right?

But if at some point in the future one of those browsers would lag behind on a feature that becomes standard, the "modern" bundle would not receive that polyfill even though the browser might have been configured correctly in `browserslist` to get it.

Am I thinking correctly here?

If so - is there another way to implement this? Can we somehow reset the browser targets which are defined in package.json's `browserslist` field while we are building MODERN?

LinusBorg:
Ok so I think I got it now. from preset-env docs:

> You may also target browsers supporting ES Modules (https://www.ecma-international.org/ecma-262/6.0/#sec-modules). When specifying this option, the browsers field will be ignored. 

This makes it sound as if `browserslist` will be ignored, but it seems that it only ignores browsers defined in the plugin options. This may even be a bug?

Anyway, instead of setting `useBuiltins: false` and thereby deactivating any and all polyfills, we can use [`ignoreBrowserslistConfig`](https://babeljs.io/docs/en/next/babel-preset-env.html#ignorebrowserslistconfig) to skip `package.browserslist`.

So if in the future the browsers supporting esmodules may require a polyfill to run then-current standard js, `useBuiltins` will still add that polyfill.

Seems much better to me.

Thoughts @yyx990803 ?

LinusBorg:
This is ready as far as I'm concerned. I think I weighted all the options.

HKochniss:
I think this behavior is interesting for performance-minded people, who want to be sure that modern is really a "minimal" build. Therefor: will this be menioned in the docs? :)

LinusBorg:
> will this be mentioned in the docs?

We are always happy to accept PRs :)

PR #1767, opened 2018-07-05
added correct hostname for webpack-dev-server which will be provided in vue-cli flag

shemul:
I tried to run the vue-cli bootstrapped app in my cloud9 ide. As cloud9 provides own layered hostname for each instance of their app. I could run the app using 'npm run serve' but the app was looking for hot reload socket at `localhost ` but my hostname was different than `localhost`. I put `--host 0.0.0.0` and did not work. So I checked the cli serve command and which this i found.


LinusBorg:
You could achieve a similar result by using the --public argument:  `--public 0.0.0.0:8080`

You can also set this flag in the devServer options of `vue.config.js`

pksunkara:
Can confirm what @LinusBorg said. The issue should be closed.

PR #1772, opened 2018-07-05
chore: update main deps

Akryum:


yyx990803:
Group 2 is failing, specifically TS + Jest (`ts-jest` somehow cannot be found?)

Akryum:
I think it's because I didn't update de generators yet.

Akryum:
It seems version 23.0.0 breaks things.

Akryum:
Well that's not it. :|

PR #1774, opened 2018-07-05
Don't allow duplicate injections of import statements and root options by plugins

pksunkara:
This gives more flexibility for the plugins.


pksunkara:
🎉

Issue #1753, opened 2018-07-04
Disabling hash from filename / build ignores assetsDir


CoolGoose:
### Version
3.0.0-rc.3

### Reproduction link
[https://codepen.io/anon/pen/zaXgNz](https://codepen.io/anon/pen/zaXgNz)

### Steps to reproduce
Setup vue-cli , and use the setup of vue.config.js from the link above.

### What is expected?
To have dist/assets/app.css

### What is actually happening?
Without specifying the assets folder manually in the filename it ignores the assetsDir option.

Akryum:
I'm not sure you can specify a path in the file names. Please try:

```js
module.exports = {
  css: {
    extract: {
      filename: '[name].css',
      chunkFilename: '[id].css'
    }
  }
}
```

CoolGoose:
@Akryum actually it works, the example on codepen is working. Just that it's weird to hack around that way :)

CoolGoose:
@Akryum Why is this closed ? The question remains on why a manual overwrite needs to happen since it doesn't respect the assetsDir setting

Akryum:
You issue wasn't clear to me. You mean the extracted CSS isn't under `assetsDir`?

CoolGoose:
@Akryum 
That's correct, the CSS and JS aren't under assetsDir. Hence the hack of adding assetsDir + filename in chainWebpack and configureWebpack.

It should just work with 
```
configureWebpack: {
    output: {
      filename: "[name].js",
      chunkFilename: "[name].js"
    }
  },
```

and not with 
```
configureWebpack: {
    output: {
      filename: assetsDir + "/[name].js",
      chunkFilename: assetsDir + "/[name].js"
    }
  },
```

Or is the assetsDir not supposed to apply anymore once you manually overwrite the filename / chunkFilename ?

Akryum:
> Or is the assetsDir not supposed to apply anymore once you manually overwrite the filename / chunkFilename ?

Yes, because we don't know what you really want. By default we set `filename` to `assetsDir + '/css/[name].[contenthash:8].css'`, so if you overwrite it well it doesn't have `assetsDir` in it.

What you should do in `vue.config.js`:

```js
module.exports = {
  css: {
    extract: {
      filename: 'my-assets/[name].css',
      chunkFilename: 'my-assets/[id].css'
    }
  }
}
```

[...]

Issue #1649, opened 2018-06-21
Can't disable hashing from build


ram33:
### Version
3.0.0-rc.3

### Steps to reproduce
Have tried below two options as suggested in docs... but doesn't seem to work

//vue.config.js
```
module.exports = {
  chainWebpack: config => {
    config
      .plugin('html')
      .tap(args => {
        args[0].hash = false;
        return args
      })
  }
}

module.exports = {
  chainWebpack: config => {
    config
      .plugin('html')
      .tap(args => [{
        hash: false
      }])
  }
}

```

### What is expected?
Was expecting output file name to be just app.css instead of app.########.css

### What is actually happening?
I am getting app.########.css

yyx990803:
Use

- `'[name].js'` for `output.filename` and `output.chunkFilename` in `chainWebpack`
- `{ filename: '[name].css' }` for `css.extract` option in `vue.config.js`.

Akryum:
Try:

```js
// vue.config.js
module.exports = {
  chainWebpack: config => {
    if(config.plugins.has('extract-css')) {
      const extractCSSPlugin = config.plugin('extract-css')
      extractCSSPlugin && extractCSSPlugin.tap(() => [{
        filename: '[name].css',
        chunkFilename: '[name].css'
      }])
    }
  },
  configureWebpack: {
    output: {
      filename: '[name].js',
      chunkFilename: '[name].js'
    }
  }
}
```

ram33:
@yyx990803 @Akryum That worked... Thank you.

pedroresende:
@Akryum This should really go into documentation, it would help a lot of people.

Akryum:
@yyx990803 Maybe we can add a "Common recipes" page in the docs?

DrSensor:
How to disable `filenameHashing` on a specific file/chunk?
I have a case to disable the hashing for only each of theme.scss

benlind:
Evan's solution looks like this in `vue.config.js`:

```
module.exports = {
  configureWebpack: config => {
    config.output.filename = '[name].js'
    config.output.chunkFilename = '[name].js'
  },
  css: {
    extract: {
      filename: '[name].css',
    },
  },
}
```

ffxsam:
Just adding this here, to keep people up to date. This is now an option in `vue.config.js`:

https://cli.vuejs.org/config/#filenamehashing

kambanwait:
> Try:
> 
> ```js
> // vue.config.js
[...]

Issue #1478, opened 2018-06-05
Allowing the disabling or removal of the HTMLPlugin


cecilphillip:
### What problem does this feature solve?
I'm using v3.0.0-beta.15 of the CLI.

I have an existing backend that generates it's down default document. I don't need an index.html file generated or copied to my output folder.

I'm able to run `vue create .` in a sub folder of my project. I've manually deleted the `public` folder. Then I've created a `vue.config.js` file and tried the follow 3 options. However, non of these options work in preventing the index.html being copied.
```JavaScript
    chainWebpack: config => {
        config.plugin('html').clear();
    }

....

    chainWebpack: config => {
        config.plugins.delete('html');
    }

....

    chainWebpack: config => {
        config.plugins.delete('copy');
    }
```

### What does the proposed API look like?
This configuration setting should work, unless there's an alternative API already available that would be a better fit.

```JavaScript
   chainWebpack: config => {
           config.plugins.delete('html');
   }
```

cecilphillip:
I was able to get this to work after disabling some other plugins, but I still think it would be better behind a config setting.

```JavaScript
module.exports = {
    outputDir: '../wwwroot',
    chainWebpack: config => {
        config.plugins.delete('html');
        config.plugins.delete('preload');
        config.plugins.delete('prefetch');
    }
};
```

yyx990803:
This should work:

``` js
module.exports = {
  chainWebpack: config => {
    config.plugins.delete('html')
    config.plugins.delete('preload')
    config.plugins.delete('prefetch')
  }
}
```

MrToy:
If in multi-page mode,How to disable index generation?
Here is my option, however it still generate the index.html and subpage.html
```js
module.exports = {
    filenameHashing: false,
    pages: {
        index: 'src/main.js',
        subpage: 'src/sub.js'
    },
    chainWebpack: config => {
        config.plugins.delete('html')
        config.plugins.delete('preload')
        config.plugins.delete('prefetch')
    }
}
```

mikedidthis:
```JS
module.exports = {
    filenameHashing: false,
    pages: {
        index: 'src/main.js',
        subpage: 'src/sub.js'
    },
    chainWebpack: config => {
        config.plugins.delete('html-index')
        config.plugins.delete('preload-index')
        config.plugins.delete('prefetch-index')
        config.plugins.delete('html-subpage')
        config.plugins.delete('preload-subpage')
[...]

PR #1779, opened 2018-07-06
Add configuration examples section

betweenbrain:
* Adds example to disable hashed asset filenames, as discussed in #1649 
* Adds example of disabling the generation of `index.html`, as discussed in #1478 
* Adds a tip about not needing to use `assetsDir` when overriding `filename` or `chunkFilename`, as discussed in #1753

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -341,3 +341,48 @@ See [@vue/cli-plugin-e2e-cypress](https://github.com/vuejs/vue-cli/tree/dev/pack
 ### Nightwatch
 
 See [@vue/cli-plugin-e2e-nightwatch](https://github.com/vuejs/vue-cli/tree/dev/packages/%40vue/cli-plugin-e2e-nightwatch) for more details.
+
+## Example Configurations
+### Disable Hashed Filenames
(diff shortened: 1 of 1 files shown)


LinusBorg:
I like this. 

While the Guide contains a few examples of the general approach, there sure are some adjustments that people tend to ask for quite often, so a collection of those does make sense.

docs/config/README.md
```diff
@@ -341,3 +341,48 @@ See [@vue/cli-plugin-e2e-cypress](https://github.com/vuejs/vue-cli/tree/dev/pack
...
+    },
+  },
+};
+```
+::: tip
+When manually overwriting `filename` or `chunkFilename`, `assetsDir` does not need to be included in their path values.
+:::
+## Disable Generating index.html
+When using Vue CLI with an existing backend, you may need to disable the generation of `index.html` so that they generated assets can be used with another default document. To do so, the following can be added to `[vue.config.js](https://cli.vuejs.org/config/#vue-config-js)`:
+
```
LinusBorg: I should be noted that `--modern` mode doesn't work without the html-webpack-plugin.

betweenbrain: Thanks! I have added a warning about that.

PR #1794, opened 2018-07-09
WIP: fix(cli): Inform earlier in promts flow about PostCSS

### LINKED ISSUE #1655: CSS Pre-processors should have a better warning about PostCSS, Autoprefixer and CSS Modules already being supported.
### Version
3.0.0-rc.3

### Reproduction link
[https://codesandbox.io/s/vue](https://codesandbox.io/s/vue)

### Steps to reproduce
`vue create hello-world`
"Manually select features"
Select "CSS Pre-processors"

### What is expected?
As someone new to the CLI, I did not know PostCSS, Autoprefixer and CSS Modules are supported by default.

After learning this, I would like an option called "None". Instead, I have to ctrl-c and then redo the wizard.

I am forced to choose from SCSS/SASS, LESS, or Stylus.

LinusBorg:
CLI should inform earlier that PostCSS, Autoprefixer and CSS Modules already being supported.
closes #1655

Just a UX improvement. Without this, people might think they have to select "yes" for "Add Pre-Processor support" to get PostCSS, only to realize that they can't (and don't have to) and must start over if they don't want to select any of the pre-processors offered.

### DIFF
--- a/packages/@vue/cli/lib/promptModules/cssPreprocessors.js
+++ b/packages/@vue/cli/lib/promptModules/cssPreprocessors.js
@@ -2,7 +2,7 @@ module.exports = cli => {
   cli.injectFeature({
     name: 'CSS Pre-processors',
     value: 'css-preprocessor',
-    description: 'Add support for CSS pre-processors like SASS, Less or Stylus',
+    description: 'Add support for CSS pre-processors like SASS, Less or Stylus (PostCSS is always included)',
     link: 'https://cli.vuejs.org/guide/css.html'
   })
 


packages/@vue/cli/lib/promptModules/cssPreprocessors.js
```diff
@@ -2,7 +2,7 @@ module.exports = cli => {
   cli.injectFeature({
     name: 'CSS Pre-processors',
     value: 'css-preprocessor',
-    description: 'Add support for CSS pre-processors like SASS, Less or Stylus',
+    description: 'Add support for CSS pre-processors like SASS, Less or Stylus (PostCSS is always included)',
```
Akryum: Note: This will only be displayed in the UI.

LinusBorg: Oh ^^

LinusBorg: ...in the UI we can go back& forth anyway, right? Then maybe we have to solve this another way.

Don't want to make the `name`prop disproportionately long compared to other prompts with this ...

pksunkara: `message` key is used by inquirer.js for the prompts.

PR #1803, opened 2018-07-10
fix docs image path

andreydos:
I found two .md files with broken images [#1](https://github.com/vuejs/vue-cli/blob/dev/docs/dev-guide/ui-api.md) and [#2](https://github.com/vuejs/vue-cli/blob/dev/docs/dev-guide/ui-info.md).
In this PR I update path for this resources


Akryum:
The docs are meant to be red here: https://cli.vuejs.org/dev-guide/ui-api.html
Thanks any way!

PR #1818, opened 2018-07-11
Fix typo

betweenbrain:
Fixes header level of last section of document


Akryum:
Good catch! Thanks! 😸

Issue #1616, opened 2018-06-19
Vue thinks local IP address is global, causing several issues


kissge:
### Version
3.0.0-rc.2

### Reproduction link
[https://github.com/kissge/vue-cli-bug](https://github.com/kissge/vue-cli-bug)

### Steps to reproduce
1. Using `vue create` to start a project, on a remote machine (say, AWS EC2 instance)
2. Run `npm run serve`

### What is expected?
- Global IP address is shown in console
- HMR works
- App is accessible from the internet via http://mydomain.example.com/:8080

### What is actually happening?
- Local IP address is shown in console
- HMR fails because sockjs tries to connect to local IP address
- App cannot be accessed from outside the network, so I have to use tricks like ssh port forwarding

---
I also tried tweaking webpack config from vue.config.js, with no luck.

LinusBorg:
The devserver is meant for local development and is not accessible from outside by default. 

However, you can adjust this behaviour in vue.config.js with the `devServer` option. 

https://cli.vuejs.org/config/#devserver

Options to look into: 

https://webpack.js.org/configuration/dev-server/#devserver-disablehostcheck

https://webpack.js.org/configuration/dev-server/#devserver-host

LinusBorg:
the devServer options have their own options property in vue.config.js because we process them interlly before handing them to the devserver:
```javascript
module.exports = {

module.exports = {
  configureWebpack: {
    // other webpack options to merge in ...
  },
  // devServer Options don't belong into `configureWebpack`
  devServer: {
    host: '0.0.0.0',
    hot: true,
    disableHostCheck: true,
  },
};
```

kissge:
Oops, my fault... thanks for pointing out.
But still, two other problems persist.
I think `address.ip()` used [here](https://github.com/vuejs/vue-cli/blob/a9e1286/packages/%40vue/cli-service/lib/util/prepareURLs.js#L35) is something meant to get local IP address, unlike other things like [this](https://github.com/sindresorhus/public-ip) and [this](https://github.com/alsotang/externalip)?

- Local IP address is shown in console
- HMR fails because sockjs tries to connect to local IP address
- ~App cannot be accessed from outside the network, so I have to use tricks like ssh port forwarding~

LinusBorg:
The code you mention should not be running if I set the `host` option (to your external IP)

kissge:
[...]

PR #1909, opened 2018-07-20
Added interface parameter

padiazg:
Now we can use ```--interface <inteface name>``` when calling to ```vue-cli-service serve```, or set an environment variable called ```INTERFACE``` with the interface name.

Ej:
package.json
```json
"scripts": {
    "serve": "vue-cli-service serve --progress --interface Ethernet",
    "build": "vue-cli-service build"
  },
```

To show a list of the interfaces you have:
```javascript
const os = require('os');
const interfaces = os.networkInterfaces();
console.log('interfaces =>', interfaces);
```


yyx990803:
Hey, thanks a lot for this PR. After investigating the issue, I think it's better to explicitly specify your public network address via `devServer.public`. See https://github.com/vuejs/vue-cli/issues/1616#issuecomment-407439867

PR #1968, opened 2018-07-28
Typo

fengtan:
There seems to be a typo when displaying the usage of the command -- see changeset.


Akryum:
Thanks!

yyx990803:
@Akryum in the future when we merge doc fixes make sure to mark them with `docs:` instead of `fix:` because they are not bug fixes.

PR #1994, opened 2018-07-31
chore(deps): update Prettier to 1.14

Kocal:
It has been released this weekend: https://prettier.io/blog/2018/07/29/1.14.0.html


yyx990803:
Thanks, but since this is a minor version, there's no need to explicitly bump it alone. (We do periodic batch upgrades)

PR #1998, opened 2018-07-31
WIP: Fix(eslint): Set parserOptions.ecmaVersion to 2017

LinusBorg:
This keeps babel-eslint from complaining about newer syntax like async/await


yyx990803:
Is this necessary? I've never seen it complaining about async/await.

LinusBorg:
came across it here: https://forum.vuejs.org/t/how-to-use-async-await-in-vue-cli-3/39733

yyx990803:
This should only be necessary if not using babel-eslint.

LinusBorg:
Okay, you're right. Will cuss he it so that it only applies it if we create the project without Babel and typescript.

michalsnik:
@LinusBorg any update here? :)

LinusBorg:
Oh, forgot about this one ^^ Will check later.

LinusBorg:
This is too old, I'll start over

PR #2002, opened 2018-07-31
Adds css.alwaysExtract config option

Karolis2011:
I found no logical option how to override those checks. So, I added option that allows to bypass assumptions that everyone uses HMR and doesn't want to extract CSS in to separate file while in development mode.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -198,8 +198,8 @@ module.exports = {
 
 ### css.extract
 
-- Type: `boolean`
-- Default: `true` (in production mode, always `false` otherwise)
+- Type: `boolean | string`
+- Default: `'production'`
 
   Whether to extract CSS in your components into a standalone CSS files (instead of inlined in JavaScript and injected dynamically).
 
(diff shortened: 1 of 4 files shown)


Karolis2011:
I am concerning that latter suggestion might break backwards compatibility. But it would break in edge cases, like when this value is explicitly set to true. But I like it.

yyx990803:
Thanks for the PR - I'm not sure if the extra string-based value is necessary, since that is just the default behavior.

Karolis2011:
If you will point at a way how to enable extract css on development builds, without deleting and readding module sets with webpack chaining, then I will be stratified and be sure that this is unnecessary. Else I see this necessary.

yyx990803:
See my commit - I've made it so that setting it to `true` will enforce extraction even in development.

Karolis2011:
Oh sorry, didn't see commit https://github.com/vuejs/vue-cli/commit/686ec2591a4e686e2aadf36ecf666167b7607333
I was experting it to be referenced in you comment. Anyway, thank you and sorry again.

PR #2008, opened 2018-07-31
Add eslint configuration tab

### LINKED ISSUE #1423: Automatically generate ESLint prompts from eslint-plugin-vue
Currently, the prompts are hard-coded [here](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-eslint/ui.js).
We could generate the prompt list from eslint-plugin-vue directly instead, maybe using the [meta option](https://github.com/vuejs/eslint-plugin-vue/blob/master/lib/rules/attribute-hyphenation.js#L15) to add the missing information cli-ui needs.

michalsnik:
Closes #1423

This PR updates the configuration tab for `@vue/cli-plugin-eslint` in vue-ui.

Original proposition: https://github.com/vuejs/eslint-plugin-vue/issues/477

1. It adds a `general` tab with two settings, where users can both set `lint on save` and choose one of the pre-defined configs:
image

2. It add a `rules` tab with a list of all rules and categories:
image

When user selects certain category in `general` tab - lets' say `strongly-recommended` we set `error` as `default` value for all rules in `base`, `essential` and `strongly-recommended` categories.

At the very bottom we also have a list of `Uncategorized` rules, that are always off by default.

Due to the fact that it's really hard, and currently impossible to generate controls that would allow full configuration regarding actual rules' schemas we allow setting only `error`, `warning` and `off`. If the user wants he can customize rules further using `.eslintrc`, then if the rule has custom settings, we display its value in dropdown as `custom` - but this option appears only if the rule has been reconfigured.

image

Whenever user changes the configuration in `general` tab, to not introduce confusion we're not resetting any custom settings, that user might've set himself before.

cc @chrisvfritz


packages/@vue/cli-ui/src/components/ListItemInfo.vue
```diff
@@ -12,7 +12,7 @@
     </div>
     <div v-if="description || link || showDescription" class="description">
       <slot name="description">
-        <span v-html="description"/>
```
Akryum: Why this change?

michalsnik: Description of some rules contain `<tagname>` and it broked the UI 🤔

Akryum: Could we strip them out?

michalsnik: escaped

michalsnik:
Seems like because of network issues the build failed. Can you please rerun the build from CircleCI?

PR #2017, opened 2018-08-01
delete a duplicated sentence.

remote-star:
delete a duplicated sentence that wasn't translated.


yyx990803:
Thanks for the PR, already covered by #2014

PR #2022, opened 2018-08-01
Added Now deployment

ivansieder:
Updated the deployment doc to also feature now.sh


yyx990803:
Thanks @ivansieder ! I appreciate the detailed guide, but I think it's a bit too long, and for a developer who is using Now, they should be able to tell what the config fields do pretty easily. So I trimmed it down to keep it simple.

PR #2045, opened 2018-08-02
Include esnext in lib

ffxsam:
TypeScript was complaining about `Array.includes` not existing. Having `esnext` as a target isn't sufficient.

### DIFF
--- a/packages/@vue/cli-plugin-typescript/generator/template/tsconfig.json
+++ b/packages/@vue/cli-plugin-typescript/generator/template/tsconfig.json
@@ -26,10 +26,7 @@
       ]
     },
     "lib": [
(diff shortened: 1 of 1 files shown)


packages/@vue/cli-plugin-typescript/generator/template/tsconfig.json
```diff
@@ -27,6 +27,9 @@
     },
     "lib": [
       "es2015",
+      <%_ if (options.useTsWithBabel) { _%>
+      "esnext",
```
ktsn: When we include `esnext`, `es2015` is not needed because `esnext` lib file includes `es2015`.

PR #2058, opened 2018-08-03
docs: Add Now deployment guide

lucaperret:
Updated the deployment documentation to add [▲ZEIT now](https://zeit.co/now).

Alternative to #2022

### DIFF
--- a/docs/guide/deployment.md
+++ b/docs/guide/deployment.md
@@ -202,7 +202,32 @@ Please refer to the [Firebase Documentation](https://firebase.google.com/docs/ho
 
 ### Now
 
-> TODO | Open to contribution.
+Install [Now Desktop](https://zeit.co/download) or install the CLI:
+
+```bash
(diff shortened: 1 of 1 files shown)


Alex-Sokolov:
But we already can use their server, if put `now.json` in build folder with contents:

```
{
  "static": {
    "rewrites": [
      {
        "source": "**",
        "destination": "/index.html"
      }
    ]
  }
}
```

Will work with router in history mode

ivansieder:
Great and concise write-up! For curiosity, Is there an advantage to deploy it with `server` instead of deploying the static files directly?

lucaperret:
It depends on your code and dependencies. 
Building on specific node or npm engine can result differencies in the bundled code.
This is why I prefer to let now build my app.

yyx990803:
I think I'll go with #2022 due to one less local dependency to install, but thanks for the PR.

PR #2091, opened 2018-08-07
fix the wrong homepage link in package.json of cli-plugin-typescript

remote-star:
the original link cannot be opened. add 'tree/dev' into the path so it works. the same problem also occurs in other cli-plugins.


yyx990803:
Thanks for the catch! I realized we have this problem in all other packages as well so I fixed them together in d494350f.

PR #2118, opened 2018-08-09
perf: tweak uglifyOptions.js

PanJiaChen:
Removed unnecessary configuration.

Many configurations are already configured by default.

Detail see [Options](https://github.com/mishoo/UglifyJS2)

### DIFF
--- a/packages/@vue/cli-service/lib/config/uglifyOptions.js
+++ b/packages/@vue/cli-service/lib/config/uglifyOptions.js
@@ -6,30 +6,14 @@ module.exports = options => ({
       collapse_vars: false, // 0.3kb
       comparisons: false,
       computed_props: false,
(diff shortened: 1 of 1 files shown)


yyx990803:
Thanks for the PR, but these are intentionally left there for explicitness / documentation purposes.

PanJiaChen:
@yyx990803 In addition, I have a question, why should set these configurations to false by default? Is there any special reason?


```js
{
   arrows: false,
   collapse_vars: false, // 0.3kb
   comparisons: false,
   computed_props: false,
   hoist_props: false,
   inline: false,
   loops: false,
   negate_iife: false,
   properties: false,
   reduce_funcs: false,
   reduce_vars: false,
   switches: false
}    
```

PR #2137, opened 2018-08-10
docs: add example for routing with `history.pushState` on nginx.

biluochun:
add a demo conf of nginx


Kocal:
I'm probably wrong, but shouldn't be something like that _instead_: 

```nginx
server {
  root /path/to/dist;

  location / {
     try_files $uri $uri/ /index.html;
  }
}
```

yyx990803:
Thanks, but as @Kocal pointed out, it's already included in the linked [vue-router docs](https://router.vuejs.org/guide/essentials/history-mode.html#example-server-configurations).

PR #2163, opened 2018-08-11
fix: remove deprecated opptions in babel-plugin-transform-runtime 7

clarkdo:
Per [babel-plugin-transform-runtime](https://babeljs.io/docs/en/next/babel-plugin-transform-runtime#polyfill), these options have been removed in v7 by just making it the default.


yyx990803:
I think this is a recent introduction in 7.0.0-beta.56, and we are waiting for babel 7 official release, so we will do this as part of the upgrade.

PR #2205, opened 2018-08-14
fix(styleguide): add the prefix 'The' on single instance SFC name

### LINKED ISSUE #2203: Components names doesn't respect the style guide
### Version
3.0.0

### Node and OS info
Node 8.11.3 / yarn 1.9.4 / Linux Solus

### Steps to reproduce
Create a project with vue-router.

### What is expected?
The single instance component names should start with 'The' as in the style guide.
https://vuejs.org/v2/style-guide/#Single-instance-component-names-strongly-recommended

The 2 vue SFC names Home and About doesn't start with 'The'.

---
Did I understand something wrong or the components names should be TheHome and TheAbout?

maxirozay:
I added the prefix 'The' to single instance SFCs like HelloWorld, Home and About as the [style guide](https://vuejs.org/v2/style-guide/#Single-instance-component-names-strongly-recommended) strongly recommend.
close #2203


maxirozay:
All the test pass when I test locally but the CI has an error with yarn. Should I commit my `yarn.lock` file even tough I didn't change anything to the packages?

My `yarn.lock` removed from jest-validate@^23.0.0 (there's is still jest-validate@^23.4.0) and lint-staged@^7.2.0.

yyx990803:
Hmm, I'm personally not so sure about this rule... `HelloWorld` is just a demo component that is meant to be removed.

maxirozay:
What about the Home and About page? Should we ignore this rule ?

That confuse me when I read the style guide and see that the official template does something else.

PR #2229, opened 2018-08-15
Create .editorconfig

Alex-Sokolov:
This PR add [`.editorconfig` file](https://editorconfig.org/) which will autofix trailing spaces (common case in documentation), charsets and end-of-line styles when IDE supported it (native or via plugin)

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,9 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true


LinusBorg:
```
indent_style = space
indent_size = 2
```

That's probably something that will make 50% of developmers mad as hell :D

We should leave that to teh developer to decide.

Kocal:
@LinusBorg both Vue and VuePress use a `.editorconfig` with:
```
indent_style = space
indent_size = 2
```

why shouldn't be the case here? 👀

LinusBorg:
Oh, I misread the change. I thought it was added to the default template for a fresh project. 

It's probably fine for the repo itself.

PR #2248, opened 2018-08-16
fix(babel-jest): add parantheses to babel-jest regex for jsx and js f…

ph1p:
Hi!
Today I got this error by importing a **.js** file.

```bash
    SyntaxError: Unexpected token import

      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/script_transformer.js:403:17)
```
 It's described already at this thread: https://github.com/facebook/jest/issues/4604

I fixed it by adding parantheses to the jsx RegExp. 
This is a really strange behavior from jest, because the regex is completely correct.


yyx990803:
Thanks but I don't think this is the actual cause of the problem.

PR #2253, opened 2018-08-16
Use gitignore best practices

zachmullen:
The main purpose of this is to remove platform specific entries from the gitignore template. Namely, OS-specific and editor-specific files belong in developers' global gitignore configuration, not the source repository gitignore, since they do not pertain to the project code itself.

This is also in the interest of keeping the template-generated surface of vue-cli as minimal as possible, since those parts do not automatically get updated when upstream improvements are made.

### DIFF
--- a/packages/@vue/cli-service/generator/template/_gitignore
+++ b/packages/@vue/cli-service/generator/template/_gitignore
@@ -1,5 +1,4 @@
-.DS_Store
-node_modules
+node_modules/
 /dist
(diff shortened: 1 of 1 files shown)


yyx990803:
I'm not sure about this - mostly because not every dev has global gitignore configured, and the same project may be cloned by different developers. Having these explicitly existed doesn't hurt and imo outweighs the trouble of having to explain users to configure global git ignore.

PR #2265, opened 2018-08-18
doc: add tip about `--name`

ishowman:
add tip about --name in build web component command


yyx990803:
Thanks for the PR but I think the error message is already pretty clear... this tip seems a bit redundant.

PR #2271, opened 2018-08-18
fix: Use relative paths in manifest.json for compatibility with Github Pages

NoelDeMartin:
When deploying applications using Github Pages, and other scenarios, absolute paths don't work because the website is served under a subdirectory. But using relative paths should work in all scenarios where `manifest.json`, `index.html` and the icons folder are on the same location. Which should be the case with the generated files if they are not modified.

In case this is not accepted, it should at least be documented because the current setup is generating PWAs that don't have proper icons in Github Pages. This same thing happens with vuepress, so once this is resolved I'll do the same on the other repo.

### DIFF
--- a/packages/@vue/cli-plugin-pwa/generator/template/public/manifest.json
+++ b/packages/@vue/cli-plugin-pwa/generator/template/public/manifest.json
@@ -3,17 +3,17 @@
   "short_name": "<%- rootOptions.projectName %>",
   "icons": [
     {
-      "src": "/img/icons/android-chrome-192x192.png",
+      "src": "img/icons/android-chrome-192x192.png",
       "sizes": "192x192",
       "type": "image/png"
     },
     {
-      "src": "/img/icons/android-chrome-512x512.png",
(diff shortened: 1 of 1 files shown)


Kocal:
Looks good to me. 
By the way what do you think about prefixing relative paths with `./`? It would be more clear no?

NoelDeMartin:
@Kocal If you think that's better I don't mind. For web resources I am used to declare relative paths without the `./`, but it's true that it is more explicit so I guess it's an unfunded habit (writing less I guess, which is no real justification).

Should I commit that change then? If you confirm it I'll go ahead.

Kocal:
Well it's just my own opinion, as you said it's more explicit (and [Explicit is better than Implicit](https://www.python.org/dev/peps/pep-0020/#id3) :eyes:).
If it's improve readability and if your fix is still working, so yes it would be fine to commit that change, thank you :slightly_smiling_face:

NoelDeMartin:
@Kocal Ok done :+1:

PR #2337, opened 2018-08-23
More flexible hook system

### LINKED ISSUE #1754: Hook to act after others plugins invocation
(Plugins dev API)

I don't know if this already exists or possible, but I need a method to listen other plugins 
invocation inside my plugin, to alter files contents.

Simple use case:

I want to replace default import of PWA plugin (`import './registerServiceWorker'`) to :

```js
if (process.client) require('./registerServiceWorker');
```

ATM, I use `api.postProcessFiles()` function, but this one only work if I do:

```bash
vue add @vue/pwa
vue add my-plugin
```

But if I do

```bash
vue add my-plugin
vue add @vue/pwa
```

`my-plugin` is not aware that PWA was added/invoked, so I cant transform the import statement.

Something like `api.postProcessFiles()`

```js
api.hookPluginInvoke('pluginName', (files) => {})
```

Thanks !

### LINKED ISSUE #1938: Plugin depending on another plugin
Some plugins might want to depend on other plugins. Currently, there is no way to do this.

If the `package.json` `dependencies` of `vue-cli-plugin-a` contains `vue-cli-plugin-b`, then:

1. `vue add a` will run `vue add b` and then `vue add a`
2. `vue invoke a` will run `vue invoke b` and then `vue invoke a` (Optionally configured by `a`)

We can do this multiple deep levels.

### Issues

1. If `a` depends on `b` and `c`, what's the order of `b` and `c` to be run?
2. What about circular dependencies?

pksunkara:
Fixes #1754, Fixes #1938

Implements the discussion regargding hooks from https://gist.github.com/sodatea/37bc8c36d7d0c285dfea52fbe8ddaf90

### DIFF
--- a/packages/@vue/cli/lib/Generator.js
+++ b/packages/@vue/cli/lib/Generator.js
@@ -1,12 +1,14 @@
 const ejs = require('ejs')
 const debug = require('debug')
+const semver = require('semver')
 const GeneratorAPI = require('./GeneratorAPI')
+const PackageManager = require('./util/ProjectPackageManager')
 const sortObject = require('./util/sortObject')
 const writeFileTree = require('./util/writeFileTree')
 const inferRootOptions = require('./util/inferRootOptions')
 const normalizeFilePaths = require('./util/normalizeFilePaths')
 const runCodemod = require('./util/runCodemod')
-const { toShortPluginId, matchesPluginId, loadModule } = require('@vue/cli-shared-utils')
+const { toShortPluginId, matchesPluginId, loadModule, isPlugin } = require('@vue/cli-shared-utils')
 const ConfigTransform = require('./ConfigTransform')
 
 const logger = require('@vue/cli-shared-utils/lib/logger')
@@ -69,7 +71,6 @@ module.exports = class Generator {
   constructor (context, {
     pkg = {},
     plugins = [],
-    otherPlugins = [],
     afterInvokeCbs = [],
     afterAnyInvokeCbs = [],
     files = {},
@@ -77,9 +78,9 @@ module.exports = class Generator {
   } = {}) {
     this.context = context
     this.plugins = plugins
-    this.otherPlugins = otherPlugins
     this.originalPkg = pkg
     this.pkg = Object.assign({}, pkg)
+    this.pm = new PackageManager({ context })
     this.imports = {}
     this.rootOptions = {}
     this.afterInvokeCbs = []
@@ -98,13 +99,20 @@ module.exports = class Generator {
     this.exitLogs = []
 
     const pluginIds = plugins.map(p => p.id)
+
+    // load all the other plugins
+    this.otherPlugins = Object.keys(this.pkg.dependencies || {})
+      .concat(Object.keys(this.pkg.devDependencies || {}))
+      .filter(isPlugin)
+      .filter(id => pluginIds.indexOf(id) === -1)
+
     const cliService = plugins.find(p => p.id === '@vue/cli-service')
     const rootOptions = cliService
       ? cliService.options
       : inferRootOptions(pkg)
 
     // apply hooks from all plugins
-    otherPlugins.forEach(id => {
+    this.otherPlugins.forEach(id => {
       const api = new GeneratorAPI(id, this, {}, rootOptions)
       const pluginGenerator = loadModule(`${id}/generator`, context)
       if (pluginGenerator && pluginGenerator.hooks) {
@@ -265,11 +273,22 @@ module.exports = class Generator {
     debug('vue:cli-files')(this.files)
   }
 
-  hasPlugin (_id) {
+  hasPlugin (_id, _version) {
     return [
       ...this.plugins.map(p => p.id),
       ...this.otherPlugins
-    ].some(id => matchesPluginId(_id, id))
+    ].some(id => {
+      if (!matchesPluginId(_id, id)) {
+        return false
+      }
+
+      if (!_version) {
+        return true
+      }
+
+      const version = this.pm.getInstalledVersion(id)
+      return semver.satisfies(version, _version)
+    })
   }
 
   printExitLogs () {
--- a/packages/@vue/cli/lib/Generator.js
+++ b/packages/@vue/cli/lib/Generator.js
@@ -6,7 +6,7 @@ const writeFileTree = require('./util/writeFileTree')
 const inferRootOptions = require('./util/inferRootOptions')
 const normalizeFilePaths = require('./util/normalizeFilePaths')
 const runCodemod = require('./util/runCodemod')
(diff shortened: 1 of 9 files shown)


pksunkara:
@sodatea Can I get some feedback on this?

sodatea:
I'm currently working on a patch release.
Will look into this later for the next minor version.

LinusBorg:
@sodatea Is this still on the rosdmap in your opinion?

sodatea:
@LinusBorg Yeah I’ve talked to Evan before. He may have some additional ideas on the API design.

packages/@vue/cli/lib/Generator.js
```diff
@@ -242,12 +275,22 @@ module.exports = class Generator {
...
+    ].some(id => {
+      if (!matchesPluginId(_id, id)) {
+        return false
+      }
+
+      if (!_version) {
+        return true
+      }
+
+      const version = this.pkg.dependencies[id] || this.pkg.devDependencies[id]
```
sodatea: This typically returns a semver range like `^3.9.0` and is not a valid version to be used as the first argument of `semver.satisfies`.
Better use [getInstalledVersion](https://github.com/vuejs/vue-cli/blob/4af27247ccabe2ba5da57b6f216d17ad0767b8a8/packages/%40vue/cli/lib/util/ProjectPackageManager.js#L151) and do some workaround in the unit tests.

packages/@vue/cli/lib/GeneratorAPI.js
```diff
@@ -133,10 +133,11 @@ class GeneratorAPI {
    * Check if the project has a given plugin.
    *
    * @param {string} id - Plugin id, can omit the (@vue/|vue-|@scope/vue)-cli-plugin- prefix
+   * @param {string} version - Plugin version. Defaults to ''
    * @return {boolean}
    */
-  hasPlugin (id) {
-    return this.generator.hasPlugin(id)
+  hasPlugin (id, version = '') {
```
sodatea: I think `undefined` is fine here…

packages/@vue/cli/lib/add.js
```diff
@@ -27,12 +26,7 @@ async function add (pluginName, options = {}, context = process.cwd()) {
   log(`${chalk.green('✔')}  Successfully installed plugin: ${chalk.cyan(packageName)}`)
   log()
 
-  const generatorPath = resolveModule(`${packageName}/generator`, context)
-  if (generatorPath) {
-    invoke(pluginName, options, context)
-  } else {
-    log(`Plugin ${packageName} does not have a generator to invoke`)
```
sodatea: Why remove this?

packages/@vue/cli/lib/Generator.js
```diff
@@ -93,14 +95,45 @@ module.exports = class Generator {
...
     const rootOptions = cliService
       ? cliService.options
       : inferRootOptions(pkg)
+
+    // apply hooks from all plugins
+    this.otherPlugins.forEach(id => {
+      const api = new GeneratorAPI(id, this, {}, rootOptions)
+      const pluginGenerator = loadModule(`${id}/generator`, context)
+      if (pluginGenerator && pluginGenerator.hooks) {
+        pluginGenerator.hooks(api, {}, rootOptions, pluginIds)
```
sodatea: With this design & implementation, `afterInvoke` callbacks from other plugins also gets pushed into the callback arrays, this is not the expected behavior.
I think the `hooks` feature should be more restrictive. Object syntax may be a better alternative.

packages/@vue/cli/lib/add.js
pksunkara: This is just an useless duplicate. `invoke` already does the same check at https://github.com/pksunkara/vue-cli/blob/hook/packages/%40vue/cli/lib/invoke.js#L64 pretty early.

sodatea: Ah that's true, thanks.

packages/@vue/cli/lib/Generator.js
pksunkara: As you can see here, I am resetting the non-any hooks. https://github.com/vuejs/vue-cli/pull/2337/files#diff-f0e2bfadbc7f0f35b5cc089613387802R126.

sodatea: Okay.

packages/@vue/cli/lib/invoke.js
```diff
@@ -132,9 +135,18 @@ async function runGenerator (context, plugin, pkg = getPkg(context)) {
...
-    for (const cb of createCompleteCbs) {
+    for (const cb of afterInvokeCbs) {
+      await cb()
+    }
+    stopSpinner()
+    log()
+  }
+
+  if (afterAnyInvokeCbs.length) {
+    logWithSpinner('⚓', `Running completion hooks from other plugins...`)
```
sodatea: Does this include the `afterAnyInvoke` hook from the currently running generator?

pksunkara: Yes, it does.

pksunkara: From all the current plugins.

pksunkara:
Updated the PR

packages/@vue/cli/lib/invoke.js
sodatea: But the order of `afterAnyInvoke` callbacks is first all other plugins' hook, then the current plugin's `afterAnyInvoke` hook.

The exact result depends on which plugin did you invoke first.
This looks confusing to me.
It is the same kind of non-determinism like `npm install`. We should avoid it.

sodatea: Also the log is misleading in that case.

pksunkara: Fixed this. Current plugins will run first and then others. The order is now documented.

sodatea: It no longer triggers other plugins' `afterAnyInvoke` hook as I tested.

I'm also a little bit skeptical about this approach - this is still non-deterministic IMO.
For example, if a user has two plugins with `afterAnyInvoke` hook, say, `eslint` and `typescript`, the order will be:

- `eslint` -> `typescript` on creation
- `eslint` -> `typescript` if user add `eslint` first
- `typescript` -> `eslint` if user add `typescript` first

My idea is to keep the order of `afterAnyInvoke` in sync with that of plugin loading (for now, it is the key enumeration order in `package.json`) and then figure out a way to specify plugin order in a separate PR. What do you think?

pksunkara: Good idea. I didn't think of it like that. Made it completely deterministic like how you said. Should be working now.

PR #2355, opened 2018-08-25
[polish @vue/cli-init]

dailynodejs:
using argv for reuse process.argv and initIndex for init command in array search, also add if for code safe


dailynodejs:
@Akryum can you see this mr, thx~

PR #2422, opened 2018-09-01
Instructions about History Pushstate in Netlify

hmaesta:
Included instructions for users that use `History Pushstate mode` on Vue Router and run application on Netlify.

### DIFF
--- a/docs/guide/deployment.md
+++ b/docs/guide/deployment.md
@@ -133,6 +133,15 @@ Commit both the `.gitlab-ci.yml` and `vue.config.js` files before pushing to you
 
 Also checkout [vue-cli-plugin-netlify-lambda](https://github.com/netlify/vue-cli-plugin-netlify-lambda).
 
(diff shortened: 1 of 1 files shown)


docs/guide/deployment.md
```diff
@@ -133,6 +133,15 @@ Commit both the `.gitlab-ci.yml` and `vue.config.js` files before pushing to you
 
 Also checkout [vue-cli-plugin-netlify-lambda](https://github.com/netlify/vue-cli-plugin-netlify-lambda).
 
+In order to receive direct hits using `History Pushstate mode` on Vue Router, you need to create a file called `_redirects` under `/public` with the following content:
```
LinusBorg:
```suggestion
In order to receive direct hits using `history mode` on Vue Router, you need to create a file called `_redirects` under `/public` with the following content:
```
It's commonly referred to as "`history` mode" in our docs etc, so we should keep it that way.

hmaesta:
👍

PR #2428, opened 2018-09-03
 feat: implement vue upgrade

### LINKED ISSUE #1507: Command: vue upgrade
Command that should be run inside an existing project with `@vue/cli-service` present.

- Resolve the version to upgrade to.
  - `vue upgrade [patch|minor|major]`: defaults to minor
- If already latest, print message and exit
- Otherwise, confirm via prompt
- Find `@vue/cli-service` plus all packages in `package.json` that starts with `@vue/cli-plugin-`, and bump them to the target upgrade version
- Run `installDeps` - infer package manager from lockfile, if no lockfile, infer from `.vuerc`, otherwise infer using `hasYarn()`
- Print changelog of upgraded versions

sodatea:
closes #1507 

TODOs:
- [ ] Add tests
- [ ] Implement plugin upgrade hook


packages/@vue/cli-upgrade/get-upgradable-version.js
```diff
@@ -0,0 +1,38 @@
+const execa = require('execa')
+
+function getMaxSatisfying (packageName, range) {
+  let version = JSON.parse(
+    execa.shellSync(`npm view ${packageName}@${range} version --json`).stdout
```
visualfanatic: What if `npm` is not installed?

sodatea: But how is it possible?

LinusBorg: ...but then you can't really run a vue-cli project, so you have bigger problems than `vue upgrade` not working.

PR #1531, opened 2018-06-09
feat: support PNPM as a package manager

robertkruis:
Enables vue-cli to use PNPM (https://pnpm.js.org/) as package manager

PR #2472, opened 2018-09-09
Add pnpm package manager

znck:
https://pnpm.js.org


Kocal:
What about #1531?

Akryum:
#1531 seems more complete

PR #2487, opened 2018-09-11
feat: allow devServer.contentBase to be overwritten

m0dch3n:
If you add i.e. cordova to your project, you need a static access to cordova.js, which i.e. is under 
src-cordova/platforms/android/platform_www/

With this change, you can do:

contentBase: [api.resolve('public'), api.resolve('src-cordova/platforms/android/platform_www/')]

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -84,6 +84,7 @@ module.exports = (api, options) => {
         ? rawPublicUrl
         : `${protocol}://${rawPublicUrl}`
       : null
+    const contentBase = args.contentBase || projectDevServerOptions.contentBase || api.resolve('public')
 
     const urls = prepareURLs(
       protocol,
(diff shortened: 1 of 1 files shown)


LinusBorg:
you can already overwrite this in the `devServer` options which we expose in vue.config.js, so I don't see a use in adding this as a separate option.

m0dch3n:
Yes, exactly, but I think this is not working because of this hard coded line

https://github.com/vuejs/vue-cli/blob/2ab93e797008aee9e36dfb8cc2015c8a9354dd6c/packages/%40vue/cli-service/lib/commands/serve.js#L146

at least, that's what I currently remember, when I opened this issue, I should give this a retry...

m0dch3n:
https://github.com/vuejs/vue-cli/blob/2ab93e797008aee9e36dfb8cc2015c8a9354dd6c/packages/%40vue/cli-service/lib/commands/serve.js#L76-L99

these are the only possible options I think

Xantier:
@LinusBorg I believe this issue is still open judging from the hardcoded line shown above. This would have fixed bug https://github.com/vuejs/vue-cli/issues/5070.

PR #2503, opened 2018-09-12
Add module concatenation hint

sirlancelot:
Allow Webpack to process this file as an ES Module, which allows
concatenating using ModuleConcatenationPlugin, resulting in a slightly
smaller bundle and removes the overhead of a runtime `require()`.

### DIFF
--- a/packages/@vue/cli-service/lib/commands/build/setPublicPath.js
+++ b/packages/@vue/cli-service/lib/commands/build/setPublicPath.js
@@ -6,3 +6,6 @@ if (typeof window !== 'undefined') {
     __webpack_public_path__ = i[1] // eslint-disable-line
   }
 }
+
+// Indicate to webpack that this file can be concatenated
+export default null


sirlancelot:
The `export default null` will get optimized away during tree-shaking.

sirlancelot:
Before:
```js
// EXTERNAL MODULE: ./node_modules/@vue/cli-service/lib/commands/build/setPublicPath.js
var setPublicPath = __webpack_require__("1eb2");
```
Webpack module:
```js
/***/ "1eb2":
/***/ (function(module, exports, __webpack_require__) {

// This file is imported into lib/wc client bundles.

if (typeof window !== 'undefined') {
  let i
  if ((i = window.document.currentScript) && (i = i.src.match(/(.+\/)[^/]+\.js$/))) {
    __webpack_require__.p = i[1] // eslint-disable-line
  }
}


/***/ }),
```

After:
```js
// CONCATENATED MODULE: ./node_modules/@vue/cli-service/lib/commands/build/setPublicPath.js
// This file is imported into lib/wc client bundles.

if (typeof window !== 'undefined') {
  let i
  if ((i = window.document.currentScript) && (i = i.src.match(/(.+\/)[^/]+\.js$/))) {
    __webpack_require__.p = i[1] // eslint-disable-line
  }
}

/* harmony default export */ var setPublicPath = (null);
```

PR #2533, opened 2018-09-16
feat(docs): add Heroku deployment section

NataliaTepluhina:
Adds Heroku deployment instructions to docs `Deployment` chapter.


mathieutu:
Hi, thank you for this work !
However, I think the express part will not work with vue-router.

Also, I'm working on a vue-cli plugin that will permit to do that in a click: https://github.com/mathieutu/vue-cli-plugin-express.
I'll let you know when it'll be usable, ans we can integrate it to the doc.

NataliaTepluhina:
> However, I think the express part will not work with vue-router.
Hi! Here is an application deployed with this instruction: [https://vue-rx-presentation.herokuapp.com/search](https://vue-rx-presentation.herokuapp.com/search)
It has vue-router in history mode.

Could you please provide more details about possible router issues? Thanks ;)

mathieutu:
Hey @NataliaTepluhina.
You're example is perfect, as it doesn't work! 😅 

I let you click on the link you provided 😉 

It's because the `search` file doesn't exist. Express cannot serve it.

If you want to give to my plugin a try, and give me some feedback, it would be awesome! I've just release it, so a simple `vue add express` and `yarn express:run` should work out of the box.

ps: Long life to the 🦊!

NataliaTepluhina:
@mathieutu got it now, its the issue with deep links :) Posted it because I got redirected to `search` on load and it worked fine ofc ;)

Let me try the plugin
(and thanks!)

NataliaTepluhina:
Fixed instruction to include `connect-history-api-fallback` for vue-router deep links.

mathieutu:
ok, thanks. I will investigate.
An other thing: the build step should not be done in local and commited, but should be handled by the heroku build, with the heroku env variables.

For that, we have to add a `"heroku-postbuild": "npm run build"` line in the package.json.

LinusBorg:
@NataliaTepluhina Is this ready?

mathieutu:
@LinusBorg No, there are advices in this documentation to commit the dist folder, which is not a proper way to do. 

A heroku-postbuild should be used @NataliaTepluhina.

I'm still convinced that we could use the plugin, which is done for that, and must easy to set for people (but as I'm writing that, I didn't try to reproduce the issue 😅, I really need to).

NataliaTepluhina:
@mathieutu Thanks for mentioning heroku postbuild option, added. But you're wrong about building locally and pushing `dist`, if you read the documentation, you will see that local build was used only once just to check if server runs locally. There is no local build step in the actual deployment process.
@LinusBorg yes, now it's ready

mathieutu:
Perfect! Now with postbuild and dist ignored it's good to publish for me!

NataliaTepluhina:
Closing as outdated (also we have a bunch of new PRs on the same topic)

PR #2535, opened 2018-09-17
docs: fix html asset preload example

st-sloth:
`fileBlacklist` is an array of regexes, so a regex must be pushed, without being wrapped in another array, otherwise currently error is being thrown for this example:
```
TypeError: regex.test is not a function
  - index.js:58 options.fileBlacklist.options.fileBlacklist.every.regex
    .../[@vue]/preload-webpack-plugin/src/index.js:58:58
```


sodatea:
😂 Didn't see this in time. Already fixed with https://github.com/vuejs/vue-cli/commit/9757f144e68ff67cce7c7656096119ee9a0db4ab

PR #2568, opened 2018-09-21
feat(ui): add --host option

ykzts:
```console
$ vue ui --host 0.0.0.0
🚀  Starting GUI...
🌠  Ready on http://localhost:8000
```

Currently `vue ui` can not connect from outside. This is because it is bound to localhost.

In this Pull Request, an option is added so that it can be changed to host other than localhost.

### DIFF
--- a/packages/@vue/cli/bin/vue.js
+++ b/packages/@vue/cli/bin/vue.js
@@ -115,7 +115,7 @@ program
 program
   .command('ui')
   .description('start and open the vue-cli ui')
-  .option('-h, --host <host>', 'Host used for the UI server (default: localhost)')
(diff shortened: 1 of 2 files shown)


ykzts:
Unfortunately, if graphql versions do not match, you get an error :cry:

ykzts:
> Thanks for the PR! Could you remove the unrelated dependency changes from it please? Thanks!

done!

ykzts:
rebased

liugang:
It's best to change -h to -H, because `-h` is use for  `-h, --help         output usage information`

```
option('-h, --host <host>', 'Host used for the UI server (default: localhost)')
```

PR #2597, opened 2018-09-25
feat(ui): add --host option for vue ui

liugang:
We sometimes develop on a remote server, which makes it impossible to access vue ui via IP, so add an option that allows the user to set the host


kazupon:
That's PR is already existing at #2568 
Thanks!

PR #2612, opened 2018-09-26
Adding host option to vue ui tool

ben-farrelly:
- This allows for using `vue ui` in virtual development environments (like vagrant for example).


kazupon:
That's PR is already existing at #2568
Thanks!

PR #2666, opened 2018-10-02
Update creating-a-project.md

simioluwatomi:
Always remembering to use the `winpty vue.cmd create hello-world` syntax to create vue projects is kind of clunky as that is not the default syntax; someone like me will often forget.

Letting users know that they can alias the `winpty vue.cmd` command to something that feels more intuitive is better in my opinion.


sodatea:
Thanks!

PR #2671, opened 2018-10-03
fix(cli-service): Allow relative sockjs URL (close #1525)

### LINKED ISSUE #1525: Allow relative sockjs URL

JohnCampionJr:
This would simplify HMR use across multiple domain / port local dev server scenarios, while still allowing specific URLs to be selected by the public property of devServerOptions.

Added option to vue.config.js, not the devServer as that object is validated against Webpack

Default to false so as to not be a breaking change.

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -84,6 +84,7 @@ module.exports = (api, options) => {
         ? rawPublicUrl
         : `${protocol}://${rawPublicUrl}`
       : null
+    const relativeSockJsUrl = options.relativeSockJsUrl
 
     const urls = prepareURLs(
       protocol,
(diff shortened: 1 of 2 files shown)


JohnCampionJr:
Not sure what happened with 2nd AppVeyor test - first one ran fine and tests pass on my machine (ran them before commit)

PR #2724, opened 2018-10-12
chore:  add extend config parameter for "types" property in "tsconfig.json"

RequireSun:


pksunkara:
@RequireSun Can you rebase this? I will get this merged.

sodatea:
What's the use case of this feature?

pksunkara:
I would say they want to have a common preset which by default has some types.

PR #2753, opened 2018-10-16
fix: Add missing package in @vue/cli-ui

zhiyingzzhou:
The apollo-server/util/db.js  file in @vue/cli-ui uses the @vue/cli, which is not currently included in the list of dependencies. This causes an error when trying to run apollo server.

image

### DIFF
--- a/packages/@vue/cli-ui/package.json
+++ b/packages/@vue/cli-ui/package.json
@@ -55,6 +55,7 @@
     "watch": "^1.0.2"
   },
   "devDependencies": {
+    "@vue/cli": "^3.0.5",
     "@vue/cli-plugin-babel": "^3.0.5",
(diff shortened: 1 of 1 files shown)


pksunkara:
Could you please rebase this? I will merge it.

sodatea:
Don't.
It's an incorrect workaround and may introduce further confusion.
The error you encountered is likely due to not having bootstrapped the monorepo correctly.

The file `@vue/cli/lib/util/rcPath` should not be listed in `devDependencies` because it's required in runtime, thus making it fit in `dependencies`. However, that will introduce a cyclic dependency, which is more problematic.
So in the short term, we'll keep it as is. In the long run, these utils should be extracted to a separate package, like, in the ongoing `refactor-ui` branch https://github.com/vuejs/vue-cli/tree/refactor-ui/packages/%40vue/cli-global-utils

PR #2779, opened 2018-10-20
feat(tslint): Support multiple configs

zachary-kuhn:
## Cause

When I attempted to write Cypress tests in TypeScript, I discovered I [had to place a tsconfig.json file within the tests/e2e folder](https://docs.cypress.io/guides/tooling/typescript-support.html#Configure-tsconfig-json) in order to get Cypress to run without throwing its own errors. While that worked for running the file, this setup would cause `vue-cli-service lint` to still show those linting problems. This was not a problem, however, when running `tslint` directly from the command line.

## Changes

- The previous setup assumes a top-level tsconfig.json file and specifies what globs will be used when running `vue-cli-service lint` with `tslint`. This changes to look for any tsconfig.json file and to use the config file's "include", "exclude", and "files" properties
- When linting a file, use the closest tslint.json to determine its linting configuration
- Add tests for many of these cases involving the interaction of nested Typescript and linting configs as well as specifying files in the command line

## Concerns

- These changes should make linting with tslint much slower as I didn't do any optimizations. If requested, I'm certainly willing to work to optimize this code
- I tried to recreate [the specified behavior of the TypeScript compiler's interpretation of tsconfig.json](https://www.typescriptlang.org/docs/handbook/tsconfig-json.html) and how its commands behave, but I don't have enough experience with TypeScript to be certain it behaves as it should. I initially tried to use the TypeScript library's handling for determining files from tsconfig.json, but my experience was it ignored .vue files despite the config including them.

## Possible Future Enhancements

- Changing `vue create` to automatically set up Cypress to use TypeScript if both Cypress and TypeScript options are selected

### DIFF
--- a/packages/@vue/cli-plugin-typescript/__tests__/tsPluginTSLint.spec.js
+++ b/packages/@vue/cli-plugin-typescript/__tests__/tsPluginTSLint.spec.js
@@ -88,3 +88,132 @@ test('should ignore issues in node_modules', async () => {
   await run('vue-cli-service lint')
   expect(await read('node_modules/bad.ts')).toMatch(updatedMain)
 })
+
+test('should handle nested tsconfig.json files', async () => {
+  const project = await create('ts-lint-nested-tsconfig', {
+    plugins: {
+      '@vue/cli-plugin-typescript': {
+        tsLint: true
+      }
+    }
+  })
+
+  const { read, write, run } = project
+  const main = await read('src/main.ts')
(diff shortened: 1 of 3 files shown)


sodatea:
Thanks for the help.
But as TSLint has been [officially deprecated](https://github.com/palantir/tslint/issues/4534), we do not plan to add any new feature to this functionality and it will be removed in the next major.
So I'm closing this PR.

Thanks anyway!

PR #2795, opened 2018-10-23
fix(serve): also detect kubernetes containers

lbogdan:
While working on a Vue CLI 3 server / container template for CodeSandbox and having an HMR issue (an invalid webpack socket URL), looking at how HMR is configured in `serve.js`, I've noticed that `serve` should be detecting that it's running in a container, but it's not. Looking further at the [container detection code](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-service/lib/commands/serve.js#L296), and then at `/proc/1/cgroup` inside the container, it looks like
```
12:memory:/kubepods/burstable/pod[redacted]
11:blkio:/kubepods/burstable/pod[redacted]
[...more similar lines...]
```
, so I added `kubepods` to the detection regex.

You can see it working at https://codesandbox.io/s/qv463kv6z6 (where I manually changed `serve.js` inside `node_modules/@vue/cli-service`).

image

The "access the dev server via http://localhost..." message is still a bit confusing, I could add CodeSandbox container detection if desired (@Akryum: I pinged you on Vue Discord, maybe we can discuss this further there).

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -297,7 +297,7 @@ function checkInContainer () {
   const fs = require('fs')
   if (fs.existsSync(`/proc/1/cgroup`)) {
     const content = fs.readFileSync(`/proc/1/cgroup`, 'utf-8')
-    return /:\/(lxc|docker)\//.test(content)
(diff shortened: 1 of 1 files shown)


lbogdan:
Hm, I see the CI failed, is it possible that `AppVeyor` uses kubernetes containers? 🙂

Issue #2544, opened 2018-09-18
Pass meta for multi-page build into html-webpack-plugin


wuservices:
### What problem does this feature solve?
https://github.com/jantimon/html-webpack-plugin supports a `meta` object to inject meta tags into a page. This might be useful for setting robots, viewport, og:image or other meta tags depending on the page in a multi-page build.

By just adding `meta,` on 2 lines of `cli-service/lib/config/app.js` this seems to work fine:

```
      pages.forEach(name => {
        const {
          title,
          meta,
          entry,
          template = `public/${name}.html`,
          filename = `${name}.html`,
          chunks
        } = normalizePageConfig(multiPageConfig[name])
        // inject entry
        webpackConfig.entry(name).add(api.resolve(entry))

        // resolve page index template
        const hasDedicatedTemplate = fs.existsSync(api.resolve(template))
        if (hasDedicatedTemplate) {
          publicCopyIgnore.push(template)
        }
        const templatePath = hasDedicatedTemplate
          ? template
          : fs.existsSync(htmlPath)
            ? htmlPath
            : defaultHtmlPath

        // inject html plugin for the page
        const pageHtmlOptions = Object.assign({}, htmlOptions, {
          chunks: chunks || ['chunk-vendors', 'chunk-common', name],
          template: templatePath,
          filename: ensureRelative(outputDir, filename),
          title,
          meta
        })
```

### What does the proposed API look like?
A config like this should allow you to inject a viewport into a layout. This way you can use the same layout with different meta information, injected in via `vue.config.js`:
```
module.exports = {
  pages: {
    index: {
      entry: 'src/pages/index/main.js',
      meta: {
        viewport: 'width=device-width, initial-scale=1, shrink-to-fit=no'
      }
    }
  }
}
```

RaphaelDDL:
Nice, with sodatea's change, I can finally use `html-webpack-template` correctly. Example:

    pages: {
        index: {
            entry: "src/main.js",
            template: require('html-webpack-template'),
            //specific html-webpack-template options
            appMountId: 'app',
            mobile:  true,
            googleAnalytics: {
                trackingId: '#####',
                pageViewOnLoad: true
            }
        },

Looking forward to 3.1.0 👍

PR #2798, opened 2018-10-23
update config pages README.md

RaphaelDDL:
Updates readme to reflect changes done on #2544 (commit 4cabf5e8c798287825908fb193d61fcc1e8d62af ) for v3.1.0.
Not sure if properly worded but with those changes, it fares an explanation that any other properties will also get passed along to `html-webpack-plugin`, not only listed properties.

An example would, for example, to pass data to template, like I illustrated on above issue, using `require('html-webpack-template')` as a template.


sodatea:
Thanks!

PR #2819, opened 2018-10-26
Add support for loading WebAssembly and ES Modules

ScreamZ:
Hello,

It is pretty common to see `.wasm` and `.mjs` files nowadays, some library start using them and this break by default in `vue-cli` without extending like this.

This pull request add support for those kind of files out-of-the-box.

**vue.config.js**
```JS
module.exports = {
  chainWebpack: config => config.resolve.extensions.prepend('.mjs').prepend('.wasm')
}
```
**Do not hesitate to ping me if you have any questions / need something.**

Regards

### DIFF
--- a/packages/@vue/cli-service/lib/config/base.js
+++ b/packages/@vue/cli-service/lib/config/base.js
@@ -38,7 +38,7 @@ module.exports = (api, options) => {
 
     webpackConfig.resolve
       .extensions
-        .merge(['.js', '.jsx', '.vue', '.json'])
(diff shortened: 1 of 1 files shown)


sodatea:
As a side note: jest currently does not support these extensions (https://github.com/facebook/jest/issues/4842). We should document it somewhere.

ScreamZ:
@sodatea What do you think of [here](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-unit-jest/README.md) ?

sodatea:
@ScreamZ Looks good to me.

Issue #2813, opened 2018-10-25
Regex in "setPublicPath.js" cause async scripts not found


zh-rocco:
### Version
3.0.5

### Reproduction link
[https://github.com/zh-rocco/vue-cli-lib-error](https://github.com/zh-rocco/vue-cli-lib-error)

### Node and OS info
Node 10.11.0 / yarn 1.10.1 / macOS 10.14

### Steps to reproduce
1. Open [the error page](https://zh-rocco.github.io/vue-cli-lib-error/error.html)
2. Open Chrome Dev Tools
3. View NetWork panel, there are some files not found

**supplement**: [the correct page](https://zh-rocco.github.io/vue-cli-lib-error/correct.html)

### What is expected?
Load async scripts correctly.

### What is actually happening?
Load async scripts failed.

---
When loading a script build by "lib" mode, I encountered a 404 error that async scripts not found.

This error is caused by [the regex](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-service/lib/commands/build/setPublicPath.js#L5).

Regex `/(. \/)[^/] \.js$/` could not match url with query, for example `http:www.example.com/some-script.js?version=1.0.0`.

Change the regex to `/(. \/)[^/] \.js(\?.*)?$/`, bug is fixed.

PR #2830, opened 2018-10-29
fix(cli-service): fix path RegEx error

zh-rocco:
This fixes [issues 2813](https://github.com/vuejs/vue-cli/issues/2813)


zh-rocco:
I don't know why I have this problem.
image

sodatea:
This failure is due to the slow execution in CI environment and can be ignored.

Issue #2868, opened 2018-11-01
增强proxy: 对齐webpack devServer的proxy


zuoxue0119:
### Version
3.1.1

### Node and OS info
Node 10.13.0

### Steps to reproduce
```
proxy: [
      {
        context: '/ajax',
        target: 'https://localhost:3000',
      },
      {
        context: function (pathname, req) {
          const env = getPathnameEnv(pathname);
          return env === 'online';
        },
        target: 'https://localhost:8081',
        changeOrigin: true,
      },
]
```

### What is expected?
增强proxy, 对齐 [webpack devServer的proxy](https://webpack.docschina.org/configuration/dev-server/)
支持proxy数组与 [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware/)中所有格式的context， 例如
```
proxy: [
      {
        context: '/ajax',
        target: 'https://localhost:3000',
      },
      {
        context: function (pathname, req) {
          const env = getPathnameEnv(pathname);
          return env === 'online';
        },
        target: 'https://localhost:8081',
        changeOrigin: true,
      },
]
```

### What is actually happening?
目前不支持数组类型的proxy， 也不支持自定义函数的context

Issue #2320, opened 2018-08-22
Add regexp for proxy context


Deraw-:
### What problem does this feature solve?
In the docs for devServer.proxy, you link to this documentation : https://github.com/chimurai/http-proxy-middleware#context-matching, which contains multiple options for context maching, like exculsion, wich, I think, would be nice to have !

### What does the proposed API look like?
```
// vue.config.js
devServer: {
	index: '',
	proxy: {
		'!\/dist\/*': {
			target: 'http://127.0.0.1:8000'
		}
	}
}
```

Deraw-:
I've juste made a small plugin https://www.npmjs.com/package/@cnamts/vue-cli-plugin-proxy to fix my issue, but it would still be nice to have this by default !

aldencolerain:
I had a really hard time with the proxy settings, there are a number of bugs.  Its also a bit confusing that the docs link to http-proxy-middleware but vue clie re-writes your options and they are subtly different. Why not just bass the proxy settings as is to the webpack dev server?

jpicton:
Couldn't agree with @aldencolerain more. After reading https://cli.vuejs.org/config/#devserver-proxy I fully expected the devServer/proxy property to be compatible with all the http-proxy-middleware options. 

Even when I realised that wasn't the case and the devServer/proxy property format was too restrictive for what I wanted to achieve, I couldn't even eschew it entirely and configure devserver via the chainWebpack property instead because it looks like anything devserver related you configure via the webpack properties is overwritten by vue-cli. Really shouldn't be this hard.

Deraw-:
Hey @aldencolerain and @jpicton! I've published `v2` of the [vue-cli-plugin-proxy](https://www.npmjs.com/package/@cnamts/vue-cli-plugin-proxy), which introduce pluginOptions, that means that you can pass any context and any options, in a simply and elegant way, try it out!

woodcoder:
I've just migrated from a previous vue cli template setup where I was using the `proxyTable` with the `filter` option to do this sort of url matching (see [old docs](https://vuejs-templates.github.io/webpack/proxy.html#url-matching)).

[...]

PR #2869, opened 2018-11-01
 增强proxy: 对齐webpack devServer的proxy

zuoxue0119:
#2868 

增强proxy, 对齐 [webpack devServer的proxy](https://webpack.docschina.org/configuration/dev-server/)
支持proxy数组与 [http-proxy-middleware](https://github.com/chimurai/http-proxy-middleware/)中所有格式的context， 例如
```
proxy: [
      {
        context: '/ajax',
        target: 'https://localhost:3000',
      },
      {
        context: function (pathname, req) {
          const env = getPathnameEnv(pathname);
          return env === 'online';
        },
        target: 'https://localhost:8081',
        changeOrigin: true,
      },
]
```

### DIFF
--- a/packages/@vue/cli-service/lib/util/prepareProxy.js
+++ b/packages/@vue/cli-service/lib/util/prepareProxy.js
@@ -11,6 +11,7 @@ const url = require('url')
 const path = require('path')
 const chalk = require('chalk')
 const address = require('address')
+const contextMatcher = require('http-proxy-middleware/lib/context-matcher')
 
 const defaultConfig = {
   logLevel: 'silent',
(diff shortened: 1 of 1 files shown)


LukeTully:
Was about to make the exact same change. For further explanation, this uses http-proxy-middleware's matching method instead of the standard regex match because http-proxy-middleware accounts for slightly differing syntax when matching contexts. Wildcards such as /**/api.js don't currently work in vue-cli's abstraction of devServer's proxy because of this, and do work with this change applied.

dousybox:
I am very sorry, I only discovered that I was mentioned here.

I closed #2947 just because I was think I found the most most relative PR for my issue at that time.

But seems I was wrong, this PR is a little off track of the [Commit Message Convention](https://github.com/vuejs/vue-cli/blob/dev/.github/COMMIT_CONVENTION.md).

So I tried to deepin with this problem and I find #2320.

When I debugged the srouce code of `http-proxy-middleware` and `vue-cli`, I came to a conclusion here: https://github.com/vuejs/vue-cli/issues/2320#issuecomment-451188475.

But I still have no idea why `vue-cli` need to forced handle the context match, so I still waitting for  someone to resolve this issue.

kayoch1n:
2022年了，这个特性还有可能加入么？

PR #2894, opened 2018-11-05
feat: allow indexPath to be null to disable generating index.html

dsanders11:
The diff for this unfortunately comes out quite messy, but it's really just wrapping the relevant code in a check for if `options.indexPath` is `null`.

Feature is useful for use-cases where you're not going to use the generated index HTML anyway.

Can add a test if required.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -87,6 +87,8 @@ module.exports = {
 
   Specify the output path for the generated `index.html` (relative to `outputDir`). Can also be an absolute path.
 
+  If `null`, the index HTML will not be generated.
+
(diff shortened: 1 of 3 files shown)


sodatea:
Thanks for your interest in Vue CLI!

However, we are not likely to add this feature in Vue CLI, for these reasons:
1. Such use cases are not encouraged , see https://cli.vuejs.org/guide/html-and-static-assets.html#disable-index-generation
2. It's already achievable with the abovementioned config example.
3. Every feature is an extra maintenance burden. We'd like to spend time on features that bring more substantial improvements.

PR #2899, opened 2018-11-06
Fix typo

tdorsey:


packages/@vue/cli-service/generator/template/src/components/HelloWorld.vue
```diff
@@ -3,7 +3,7 @@
   <div class="hello">
     <h1>{{ msg }}</h1>
     <p>
-      For guide and recipes on how to configure / customize this project,<br>
+      For guides and recipes on how to configure / customize this project,<br>
```
LinusBorg: The docs site only has one "Guide" secion, so I think singular is fine. Would

> For a guide and recipes ...

be better?

tdorsey:
That would work as well. If that matches the site structure better, I’ll change it.

PR #2900, opened 2018-11-06
fix(cli-service): do not log absolute baseUrl in terminal, close #2877

### LINKED ISSUE #2877: baseUrl should handle full urls
Per the docs, "The base URL your application bundle will be deployed at. This is the equivalent of webpack's output.publicPath". In the [webpack docs](https://webpack.js.org/configuration/output/#output-publicpath) it's clear full urls are valid, one of the examples is 'https://cdn.example.com/assets/'.

From the [docs for `devServer.publicPath`](https://webpack.js.org/configuration/dev-server/#devserver-publicpath-): "It is also possible to use a full URL. This is necessary for Hot Module Replacement." The point about HMR is clear when you look at [issue 2525](https://github.com/vuejs/vue-cli/issues/2525#issuecomment-421573105), where the HMR is using the wrong URL for updates.

Setting `baseUrl` to a full URL results in a message from `vue-cli-service serve` such as:

>   App running at:
>  - Local:   https://localhost:8443/https://other-address.local/

It *does* fix the HMR issue, since it plumbs down correctly, but `vue-cli` isn't treating it as the full URL that it is.

If `baseUrl` is the equivalent of webpack's `output.publicPath` as the docs say, then it must correctly handle full URLs. Otherwise they are not equivalent.

N/A

jkzing:
#2877

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -42,6 +42,7 @@ module.exports = (api, options) => {
     const prepareProxy = require('../util/prepareProxy')
     const launchEditorMiddleware = require('launch-editor-middleware')
     const validateWebpackConfig = require('../util/validateWebpackConfig')
+    const isAbsoluteUrl = require('../util/isAbsoluteUrl')
 
     // resolve webpack config
     const webpackConfig = api.resolveWebpackConfig()
@@ -89,7 +90,7 @@ module.exports = (api, options) => {
       protocol,
       host,
       port,
-      options.baseUrl
+      isAbsoluteUrl(options.baseUrl) ? '/' : options.baseUrl
     )
 
     const proxySettings = prepareProxy(
--- a/packages/@vue/cli-service/lib/util/isAbsoluteUrl.js
+++ b/packages/@vue/cli-service/lib/util/isAbsoluteUrl.js
@@ -1,0 +1,4 @@
+module.exports = function isAbsoluteUrl (url) {
+  // A URL is considered absolute if it begins with "<scheme>://" or "//"
+  return /^([a-z][a-z\d\+\-\.]*:)?\/\//i.test(url)
+}


packages/@vue/cli-service/lib/commands/serve.js
```diff
@@ -89,7 +90,7 @@ module.exports = (api, options) => {
       protocol,
       host,
       port,
-      options.baseUrl
+      isAbsoluteUrl(options.baseUrl) ? '/' : options.baseUrl
```
LinusBorg: Won't that break for `baseUrl: http://localhost:8081/someBasePath/` ?

jkzing: Set baseUrl to `http://localhost:8081/someBasePath/` doesn't mean to start the dev-server on `localhost:8081`, the dev-server would still be started on `${host}:${port}`.

LinusBorg: Sure, but doesn't the console now say " go to host: port/" even though people would have to go to "localhost:port/baseurl"?

I'm honestly confused because paths and hosts in webpack can be confusing as hell to me

jkzing: Consider those two cases here:

1. Asset files (js/css/img) are served by `webpack-dev-server` (eg. localhost:8080), and user uses a reverse proxy to access the **html** (eg. localhost:3000/app/). In this case, `baseUrl` should be set to `http://localhost:8080/`, but user need to open `http://localhost:3000/app/` in browser.

2. Asset files (js/css/img) are served by `webpack-dev-server` (eg. localhost:8080), and user uses a reverse proxy to access the **asset files** (eg. localhost:3000/assets/). In this case, `baseUrl` should be set to `http://localhost:3000/assets/`, and user need to go to `http://localhsot:8080` in browser.

What I want to say is, we cannot infer which url should user open in browser if user set `baseUrl` to a absolute url. What we can do is tell user "we have started your dev-server on localhost:8080"... 

And, yes it is confusing. 😂

LinusBorg: Awesome, that was very helpful. Thanks.

PR #2932, opened 2018-11-11
chore(eslint-config-prettier): Upgrade prettier and extend configuration

michalsnik:
This PR updates the prettier and related packages, as well as extends eslint configuration, so it doesn't conflict with `eslint-plugin-vue` visual rules.


blowsie:
There appears to be a lot of confusion around the  different options for eslint + vue + prettier
Is there a working example config somewhere? Thanks

CreativeWarlock:
Even in 2022 this feels like a conundrum for beginners. I still haven't found any good example config anywhere that makes developers who started fresh with VS Code + Vue + eslint + prettier.

PR #2939, opened 2018-11-12
Update webpack-chain version

Djaler:
New release have very useful feature - https://github.com/neutrinojs/webpack-chain/pull/84


sodatea:
This would be a breaking change to our public API. So you need to monkey patching its internal implementation for compatibility with earlier versions. Otherwise we'll have to wait until v4 to update it.

Issue #2810, opened 2018-10-25
Docs overhaul


Akryum:
*Reference issue for docs refactor*

We are planning on overhauling the documentation for vue-cli soon. There will be two docs sites:
- User Manual (for the users of vue-cli and vue-cli ui)
- Plugin developer docs

Checklist:
- Improve docs (obviously)
- Move the plugin development docs to a separate website (for example vue-cli-dev.vuejs.org)
- Create a complete plugin dev guide
- Create a full API reference

Akryum:
cc @NataliaTepluhina

NataliaTepluhina:
Working on plugin dev guide + API reference

JonathanDn:
@Akryum  I would love to help if you could guide me to a specific area in the docs that need mending/improving

christiannwamba:
@Akryum same as @JonathanDn 

Would love to help if there's a path or somewhere you would love us to get started from.

Thanks

rkunev:
Bumping the Docs overhaul issue! 

I think there's an opportunity to improve the docs about [Build Targets](https://cli.vuejs.org/guide/build-targets.html#library) and the `--target lib`. Specifically, about `useBuiltIns` config for lib authors. [@vue/babel-preset-app docs](https://github.com/vuejs/vue-cli/tree/dev/packages/%40vue/babel-preset-app#usebuiltins) already explain it, but I think it's not easy to find this information and a [number](https://github.com/gilbitron/laravel-vue-pagination) [of](https://github.com/ajerez/vue-read-progress) [projects](https://github.com/baianat/hooper) [ship](https://github.com/nono1526/vue-material-year-calendar) dist files sometimes *twice* as large, just because of the polyfills. 

Here's a fresh [example](https://github.com/gilbitron/laravel-vue-pagination/pull/58) of how just using the `Number` prop validator like this:
```js
props: {
  limit: {
    type: Number, // <- es6.number.constructor polyfill is added
    default: 0
  }
}
```
increases the (unminified) bundle size from 17.33 KiB to 39.42 KiB.

Would you be interested in a Pull Request for the docs?
cc: @Akryum @NataliaTepluhina

nklayman:
With the release of vue cli v4, it would be really nice to have a guide on migrating plugins. It also appears as if the new migrate feature is undocumented.

brenzy:
It would be great if looking at older versions of the documentation was added here as well.  It is confusing for me what works with Vue 2, 3, and 4.

PR #2973, opened 2018-11-17
docs: Plugin development guide

NataliaTepluhina:
#2810
cc @chrisvfritz 

Plugin guide:
- [x]  Getting started
- [x]  Modify the webpack config
- [x]  Add a new cli-service command
- [x]  Generator
- [x]  Prompts
- [x]  Augment the task in the UI
- [x]  Display a configuration screen
- [x]  Show prompts in the UI
- [x]  Logo and discoverability
- [x] Publish plugin to npm

API:
- [x] Service Plugin API
- [x] Generator API


docs/dev-guide/plugin-dev.md
```diff
@@ -4,199 +4,331 @@ sidebarDepth: 3
 
 # Plugin Development Guide
 
-## Core Concepts
+## Getting started
 
-There are two major parts of the system:
+A CLI plugin is an npm package that can add additional features to a `@vue/cli` project. It should always contain a [Service Plugin](#service-plugin) as its main export, and can optionally contain a Generator, a Prompt File and a Vue UI integration.
```
chrisvfritz: > a `@vue/cli` project

What would you think about "a project using Vue CLI" instead? My thinking is to reference "Vue CLI" as an ecosystem of projects rather than `@vue/cli` as a package, since `@vue/cli` may not have actually been used to create the project in cases where someone upgraded or used a boilerplate - plus, `@vue/cli` will not be listed as a dependency in the project itself, which may cause confusion.

Specifying "a project using Vue CLI" instead of "a Vue CLI project" also differentiates between plugins that _tie into_ Vue CLI and applications/libraries that _make use of_ the Vue CLI ecosystem.

> It should always contain a [Service Plugin](#service-plugin) as its main export, and can optionally contain a Generator, a Prompt File and a Vue UI integration.

At a high level, what would you think about starting with user goals rather than implementation details? We're getting into the "how" of a plugin when we still haven't really answered the "why". As a user, I might be wondering, "How do I know I want to write a plugin? What are plugins for?" This paragraph contains _hints_ that I can receive options from the user, add/update project files, and add to the Vue CLI UI, but that's only the tip of the iceberg. Plus, it doesn't give me an idea of what a plugin _isn't_ appropriate for. For example, Vuetify has a plugin (because they want to make some changes to the Webpack config). Does this mean all component libraries need a plugin? Would I create a plugin to include lodash in a project?

> Generator, a Prompt File, and a Vue UI integration

Lists like this are a great place for links, for users who want to jump directly to one of these topics. If you're not sure what you want to yet or you don't want to break your flow while writing, I often just make the link `TODO` so I don't forget to come back to it (e.g. `[Generator](TODO)`).

docs/dev-guide/plugin-dev.md
```diff
@@ -4,199 +4,162 @@ sidebarDepth: 3
...
-[Service][service-class] is the class created when invoking `vue-cli-service <command> [...args]`. Responsible for managing the internal webpack configuration, and exposes commands for serving and building the project.
+:::tip
+Don't overuse vue-cli plugins! If you want just to include a certain dependency, e.g. [Lodash](https://lodash.com/) - it's easier to do it manually with npm than create a specific plugin only to do so.
+:::
 
-### CLI Plugin
+CLI Plugin should always contain a [Service Plugin](#service-plugin) as its main export, and can optionally contain a [Generator](#generator), a [Prompt File](#prompts) and a [Vue UI integration](TODO).
 
-A CLI plugin is an npm package that can add additional features to a `@vue/cli` project. It should always contain a [Service Plugin](#service-plugin) as its main export, and can optionally contain a [Generator](#generator) and a [Prompt File](#prompts-for-3rd-party-plugins).
+As a npm package, CLI plugin must have a `package.json` file. It's also recommended to have a plugin description in `README.md` to help others find your plugin on npm
```
Justineo:
```suggestion
As an npm package, CLI plugin must have a `package.json` file. It's also recommended to have a plugin description in `README.md` to help others find your plugin on npm.
```

docs/dev-guide/plugin-api.md
```diff
@@ -1 +1,97 @@
...
+  - `{boolean}`
+
+- **Usage**:
+Check if the project has a plugin with given id
+
+- **Example**
+
+```js
+if (api.hasPlugin('typescript')) {
+  console.log('This project is using TypeScript)
```
kazupon: not enough quote.

docs/dev-guide/plugin-api.md
```diff
@@ -1 +1,97 @@
...
+
+## getCwd
+
+- **Usage**:
+Returns a current working directory
+
+## resolve
+
+- **Arguments**
+  - `{string} _path` - relative path from project root
```
kazupon: Do `_` need ?

docs/dev-guide/plugin-api.md
NataliaTepluhina: Thanks, fixed!

docs/dev-guide/plugin-api.md
NataliaTepluhina: Yes, because the source code for Plugin API is:
```
const path = require('path')
...

resolve (_path) {
  return path.resolve(this.service.context, _path)
}
```
So we need `_` to avoid the conflict with the global `path`

NataliaTepluhina:
PR is ready for review.
@chrisvfritz @Akryum could you please have a look when you have a minute?

docs/dev-guide/plugin-api.md
Akryum: I think we can omit the `_` in the context of the docs.

docs/dev-guide/plugin-dev.md
```diff
@@ -296,41 +423,397 @@ module.exports = api => {
...
+//prompts.js
+
+module.exports = [
+  {
+    //basic propmpt properties
+    name: `addExampleRoutes`,
+    type: 'confirm',
+    message: 'Add example routes?',
+    default: false,
+    //UI-related propmpt properties
```
Akryum:
```suggestion
    //UI-related prompt properties
```

docs/dev-guide/plugin-dev.md
```diff
@@ -296,41 +423,397 @@ module.exports = api => {
...
+If you want, you can display [prompts](#prompts) in the Vue UI as well. If you will install your plugin through the UI, prompts will be shown on the plugin invocation step.
+
+To display a prompt in the UI you need to extend the [inquirer object](#prompts-for-3rd-party-plugins) with additional properties. They are optional and only used by the UI:
+
+```js
+//prompts.js
+
+module.exports = [
+  {
+    //basic propmpt properties
```
Akryum:
```suggestion
    // basic prompt properties
```

docs/dev-guide/plugin-dev.md
```diff
@@ -4,198 +4,91 @@ sidebarDepth: 3
...
 ```
 
-This is because the command's expected mode needs to be known before loading environment variables, which in turn needs to happen before loading user options / applying the plugins.
+## Generator
 
-#### Resolving Webpack Config in Plugins
+A Generator part of the CLI plugin is usually needed when you want to extend your package with new dependencies, create new files in your project or edit existing ones.
 
-A plugin can retrieve the resolved webpack config by calling `api.resolveWebpackConfig()`. Every call generates a fresh webpack config which can be further mutated as needed:
+Inside the CLI plugin the generator should be placed inside a `generator.js` or `generator/index.js` file. It will be invoked in two possible scenarios:
```
Akryum:
```suggestion
Inside the CLI plugin the generator should be placed in a `generator.js` or `generator/index.js` file. It will be invoked in two possible scenarios:
```

docs/dev-guide/plugin-dev.md
```diff
@@ -4,198 +4,91 @@ sidebarDepth: 3
...
-Alternatively, a plugin can also obtain a fresh [chainable config](https://github.com/mozilla-neutrino/webpack-chain) by calling `api.resolveChainableWebpackConfig()`:
-
-``` js
-api.registerCommand('my-build', args => {
-  const configA = api.resolveChainableWebpackConfig()
-  const configB = api.resolveChainableWebpackConfig()
+- During a project's initial creation, if the CLI plugin is installed as part of the project creation preset.
 
-  // chain-modify configA and configB for different purposes...
+- When the plugin is installed after project's creation and invoked individually via `vue invoke`.
```
Akryum:
```suggestion
- When the plugin is installed after project's creation and invoked individually via `vue add` or `vue invoke`.
```

docs/dev-guide/plugin-dev.md
```diff
@@ -4,198 +4,91 @@ sidebarDepth: 3
...
-  // copy and render all files in ./template with ejs
-  api.render('./template')
+After template is created, you should add `api.render` call to the `generator/index.js` file:
 
-  if (options.foo) {
-    // conditionally generate files
-  }
-}
+```js
+api.render('./template')
```
Akryum: Let's wrap this in the `module.exports = (api) => {}` function since there is no example of this before :)

docs/dev-guide/plugin-dev.md
```diff
@@ -235,7 +128,98 @@ export default {
...
+
+```js
+// generator/index.js
+
+module.exports = api => {
+  api.extendPackage({
+    dependencies: {
+      'vue-router-layout': '^0.1.2'
+    },
+  });
```
Akryum: Missing end `}`

docs/dev-guide/plugin-dev.md
```diff
@@ -235,7 +128,98 @@ export default {
...
+
+```js
+// generator/index.js
+
+module.exports = api => {
+  api.extendPackage({
+    scripts: {
+      greet: 'vue-cli-service greet'
+    }
+  });
```
Akryum: Missing end `}`

docs/dev-guide/plugin-dev.md
```diff
@@ -235,7 +128,98 @@ export default {
...
+
+module.exports = api => {
+  api.extendPackage({
+    dependencies: {
+      'vue-router-layout': '^0.1.2'
+    },
+  });
+```
+
+In the example above we added one dependency: `vue-router-layout`. During the plugin invocation this npm module will be installed and this dependency will be added to `package.json` file.
```
Akryum:
```suggestion
In the example above we added one dependency: `vue-router-layout`. During the plugin invocation this npm module will be installed and this dependency will be added to the user `package.json` file.
```

docs/dev-guide/plugin-dev.md
```diff
@@ -235,7 +128,98 @@ export default {
...
+
+With the same API method we can add new npm tasks to the project. To do so, we need to specify task name and a command that should be run in the `scripts` section:
```
Akryum:
```suggestion
With the same API method we can add new npm tasks to the project. To do so, we need to specify task name and a command that should be run in the `scripts` section of the user `package.json`:
```

docs/dev-guide/plugin-dev.md
```diff
@@ -235,7 +128,98 @@ export default {
...
+
+module.exports = api => {
+  api.extendPackage({
+    scripts: {
+      greet: 'vue-cli-service greet'
+    }
+  });
+```
+
+In the example above we're adding a new `greet` task to run a custom vue-cli service command created in [Service section](#add-a-new-cli-service-command)
```
Akryum:
```suggestion
In the example above we're adding a new `greet` task to run a custom vue-cli service command created in [Service section](#add-a-new-cli-service-command).
```

docs/dev-guide/plugin-dev.md
```diff
@@ -235,7 +128,98 @@ export default {
...
+
+Let's consider the case where we have created a `router.js` file via [templating](#creating-new-templates) and now we want to import this router to the main file. We will use two Generator API methods: `entryFile` will return the main file of the project (`main.js` or `main.ts`) and `injectImports` serves for adding new imports to this file:
+
+```js
+// generator/index.js
+
+api.injectImports(api.entryFile, `import router from './router'`)
+```
+
+Now, when we have a router imported, we can inject this router to the Vue instance in the main file. We will use `onCreateComplete` hook which is to be called when the files have been written to disk
```
Akryum:
```suggestion
Now, when we have a router imported, we can inject this router to the Vue instance in the main file. We will use `onCreateComplete` hook which is to be called when the files have been written to disk.
```

docs/dev-guide/plugin-dev.md
```diff
@@ -296,41 +423,397 @@ module.exports = api => {
...
+
+```js
+module.exports = api => {
+  // Use the API here...
+}
+```
+
+### Augment the task in the UI
+
+Vue CLI plugin allows you not only add new npm tasks to the project [via Generator](#extending-package) but also create a view fro them in Vue UI. It's useful when you want to run the the task right from the UI and see it's output there.
```
Akryum:
```suggestion
Vue CLI plugin allows you not only add new npm tasks to the project [via Generator](#extending-package) but also create a view for them in Vue UI. It's useful when you want to run the the task right from the UI and see its output there.
```

docs/dev-guide/plugin-dev.md
```diff
@@ -296,41 +423,397 @@ module.exports = api => {
...
+  // Shown below the name
+  description: 'This config defines the color of the greeting printed',
+  // "More info" link
+  link: 'https://cli.vuejs.org/dev-guide/ui-api.html#ui-api'
+})
+```
+
+#### Config logo
+
+You can also select an icon for your config. It can be either a [Material icon](https://material.io/tools/icons/?style=baseline) code or a custom image (see Public static files).
```
Akryum: Maybe add a link on `Public static files`?

docs/dev-guide/plugin-dev.md
```diff
@@ -296,41 +423,397 @@ module.exports = api => {
...
+module.exports = {
+  color: 'red',
+};
+```
+
+### Display prompts
+
+If you want, you can display [prompts](#prompts) in the Vue UI as well. If you will install your plugin through the UI, prompts will be shown on the plugin invocation step.
+
+To display a prompt in the UI you need to extend the [inquirer object](#prompts-for-3rd-party-plugins) with additional properties. They are optional and only used by the UI:
```
Akryum: Actually you don't have to. Prompts will still be displayed anyway. However you can "augment" them for the UI with the UI-specific properties.

docs/dev-guide/plugin-dev.md
```diff
@@ -296,41 +423,397 @@ module.exports = api => {
...
+
+As a result, you will have this screen on plugin invocation:
+
+![UI Prompts](/ui-prompts.png)
+
+### Logo
+
+You can put a `logo.png` file in the root directory of the folder that will be published on npm. It will be displayed in several places:
+ - When searching for a plugin to install
+ - In the installed plugin list
```
Akryum: - In the configuration
```suggestion
 - In the installed plugin list
 - In the configurations list (by default)
 - In the tasks list for augmented tasks (by default)
```

docs/dev-guide/plugin-dev.md
```diff
@@ -296,41 +423,397 @@ module.exports = api => {
...
+
+![Plugins](/plugins.png)
+
+The logo should be a square non-transparent image (ideally 84x84).
+
+### Discoverability
 
 For a CLI plugin to be usable by other developers, it must be published on npm following the name convention `vue-cli-plugin-<name>`. Following the name convention allows your plugin to be:
```
Akryum: Or `@scope/vue-cli-plugin-<name>`

docs/dev-guide/plugin-api.md
NataliaTepluhina: Ok, fixed

NataliaTepluhina:
@Akryum thank you for such a thorough review and suggestions! 🤗 
I've added requested fixes

NataliaTepluhina:
A little bump on this. Requested changes are done 🙌

Akryum:
I'll look into this very soon, I promise 🙏

docs/dev-guide/plugin-dev.md
```diff
@@ -235,37 +130,274 @@ export default {
...
+}
+```
+
+This is because the command's expected mode needs to be known before loading environment variables, which in turn needs to happen before loading user options / applying the plugins.
+
+## Prompts
+
+Prompts are required to handle user choices when creating a new project or adding a new plugin to the existing one. All prompts logic is stored inside the `prompts.js` file. The prompts are presented using [inquirer](https://github.com/SBoudrias/Inquirer.js) under the hood.
+
+### Prompts for Built-in Plugins
```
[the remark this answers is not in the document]
Akryum: Yes! 😸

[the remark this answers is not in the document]
NataliaTepluhina: Done!

PR #2981, opened 2018-11-20
feat!: manifest.json should be generated by cli-plugin-pwa

aweikalee:
English description provided by Google Translate.
中文在下方

### English description
`manifest.json` should be generated by `cli-plugin-pwa` and should not be copied directly from the `public` folder.

`manifest.json` is a separate file that cannot be modified by `vue.config.js`, and these values are very suitable for configuration with `vue.config.js`. For example: `name`, `short_name`, `start_url`, `icons[].src`.

Some attributes of `pwa` are the same or similar to those of `manifest.json`. For example, `pwa.themeColor` and `manifest.theme_color`, `pwa.name` and `manifest.name`.

### 中文说明
`manifest.json`应该由`cli-plugin-pwa`生成，不应该直接从`public`文件夹复制。

原本`manifest.json`是一个完全独立的文件，通过`vue.config.js`无法修改某些值，而这些值又非常合适使用`vue.config.js`配置。比如：`name`, `short_name`, `start_url`, `icons[].src`。

在原有的`pwa`的属性中，有些属性与`manifest.json`的属性相似或本应该是同一个。比如`pwa.themeColor`和`manifest.theme_color`, `pwa.name`和`manifest.name`。


jonaskuske:
We could also use [webpack-pwa-manifest](https://github.com/arthurbergmz/webpack-pwa-manifest) to generate the manifest instead of a simple `Object.assign()`.  

This would allow users to specify just one image/icon, which would then be down-scaled to all the required different icon sizes. That part is the biggest advantage of creating the manifest programmatically instead of by hand, im my opinion 🙂

aweikalee:
#3711 
Sorry, I have to create a new PR, because of my wrong operation.

aweikalee:
How can I throw a warn when build complete?
I can't find a way in other vue-cli code.
I try to use `setTimeout`, but can't pass the test.

sodatea:
Now that we are planning a v4 alpha release, The compatibility layer is no longer needed. I'll merge this PR directly. Thanks for your contribution!

Akryum:
I see the UI hasn't be updated to not use `manifest.json`

septatrix:
This does not seem to work with the default setup using vite any longer. Looking at the code this plugin seems to hook into webpack so is this not compatible with vite? Are there plans to add this support or is this basically in maintenance-only mode and I should look for alternative solutions for vite?

PR #3021, opened 2018-11-27
Make the demo msg property public

revolter:


sodatea:
This is intentional. All fields inside a component should be considered as private.

revolter:
But it doesn't work if it's private, because it's set from outside of the component:

https://github.com/vuejs/vue-cli/blob/7147aab85c8984858cddcc29fec3624b20bb3a8d/packages/%40vue/cli-service/generator/router/template/src/views/Home.vue#L5

sodatea:
"Being set from outside" is like class constructor parameters. This does not affect the fact that when accessing props inside a component (using `this.x` syntax), they should be considered local and private.

Issue #3031, opened 2018-11-28
log a warning when running `vue-cli-service serve [entry]` in a multi-page setup


chiaweilee:
### Version
3.2.0

### Reproduction link
[https://github.com/chiaweilee/vidget/blob/dev/scripts/serve.js](https://github.com/chiaweilee/vidget/blob/dev/scripts/serve.js)

### Node and OS info
Node 10.13.0 / npm 6.4.1 / MacOS Mojave 10.14.1

### Steps to reproduce
- add "vue-cli-service serve src/main.js" to "scripts"
- npm run "scripts"
- serve works
- open serve url in browser

### What is expected?
- serve website works good

### What is actually happening?
- serve website works nothing
- no js file have injected into DOM

---
1. Entry been set

check @vue/cli-service/lib/commands/serve.js:69

```JavaScript
// entry arg
    const entry = args._[0]
    if (entry) {
      webpackConfig.entry = {
        app: api.resolve(entry)
      }
    }
```

webpackConfig.entry is expected of resolved "src/main.js"

2. usage: 'vue-cli-service serve [options] [entry]'

```JavaScript
api.registerCommand('serve', {
    description: 'start development server',
    usage: 'vue-cli-service serve [options] [entry]',
    options: {
      '--open': `open browser on server start`,
      '--copy': `copy url to clipboard on server start`,
      '--mode': `specify env mode (default: development)`,
      '--host': `specify host (default: ${defaults.host})`,
      '--port': `specify port (default: ${defaults.port})`,
      '--https': `use https (default: ${defaults.https})`,
      '--public': `specify the public network URL for the HMR client`
    }
  }
```

3. fine when use vue.config.js

```JavaScript
// vue.config.js
module.exports = {
  pages: {
    index: {
      entry: 'src/main.js'
    }
  }
}
```

LinusBorg:
> It seems like vue-cli-service serve [options] [entry] can not work with vue.config.js's pages together?

Yes, naturally. having many pages means there is not one entry file, but many. So you can't provide *one* specific entry file when calling `serve`.

If you only have one page, don't use the `pages` option.

LinusBorg:
> dozen ways of improvement could be made.

Enlighten us, please

> should it console a warning?

Certainly a good idea.

chiaweilee:
```JavaScript
// lib/command/serve.js
const webpackConfig = api.resolveWebpackConfig() // line 48
// ...
// entry arg
[...]

PR #3041, opened 2018-11-29
fix: #3031

chiaweilee:
https://github.com/vuejs/vue-cli/issues/3031


LinusBorg:
Please explain the use case

Issue #2278, opened 2018-08-19
Add considerations about generated service-worker only enabled in the production environment


kwhsiung:
### What problem does this feature solve?
Let developers who want to build a PWA using `@vue/cli-plugin-pwa` know the reasons why [no-op service-worker](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-pwa/lib/noopServiceWorker.js) exist and why [checking](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-pwa/generator/template/src/registerServiceWorker.js#L5) `NODE_ENV === 'production'` in `registerServiceWorker.js`.

### What does the proposed API look like?
Add [considerations](https://github.com/facebook/create-react-app/blob/master/packages/react-scripts/template/README.md#offline-first-considerations) like create-react-app in `@vue/cli-plugin-pwa`'s `README.md`, or in the vue-cli's doc.

dadoeyad:
Yes please!
It took me a long while to get why my `service-worker.js` wasn't loading locally.

I'll add that it would be helpful if we can pass a list of which `NODE_ENV` to generate `service-worker.js` for. Instead of it being [hard coded for 'production' only](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-pwa/index.js#L22).
This hard coded value won't allow us to build for our staging environment :)

PR #3050, opened 2018-11-30
docs(pwa): explanation why SW is enabled only in production mode

NataliaTepluhina:
Requested in #2278


stoivo:
Maybe not so cool to comment on a closed MR, but are there any arguments why this can't be enabled in staging?

Issue #3009, opened 2018-11-26
Missing dependencies on default template install on windows


badpunman:
### Version
3.1.1

### Node and OS info
node 11.2, npm 6.4.1, windows 2012 

### Steps to reproduce
vue create project
go with default options: babel eslint

### What is expected?
No missing dependencies

### What is actually happening?
Missing dependencies
![image](https://user-images.githubusercontent.com/937289/49013261-cea99f00-f184-11e8-992f-2fa8e8e90623.png)

Requires different versions of ajv, is this an issue? (version 5 required by eslint, 6 by the rest.)
![image](https://user-images.githubusercontent.com/937289/49013423-5f807a80-f185-11e8-8042-b378914f853e.png)

philBrown:
I'm getting the same warnings on a fresh Linux install

Node v10.14.0, NPM 6.4.1 Vue CLI 3.2.1

```bash
npm WARN ajv-keywords@2.1.1 requires a peer of ajv@^5.0.0 but none is installed. You must install peer dependencies yourself.
```

Issue #2567, opened 2018-09-21
Maximum call stack size exceeded when creating new vue cli 3 project


ilrock:
### Version
3.0.3

### Node and OS info
Node 8.12.0 / npm 6.4.1 / OSx 10.13.4

### Steps to reproduce
vue create project_name

### What is expected?
Project being created

### What is actually happening?
The following error:

npm ERR! Maximum call stack size exceeded

npm ERR! A complete log of this run can be found in:
npm ERR!     /Users/ilrock/.npm/_logs/2018-09-21T12_44_37_935Z-debug.log
 ERROR  command failed: npm install --loglevel error

timfcsm:
i have same issue 
node 8.11.4 / npm 5.6.0 / vue  3.0.3 / win10

sodatea:
Can't reproduce. Could you provide a more detailed error log?

fer22f:
Here are two:

* [When trying to create a new project](https://github.com/vuejs/vue-cli/files/2492600/install.log)
* [Trying to add vuetify to a project created with yarn (can I use yarn here?)](https://github.com/vuejs/vue-cli/files/2492599/add_vuetify.log)

sodatea:
```
4780 http fetch GET 200 https://registry.npmjs.org/@webassemblyjs%2fwast-printer 2ms (from cache)
4781 silly fetchPackageMetaData error for @webassemblyjs/wast-printer@1.7.8 Unexpected end of JSON input while parsing near '..."}},"1.7.8":{"name":"'
```

Network issue or broken cache. Try to run `npm cache verify` and then run installation again.

fer22f:
Ok yea, broken cache. `npm cache clean --force` worked. And they advertise this cache as "self-healing", mhm

VadimIzmalkov:
Same issue

`npm cache clean --force` does not help

![default](https://user-images.githubusercontent.com/6684336/49364435-fd77c600-f6ec-11e8-93bc-21ee4be64818.png)

![default](https://user-images.githubusercontent.com/6684336/49364503-25672980-f6ed-11e8-9264-ae8c7bd3128a.png)

![default](https://user-images.githubusercontent.com/6684336/49364556-3fa10780-f6ed-11e8-8d61-a63bf10bca4b.png)

![default](https://user-images.githubusercontent.com/6684336/49364579-4b8cc980-f6ed-11e8-9475-df0d39b0ae9d.png)

Issue #1242, opened 2018-05-07
Cypress required when building production


cybercode:
### Version
3.0.0-beta.6

### Reproduction link
[https://gist.github.com/cybercode/49e97783e92b2bdd2710c1e5d9e55508](https://gist.github.com/cybercode/49e97783e92b2bdd2710c1e5d9e55508)

### Steps to reproduce
move build dependencies out of dev. Leave test dependencies in dev

### What is expected?
yarn build should build the dist

### What is actually happening?
error from vue-cli-service due to missing cypress library

---
Esp. deploying from a docker environment, the time required to download cypress, which is not needed for the build is problematic. You should be able to run `vue-cli-service build` without having to install the test tools.

AngeloCicero:
This would be really helpful.

yyx990803:
You probably should have some form of dependency caching based on lockfile hash...

cybercode:
That’s possible. I will test that premise

sodatea:
Moving build tools to `dependencies` does not make sense.

IMHO, the build result should be e2e tested before deploying, so cypress is required in that case.
Otherwise, you can move `@vue/cli-plugin-e2e-cypress` to `optionalDependencies` and use `yarn install --ignore-optional ` / `npm install --no-optional` in docker environments.

gjuchault:
@sodatea I've tried to move `@vue/cli-plugin-e2e-cypress` to `optionalDependencies` and I have the issue described by the author: vue-cli needs all plugins to be installed and thus I have a missing package error.

Could we reopen this issue ?

tommoyangn:
I'd like to +1 this issue. It doesn't make sense to me that we can't separate build tools from test tools. Yes, builds should be tested before deploying, but it doesn't make sense that it's tightly coupled in this way.

In our Jenkins setup where we build our app serially but then test different aspects concurrently this adds a significant amount of time to each build that isn't necessary, since not all of our tests require cypress (We're a PHP shop).

sodatea:
With the newest release, it is now possible to move `@vue/cli-plugin-e2e-cypress` to `optionalDependencies` without causing any errors. So I'll mark this issue as resolved.

bpkennedy:
This doesn't appear to be the case.  I am on @vue/cli-service ^3.3.0 (for all plugins, as well) and I have moved @vue/cli-plugin-e2e-cypress to `optionalDependencies` and I receive a failure in my gitlabCI build when running the default lint task:

```
[...]

PR #3068, opened 2018-12-03
fix(cli-plugin-eslint): make eslint deps optional

sodatea:
1. In @vue/cli-plugin-eslint 3.1.0+, eslint will be explicitly added to
the project's package.json, so it is **optional**.
2. `optionalDependencies` will still be installed by npm by default,
just no longer cause installation to fail when error happens.
3. ESLint v4 did cause weird errors sometimes, like `npm WARN
ajv-keywords@2.1.1 requires a peer of ajv@^5.0.0 but none is installed.`
Though this warning never caused any errors on my machine, I think it's
good to make the plugin fail-safe.

This change fixes https://github.com/vuejs/vue-cli/issues/3009


sodatea:
This should also fix https://github.com/vuejs/vue-cli/issues/2567#issuecomment-443641175

sodatea:
Note: This may break use cases like this if their projects were created by @vue/cli v3.0.x: https://github.com/vuejs/vue-cli/issues/1242#issuecomment-423000452

one-pieces:
This update will break the projects were created by @vue/cli v3.0.x

sodatea:
@one-pieces Did you run `npm install` with `--no-optional`? I didn't expect that use case to exist… 😂

PR #3097, opened 2018-12-07
fix: .vue file script lang is tsx but createSourceFile as JS. so  join ScriptKind.

lihuanji:
Support for TSX writing in.vue files and validation by tslint.


Yama-Tomo:
I have the same problem, hope to merge 👍

PR #3107, opened 2018-12-10
fix(cli-plugin-unit-jest): When using TS & Babel, make ts-jest use babelConfig 

### LINKED ISSUE #3100: Unit Test plugin and TypeScript w/ TSX not working
### Version
3.2.1

### Reproduction link
[https://github.com/lmiller1990/vue-cl-tsx-bug/pull/new/master](https://github.com/lmiller1990/vue-cl-tsx-bug/pull/new/master)

### Environment info
```
Mac OS
```

### Steps to reproduce
Clone above and `yarn test:unit`, or if you want your own: 

Scaffold app

Select unit tests, typescript and babel

Change HelloWorld.vue to HelloWorld.tsx and fill with the following:

```
export const MyComponent = {
  name: 'MyComponent',
  render() {
     return ( <div>asdf</div>)
  }
}
```

Update `example.spec.ts` to `import { MyComponent } from '@/components/MyComponent` and `shallowMount(MyComponent)...`

Run the tests with `yarn test:unit`. I get:

```
 Jest encountered an unexpected token

    This usually means that you are trying to import a file which Jest cannot parse, e.g. it's not plain JavaScript.

    By default, if Jest sees a Babel config, it will use that to transform your files, ignoring "node_modules".

    Here's what you can do:
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/en/configuration.html

    Details:

    /Users/lachlanmiller/javascript/vue/tsx/src/components/HelloWorld.tsx:6
            return (<div>asdf</div>);
                    ^

    SyntaxError: Unexpected token <

      1 | import { shallowMount } from '@vue/test-utils'
    > 2 | import { MyComponent } from '@/components/HelloWorld.tsx'
        | ^
      3 |
      4 | describe('HelloWorld.vue', () => {
      5 |   it('renders props.msg when passed', () => {

      at ScriptTransformer._transformAndBuildScript (node_modules/jest-runtime/build/script_transformer.js:403:17)
      at Object.<anonymous> (tests/unit/example.spec.ts:2:1)

```

### What is expected?
Runs the test fine

Does not appear to transpile the JSX/TSX

---
Maybe some work involving babel can fix this? I think all the official tools should work together. Maybe babel can do it.

LinusBorg:
tsconfig.json is set to preserve jsx:
```json
"jsx": "preserve"
``` 
Because Typescript can't compile JSX into Vue-style render functions, only React-style.

But that means:
* When using Typescript and Babel, Typescript will emit Javascript that still contains JSX. That's fine with`build` and `serve` if we also use babel to conmplie the JSX afterwards. But in Jest, we can only pick one transform, we can't chain them.
* When Only using Typescript, we can't use JSX at all.

This PR is meant to fix the first issue: ts-jest should use babel to transform result for correct JSX transforms by setting its `babelConfig` option.

close #3100

Sidenote: The second issue still exists, and likely isn't resolable. We should probably document it, though, and also mention in the prompt messages.

### DIFF
--- a/packages/@vue/cli-plugin-unit-jest/generator/index.js
+++ b/packages/@vue/cli-plugin-unit-jest/generator/index.js
@@ -79,6 +79,12 @@ const applyTS = module.exports.applyTS = (api, invoking) => {
       moduleFileExtensions: ['ts', 'tsx'],
       transform: {
         '^.+\\.tsx?$': 'ts-jest'
+      },
+      global: !api.hasPlugin('babel') ? undefined : {
+        'ts-jest': {
+          // we need babel to transpile JSX
+          babelConfig: true
+        }
       }
     },
     devDependencies: {
--- a/packages/@vue/cli-plugin-unit-jest/generator/index.js
+++ b/packages/@vue/cli-plugin-unit-jest/generator/index.js
@@ -79,12 +79,6 @@ const applyTS = module.exports.applyTS = (api, invoking) => {
       moduleFileExtensions: ['ts', 'tsx'],
       transform: {
         '^.+\\.tsx?$': 'ts-jest'
-      },
-      global: !api.hasPlugin('babel') ? undefined : {
-        'ts-jest': {
-          // we need babel to transpile JSX
-          babelConfig: true
-        }
       }
     },
     devDependencies: {
@@ -94,6 +88,14 @@ const applyTS = module.exports.applyTS = (api, invoking) => {
   })
   if (api.hasPlugin('babel')) {
     api.extendPackage({
+      jest: {
+        global: {
+          'ts-jest': {
+            // we need babel to transpile JSX
+            babelConfig: true
+          }
+        }
+      },
       devDependencies: {
         // this is for now necessary to force ts-jest and vue-jest to use babel 7
         'babel-core': '7.0.0-bridge.0'


packages/@vue/cli-plugin-unit-jest/generator/index.js
```diff
@@ -79,6 +79,12 @@ const applyTS = module.exports.applyTS = (api, invoking) => {
       moduleFileExtensions: ['ts', 'tsx'],
       transform: {
         '^.+\\.tsx?$': 'ts-jest'
+      },
+      global: !api.hasPlugin('babel') ? undefined : {
+        'ts-jest': {
+          // we need babel to transpile JSX
+          babelConfig: true
```
lmiller1990: According to `ts-jest` this option will

>  ts-jest will try to find a .babelrc, .babelrc.js file or a babel section in the package.json ...

But vue-cli generates a `babel.config.js`, a relatively new way to configure babel.

`ts-jest` lets you specify the path like so:

```jsglobals: {
  'ts-jest': {
    babelConfig: 'babel.config.js'
  }
}
```

Maybe that would work.

LinusBorg: Thanks, but `true` actually worked like a charm in a vue-cli project I tried it with. So I don'T think we need to specifiy a path./filename

lmiller1990: huh, awesome. I wonder why that worked. Good to hear!

LinusBorg:
FYI I have trouble testing this locally when calling `vue add @vue/typescript`, but not directly with the change - yarn simply hangs at 2289/2290.

Will try a few things

LinusBorg:
Tested locally, works fine. 

Not sure I should add a test, the jest generator tests that currently exist don't cover much details either.

@sodatea ?

sodatea:
@mortyccp Ahhhhh you are right, thanks! We should definitely have added a test for this.

PR #3111, opened 2018-12-10
 Fix minor typo in Deployment page of Guide

da70:
Was #3109.  Closed that and made a new PR using different branch. 

- Change "making" to "make"


da70:
Glad I could be of some small assistance.  The documentation is very helpful.  Thank you for maintaining it so well.

PR #3130, opened 2018-12-13
fix(cli-service): Fix copy plugin's ignore pattern (fix #3119)

### LINKED ISSUE #3119: Multiple index.html files in public folder not being published in production
### Version
3.2.1

### Environment info
```
System:
  OS: Windows 7
  CPU: (8) x64 Intel(R) Core(TM) i7-4810MQ CPU @ 2.80GHz
Binaries:
  Node: Not Found
  Yarn: 1.3.2 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
  npm: 5.6.0 - C:\Program Files\nodejs\npm.CMD
npmGlobalPackages:
  @vue/cli: Not Found
```

### Steps to reproduce
- Create an empty Vue project
- Create the following folder structure:
 - Root/public
 - Root/public/index.html
 - Root/public/test/index.html
- Run the command npm run build

### What is expected?
The dist folder to contain public/index.html and the public/test/index.html

The dist folder only contains the public/index.html and not the one from the test directory.

---
Related vue topic:
https://forum.vuejs.org/t/vue-cli-production-build-issues-with-multiple-index-html-files-in-public-folder/51110

LinusBorg:
don't skip files named `index.html` `/public` when nested in a subfolder. We only need to ignore the index.html directly under `/public/index.html`, which is used by html-webpack-plugin as a template.

close #3119

### DIFF
--- a/packages/@vue/cli-service/lib/config/app.js
+++ b/packages/@vue/cli-service/lib/config/app.js
@@ -140,7 +140,7 @@ module.exports = (api, options) => {
     const multiPageConfig = options.pages
     const htmlPath = api.resolve('public/index.html')
     const defaultHtmlPath = path.resolve(__dirname, 'index-default.html')
-    const publicCopyIgnore = ['index.html', '.DS_Store']
+    const publicCopyIgnore = ['/index.html', '.DS_Store']
 
     if (!multiPageConfig) {
       // default, single page setup.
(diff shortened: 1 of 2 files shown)


packages/@vue/cli-service/lib/config/app.js
```diff
@@ -140,7 +140,7 @@ module.exports = (api, options) => {
     const multiPageConfig = options.pages
     const htmlPath = api.resolve('public/index.html')
     const defaultHtmlPath = path.resolve(__dirname, 'index-default.html')
-    const publicCopyIgnore = ['index.html', '.DS_Store']
+    const publicCopyIgnore = ['/index.html', '.DS_Store']
```
sodatea: Actually we can just delete the `'/index.html'` pattern…
It doesn't match any files as I tested out. But it doesn't break anything because that html-webpack-plugin's output overwrites copy-webpack-plugin's anyway…

LinusBorg: Oh, thought copy would run after the html plugin

PR #3131, opened 2018-12-13
fix(unit-mocha): set `url` option for jsdom-global

michaelkhabarov:
Set `url: 'http://localhost'` option to jsdom-global. This fix `SecurityError: localStorage is not available for opaque origins` in mocha tests caused of jsdom (jsdom/jsdom#2304).


sodatea:
Thanks!

adambiggs:
Is there any workaround for this problem in the meantime when using mocha?

michaelkhabarov:
You can create a setup file and require it in test:unit script in `package.json`

mocha.setup.js:
```javascript
require('jsdom-global')(undefined, { pretendToBeVisual: true, url: 'http://localhost' })
```
package.json:
```javascript
"test:unit": "vue-cli-service test:unit --require ./mocha.setup.js"
```

PR #3132, opened 2018-12-14
TypeScript plugin: Fix typo in lint command registration

philipchungtech:
When you run `vue-cli-service help`, there is no description for the `lint` command. I discovered that this was because the command registration code had a typo.


sodatea:
Thanks!

Issue #2506, opened 2018-09-13
Can i  place all .env files in a directory?


Shiroyan:
### What problem does this feature solve?
I don't like placing all .env files in porject root , it makes my project structure look lengthy,  can i make a directory to place them in it ?
![image](https://user-images.githubusercontent.com/26413534/45466590-4868e980-b74d-11e8-9efb-e336db0c1c58.png)

### What does the proposed API look like?
I can config it in vue.config.js

jtommy:
@LinusBorg I'm working on it and I thought to add a new option `environmentsDir` into vue.config.js. 
Is it a good solution ? 
Because I have just seen that it's possible to change the vue.config.js location using an env variable  (`process.env.VUE_CLI_SERVICE_CONFIG_PATH`) and I need to load environment files after vue.config.js.

LinusBorg:
might be a nice idea, yes.

yyx990803:
Our overall guideline is - all CLI projects should have consistent file structures as much as possible. The CLI is opinionated by nature. This is why we do not provide options for configuring source directory or public directory names. If we add an option for this we essentially need to add options for renaming every possible config file and directory, and that would

1. make it more work to maintain all these options
2. increase possibility of edge cases due to different directory names
3. make Vue CLI projects all look different from one another.

chenxeed:
Hello giving second thought of this.

As in my project structure, it's a monorepo that contain backend API and frontend website together.

```
my-project/
--server/ (node app)
--client/ (vue-cli)
```

Both may share the same variable, thus it's best if they can refer to the same `.env`. So my thought is to put the `.env` file in the root of the project folder:

```
my-project/
-- server/ (node app)
-- client/ (vue-cli)
-- .env
```

So, if it's related to the discussion here, can the `vue-cli` compiler in folder `client` still able to retrieve the `env` value from the root directory?

farcasmihai91:
I have the exact same issue as @chenxeed 
As a workaround I installed [node-env-run](https://github.com/dkundel/node-env-run) as a dev dependency, and now my dev script looks like:

`"dev": "nodenv --env ../../.env --exec 'vue-cli-service serve'",`

Not ideal but works for now.

c2huc2hu:
Another workaround that supports variable expansion: 

Add to the beginning of `vue.config.js`: 

```
const dotenv = require('dotenv')
const dotenvExpand = require('dotenv-expand')
var myEnv = dotenv.config({path: '../.env'})
[...]

PR #3135, opened 2018-12-14
feat(cli-service): add envs dir config

jtommy:
#2506

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -80,6 +80,13 @@ module.exports = {
   `assetsDir` is ignored when overwriting the filename or chunkFilename from the generated assets.
   :::
 
+### environmentsDir
+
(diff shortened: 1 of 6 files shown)


LinusBorg:
This change means that env Bars See no er available in the vue.config.js file directly...

LinusBorg:
> To be honest I didn't find another config file, except vue.config.js, where to add this option.

Well, vue.config.js would be the right place in theory, and there's not other config file to use.

But since this change would mean you can't access process.en.VUE_APP_WHATEVER from your env files anymore (except in functions like `chainWeback` which are executed later), this would constitutes a breaking change and we can't add it until we do 4.0, strictly speaking.

We can use #2506 to discuss other strategies, but right now I don't have any in mind.

PR #3143, opened 2018-12-17
feat: deprecate confusing `baseUrl` option, use `publicPath` instead.

sodatea:
In most cases we use `baseUrl` as a way to modify `publicPath` safely.
The name `baseUrl` did mislead some people: in some rare cases we want to modify the `publicPath` of static assets while not influencing the page url (https://github.com/vuejs/vue-cli/issues/2788). In that case this option does not indicate the base location of the web page but the application bundle, `publicPath` would be a better and more common name for it.

(One exception is `process.env.BASE_URL`, I'm still wondering if we should rename that env variable too).


IndexXuan:
only change in vue.config.js and the same with baseUrl ? 
why not use process.env.PUBLIC_URL like create-react-app ?

sodatea:
@IndexXuan `PUBLIC_URL` is not the same as `publicPath` in webpack config.

IndexXuan:
@sodatea replace baseUrl as publicPath do the same with before ?
what different with publicPath and PUBLIC_URL in production

sodatea:
@IndexXuan 
`publicPath` better reflects the actual use of this option.

`PUBLIC_URL` is like `BASE_URL` in this PR. It's meant to be read-only and exposed by Vue CLI rather than the user.
I was thinking about changing the name too.

IndexXuan:
@sodatea Our product is a service that user upload source code and we build them and publish in CDN and html serve in somewhere. Our CI export PUBLIC_URL so that user not care about that, just use it in build scripts(Vue CLI 3 is vue.config.js#baseUrl before) and CRA user use PUBLIC_URL too.

Change baseUrl to publicPath is safely for us ?

IndexXuan:
@sodatea this name change break some Vue CLI 3 plugin which use projectOptions#baseUrl, like:
https://github.com/longshihui/vue-cli-plugin-navigator/blob/master/service/index.js#L68

sodatea:
@IndexXuan Got it. Will fix it in the next patch.

PR #3151, opened 2018-12-18
fix(ie compat): include polyfill for currentScript directly

ameinhardt:
it would be nice if vue-cli produces libs, that can be used in IE directly.

### DIFF
--- a/packages/@vue/cli-service/lib/commands/build/setPublicPath.js
+++ b/packages/@vue/cli-service/lib/commands/build/setPublicPath.js
@@ -1,9 +1,25 @@
 // This file is imported into lib/wc client bundles.
 
 if (typeof window !== 'undefined') {
+  var currentScript = window.document.currentScript;
+  if(!currentScript) {
+    // IE compat
+    try {
+      throw new Error();
+    } catch(err) {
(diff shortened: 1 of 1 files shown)


ameinhardt:
workaround to overwrite the existing file in vue.config.js:
```
config
      .plugin('fix-vue-cli-ie')
      .use(NormalModuleReplacementPlugin)
      .init(Plugin => new Plugin(/setPublicPath\.js/, __dirname + '/src/betterSetPublicPath.js'));
```

yyx990803:
I think what we can do is:

- detect whether user intends to support IE by analyzing `browserslist` in `package.json`
- If need IE support, inject an env variable `NEED_CURRENTSCRIPT_POLYFILL` and wrap the polyfill in a conditional block.
- This ensures the polyfill is dropped if user doesn't need IE support
- For the rare use case where user is producing multiple independent webpack bundles (thus duplicating the polyfill a lot), they can manually set `NEED_CURRENTSCRIPT_POLYFILL` to `false` in `.env` files and manage the polyfill themselves.

sodatea:
Closed in favor of https://github.com/vuejs/vue-cli/pull/3760

PR #3166, opened 2018-12-21
Validates multiple arguments for vue create command

jamesgeorge007:
As per now `vue create <appname>` doesn't produce any warnings if the user fires in `vue create <appname> <randomname>`
> The user can specify `n` number of arguments which doesn't makes sense.

### DIFF
--- a/packages/@vue/cli/bin/vue.js
+++ b/packages/@vue/cli/bin/vue.js
@@ -57,6 +57,9 @@ program
   .option('-b, --bare', 'Scaffold project without beginner instructions')
   .action((name, cmd) => {
     const options = cleanArgs(cmd)
+    if (program.args.length > 1){
+      console.log(chalk.yellow('\n Info: You provided more than one argument. The first one will be used as the app\'s name, the rest are ignored.'))
+    }  
     // --git makes commander to default git to true
     if (process.argv.includes('-g') || process.argv.includes('--git')) {
       options.forceGit = true


packages/@vue/cli/bin/vue.js
```diff
@@ -57,6 +57,10 @@ program
   .option('-b, --bare', 'Scaffold project without beginner instructions')
```
kazupon: please adjust the indent (L57-L69) ...

packages/@vue/cli/bin/vue.js
```diff
@@ -57,6 +57,10 @@ program
...
   .action((name, cmd) => {
     const options = cleanArgs(cmd)
+    if(program.args.length > 1){
+      program.outputHelp()
+      console.log(chalk.red.bold('\n Kindly provide only one argument as the appname'))
```
LinusBorg: We don't throw an error or anything (rightfully so), we continue processing, so asking the user to only provide one argument when he can't (we continue processing, right?), I would rather phrase it like this:

> Info: You provided more than one argument. The first one will be used as the app's name, the rest are ignored.

packages/@vue/cli/bin/vue.js
jamesgeorge007: Modified the if statement (introduced space in between the `if` keyword and parentheses)

packages/@vue/cli/bin/vue.js
jamesgeorge007: Rephrased the comment as you said and removed logging help.

jamesgeorge007:
@kazupon @LinusBorg What do you think?

PR #3181, opened 2018-12-25
feat(cli-service): add history api fallback for multi-page mode

jkzing:
With `multi-page` configuration like: 

```js
module.exports = {
  pages: {
    main: {
      entry: 'src/main.js',
      template: 'public/index.html',
      filename: 'index.html'
    },
    preview: {
      entry: 'src/preview.js',
      template: 'public/index.html',
      filename: 'preview.html'
    }
  }
}
```

HTML for page `preview` can only be accessed by `localhost:8080/preview.html`.

This plays well with vue-router + `hash` mode. But with `history` mode, `preview.html` would not be accessible. 

eg. accessing `localhost:8080/preview.html/about` will result in `index.html` because `/preview.html/about` is a not found request and being fallback.

This PR adds support for multi-page fallback, it fallback `${baseUrl}/${name of page}` to the correct html file (filename specified or `${name}.html`). 

In above case, the two pages could be accessed by `localhost:8080/main` and `localhost:8080/preview`.

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -303,6 +303,9 @@ function genHistoryApiFallbackRewrites (baseUrl, pages = {}) {
   const path = require('path')
   const multiPageRewrites = Object
     .keys(pages)
+    // sort by length in reversed order to avoid overrides
+    // eg. 'page11' should appear in front of 'page1'
+    .sort((a, b) => b.length - a.length)
     .map(name => ({
       from: new RegExp(`^/${name}`),
       to: path.posix.join(baseUrl, pages[name].filename || `${name}.html`)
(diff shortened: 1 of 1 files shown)


packages/@vue/cli-service/lib/commands/serve.js
```diff
@@ -302,6 +299,20 @@ function checkInContainer () {
   }
 }
 
+function genHistoryApiFallbackRewrites (baseUrl, pages = {}) {
+  const path = require('path')
+  const multiPageRewrites = Object
+    .keys(pages)
+    .map(name => ({
+      from: new RegExp(`^/${name}`),
```
sodatea: Will fail if several page names share the same prefix, e.g. `page1`, `page10` and `page11`.

jkzing: Yes you are right. 

I found it's hardly to avoid this kind of situation. We can't simply add a trailing slash in the regex. In that way user have to add a trailing slash in url to access specified page, eg `localhost:8080/page1/`.

Any thoughts?

sodatea: What about `Object.keys(pages).sort().reverse().map(...)`?

jkzing: Nice solution, sort page names by length should be enough to avoid this issue. Thanks!

PR #3191, opened 2018-12-26
feat!: upgrade webpack-chain to v5 and use named optimization.minimizer() API

mc-zone:
webpack-chain v5([Changelog](https://github.com/neutrinojs/webpack-chain/blob/master/CHANGELOG.md#v500)) has upgraded `optimization.minimizer` to ChainMap that we can use like following:
```javascript
webpackConfig.optimization.minimizer('terser')
  .use(TerserPlugin, [terserOptions(options)])
```
So users can modify the minimizer options more elegant:
```javascript
chainWebpack: chain => {
  chain.optimization.minimizer('terser')
    .tap(options => mergeCustomOptions(options))
}
```

Or just delete it conveniently:
```javascript
chainWebpack: chain => {
  chain.optimization.minimizers.delete('terser');
}
```

### DIFF
--- a/packages/@vue/cli-service/lib/config/prod.js
+++ b/packages/@vue/cli-service/lib/config/prod.js
@@ -28,9 +28,8 @@ module.exports = (api, options) => {
       } else {
         const TerserPlugin = require('terser-webpack-plugin')
         const terserOptions = require('./terserOptions')
-        webpackConfig.optimization.minimizer([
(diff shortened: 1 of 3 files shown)


sodatea:
Landed in v4.0.0-alpha.0
Thanks for this PR, and sorry for not using your commit, as I found it to tedious to fix the merge conflicts…

PR #3195, opened 2018-12-27
validates multiple arguments for vue create command

jamesgeorge007:
As per now `vue create <appname>` doesn't produce any warnings if the user fires in `vue create <appname> <randomname>`
> The user can specify `n` number of arguments which doesn't makes sense.


packages/@vue/cli/bin/vue.js
```diff
@@ -57,6 +57,9 @@ program
   .option('-b, --bare', 'Scaffold project without beginner instructions')
   .action((name, cmd) => {
     const options = cleanArgs(cmd)
+    if (program.args.length > 1){
```
sodatea: There're quite a few false positives with this conditional expression actually.
It might be helpful to write a test case for it so that the problem could be identified earlier.

jamesgeorge007: @sodatea So you are expecting me to write separate test case right within the `scripts/test.js` file?

sodatea: @jamesgeorge007 
Ah, I forgot we didn't have a test setup for cli utilities…
So you can skip it if you feel it's too tedious for such a small patch.

But the false positives have to be fixed. Your code would log warning even for commands like`vue create a` but not for `vue create -d a b`

jamesgeorge007:
@sodatea Also, the warning goes away in a flash since `clearConsole()` is invoked within the respective method triggered for the `create` command. What I think is either shifting this patch to that particular file or instead using  `setTimeout` to delay invoking the respective method.
What do you think?

jamesgeorge007:
@sodatea Is this fine now?

jamesgeorge007:
@sodatea Is this fine now?

jamesgeorge007:
@sodatea Is this fine now?

jamesgeorge007:
@sodatea I decided to go the manual way as commander's API didn't worked well in all cases.
This may appear weird to you, but still serves the purpose.

> What left over now is with the `--git` option as it has got an optional argument

jamesgeorge007:
@sodatea I decided to go the manual way as `commander's API `didn't worked well in all cases.
This may appear weird to you, but still serves the purpose.

> What left over now is the `--git` option as it has got an optional argument

jamesgeorge007:
@sodatea I decided to go the manual way as `commander's API `didn't worked well in all cases.
This may appear weird to you, but still serves the purpose.

jamesgeorge007:
@sodatea I decided to go the manual way as `commander's API `didn't worked well in all cases.
This may appear weird to you, but still serves the purpose.

jamesgeorge007:
@sodatea I decided to go the manual way as `commander's API `didn't worked well in all cases.
This may appear weird to you, but still serves the purpose.

jamesgeorge007:
@Kazupon'@LinusBorg What do you think?

jamesgeorge007:
@Kazupon @LinusBorg What do you think?

packages/@vue/cli/bin/vue.js
```diff
@@ -65,6 +67,24 @@ program
   .option('-b, --bare', 'Scaffold project without beginner instructions')
   .action((name, cmd) => {
     const options = cleanArgs(cmd)
+
```
jkzing: Actuall we have [minimist](https://github.com/substack/minimist) in dependency, why not just use `minimist` to parse args ? Something like:

```suggestion
if (minimist(process.argv.slice(3))._.length > 1) {
  // do warning here
}
```

jamesgeorge007:
@jkzing What do you think now?

jamesgeorge007:
@sodatea I decided to go the manual way as `commander's API `didn't worked well in all cases.
This may appear weird to you, but still serves the purpose.

jamesgeorge007:
@jkzing What do you think now?

jkzing:
@jamesgeorge007 I didn't see any changes since my comment.

jamesgeorge007:
@jkzing Sorry, it was not updated for some reason. Kindly have a look at it again :+1:

jamesgeorge007:
@jkzing Hope it's fine now :+1:

jamesgeorge007:
@sodatea Haven't heard from you for a while?

PR #3202, opened 2018-12-27
chore: add deprecation warning for Node.js 9.x

sodatea:


packages/@vue/cli/bin/vue.js
```diff
@@ -19,6 +19,14 @@ function checkNodeVersion (wanted, id) {
 
 checkNodeVersion(requiredVersion, 'vue-cli')
 
+if (semver.satisfies(process.version, '9.x')) {
+  console.log(chalk.red(
+    `You are using Node ${process.version}.\n` +
+    `Node.js 9.x has already reached end-of-life and will not be supported in future major releases.\n` +
+    `It's strongly recommended to upgrade your Node.js version as soon as possible.`
```
yyx990803: Not necessarily an upgrade - maybe encourage users to use an active LTS version like 8.x or 10.x.

PR #3229, opened 2019-01-02
fix(cli-plugin-babel): transpileDependencies should only match packages inside `node_modules`, close #3057

### LINKED ISSUE #3057: Clarification around transpileDependencies
`transpileDependencies` searches for the provided terms in the **full path**, meaning that if we have something like `transpileDependencies: ['vuetify']` (see https://github.com/vuetifyjs/vue-cli-plugin-vuetify/issues/55) then any `vuetify` in the path will cause Babel to translate the files.

For example, if my path is `/home/aymkdn/projects/vuetify/my-app/`, EVERYTHING is compiled, causing multiple errors. Even if the path is `/home/aymkdn/projects/my-app/` Babel will translate all node modules with "vuetify" in their name (like `vue-cli-plugin-vuetify` or `vuetify-loader`)

The expected behavior when reading the documentation is to only compile the module called `"vuetify"`

I see three options here : 
1. Clarify the [documentation](https://cli.vuejs.org/config/#transpiledependencies) to explain that this option will search for the terms in the whole path
2. Change `transpileDependencies` to only look at node module names (for example `transpileDependencies: ['vuetify']` will be equivalent to `/home/aymkdn/projects/vuetify/my-app/node_modules/vuetify/`)
3. Add a new parameter, called `transpileModules` that will do the same as option 2 above

jkzing:
#3057


packages/@vue/cli-plugin-babel/__tests__/transpileDependencies.spec.js
```diff
@@ -70,10 +70,23 @@ test('dep from node_modules should not been transpiled', async () => {
 test('dep from node_modules should been transpiled', async () => {
   await project.write(
     'vue.config.js',
-    `module.exports = { transpileDependencies: ['external-dep'] }`
+    `module.exports = { transpileDependencies: ['external-dep', '@scope/external-dep'] }`
```
jkzing: Why do we want to implicit include scoped package with the same name? This behavior isn't documented...

sodatea: I think it was unintended… Thanks for pointing out.

PR #3256, opened 2019-01-06
Add Tipp for group projects

pinguin999:
Source: https://youtu.be/TWqh9MtT4Bg?t=129


sodatea:
IMHO if we want to add tips on urls, it's better to follow step 1 of the preceding section on GitHub Pages as their strategies are quite similar, and we do need to remind users to change the `publicPath` accordingly.

PR #3294, opened 2019-01-14
report command added to generated readme.md

atilkan:
I tried to implement this myself but it is already implemented. This information might help somebody.


sodatea:
`report` is not an npm script, it's a boolean argument of `vue-cli-service build`.

PR #3306, opened 2019-01-15
fix:  failed to load config "plugin:vue/no-layout-rules" to extend from.

liaoyinglong:
[to fix this issues](https://github.com/prettier/eslint-config-prettier/issues/70)


sodatea:
Sorry for the very late reply.

I didn't merge this PR at the beginning because upgrading the eslint plugin was a breaking change. The issue is fixed now after the release of Vue CLI 4.

PR #3320, opened 2019-01-17
docs(proxy): add warning when proxy set as string (fix #3308)

### LINKED ISSUE #3308: [docs] Setting devServer.proxy to a string value can be confusing
### What problem does this docs change solve?
When setting `devServer.proxy` to a string value, proxying only happens for XHR requests, which can be confusing:
- configure `devServer.proxy` to `'http://localhost:3000'`;
- open `http://localhost:8080/api/users` from the address bar in the browser;
  - **expectation**: the request is proxyed to `http://localhost:3000/api/users`, the browser shows the API response;
  - **reality**: the browser shows the app (it loads `index.html`, because of `webpack-dev-server`'s `historyApiFallback`);
- the same request works fine (gets proxyed and returns the API response) if it's done using `XMLHttpRequest` (`axios`) or `fetch`.

If setting `devServer.proxy` explicitly to `{ '^/api': { target: 'http://localhost:3000' } }`, the same URL - `http://localhost:8080/api/users` - will work as expected both when opened from the address bar and when loading it from an XHR request.

### What should the docs change look like?
Make this distinction clear in the docs, as users might set `devServer.proxy` to a string, test that it works by loading an API URL from the browser's address bar, see the app instead of the API response, and conclude that it doesn't work.

lbogdan:
Fixes #3308 , see that for details.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -328,6 +328,10 @@ Deprecated since Vue CLI 3.3, please use [`publicPath`](#publicPath) instead.
 
   This will tell the dev server to proxy any unknown requests (requests that did not match a static file) to `http://localhost:4000`.
 
+  ::: warning
+  When `devServer.proxy` is set to a string, only XHR requests will be proxied. If you want to test an API URL, don't open it in the browser, use an API tool like Postman instead.
+  :::
+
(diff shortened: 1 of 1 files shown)


lbogdan:
Comments:
- better wording?
- should we recommend Postman? I added it because I felt "API tool" alone might not be very clear.
- should we instead recommend to make a request using axios / fetch?

NataliaTepluhina:
@lbogdan (better late than never): I think the wording is fine and I am also fine with Postman :)

PR #3323, opened 2019-01-17
Add docker based nginx deployment guide to docs

cregetycreg:
This guide shows how to build a docker image to run a vue-cli based application with nginx. It has an example nginx config that does the appropriate pushState() based routing fallbacks for 404's.

### DIFF
--- a/docs/guide/deployment.md
+++ b/docs/guide/deployment.md
@@ -319,3 +319,90 @@ Verify your project is successfully published by Surge by visiting `myawesomepro
 
     cd -
     ```
+
+
+### Docker (Nginx)
+
+Deploying your application using nginx inside of docker can be done easily with a minor modification of the nginx configuration.
+
+1. Install [docker](https://www.docker.com/get-started)
+
+2. Create a `Dockerfile` file in the root of your project.
+
+  ```Dockerfile
+  FROM node:10
+  COPY ./ /app
+  WORKDIR /app
+  RUN npm install && npm run build
+
+  FROM nginx
+  RUN mkdir /app
+  COPY --from=0 /app/dist /app
+  COPY nginx.conf /etc/nginx/nginx.conf
+  ```
+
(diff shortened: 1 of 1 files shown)


docs/guide/deployment.md
```diff
@@ -319,3 +319,90 @@ Verify your project is successfully published by Surge by visiting `myawesomepro
 
     cd -
     ```
+
+
+### Docker (Nginx)
+
+Deploying your application using nginx inside of docker can be done easily with a minor modification of the nginx configuration.
```
znck: It's unclear when you say _minor modification of the nginx configuration_, it would be nice if explicitly specify what we mean by _the nginx configuration_. Also, we can drop **easily**.

docs/guide/deployment.md
```diff
@@ -319,3 +319,90 @@ Verify your project is successfully published by Surge by visiting `myawesomepro
...
+    Setting up the `.dockerignore` file correctly prevents `node_modules` and any intermediate build artifacts from being copied to the image which can cause issues during building.
+
+    ```gitignore
+    **/node_modules
+    **/dist
+    ```
+
+4. Create a `nginx.conf` file in the root of your project
+
+    This is a simple `nginx` configuration that falls back to the root `index.html` page for `404's`. This is necessary when using `pushState()` based routing.
```
znck: It can be reworded to:

Following `nginx` configuration serves root `index.html` for `page not found` errors.

znck:
The process looks good to me. 👍

docs/guide/deployment.md
cregetycreg: I removed some of the confusing language.

docs/guide/deployment.md
cregetycreg: Done

SimonChaumet:
@cregetycreg I don't understand the part with the 50x errors, if we don't have a server how is it possible to have a 50x error ?

PR #3329, opened 2019-01-19
created toggle with process.env.VUE_APP_ALLOW_LOCAL_SW to allow local…

bobbybol:
When building PWA's with complex functionalities and interactions, it may be desirable to have a functioning service worker in development mode. This branch has a toggle in the shape of VUE_APP_ALLOW_LOCAL_SW that, when set to true, allows the developer to have the service worker under development installed on the client while running the development server.


bobbybol:
Never mind this - variables on process.env are strings, so boolean check doesn't work. Any advice for what would be a good globally injected variable to use when implementing such a toggle?

bobbybol:
I changed the toggle to specifically check for the string 'true'. This works of course, but would be glad to change it if a better option is suggested.

Issue #3253, opened 2019-01-04
Cannot create or import project from C: with vue ui


loicGR:
### Version
3.2.3

### Reproduction link
[http://no.github.for.this.issue.fr](http://no.github.for.this.issue.fr)

### Environment info
```
Windows 10
Firefox or IE or Brave
```

### Steps to reproduce
install vue cli on a window 10 desktop. 
Run vue ui. 
Go to  http://localhost:8000/project/select with Firefox or IE or Brave. 
Go to Create or goto Import. 
Clic on C: in the path input

### What is expected?
The directories under C: should be listed.

### What is actually happening?
Nothing appear and it is impossible to select a project in a directory under C:

loicGR:
![image](https://user-images.githubusercontent.com/8182967/50716952-6e135900-1084-11e9-9fe0-47f742b286ff.png)

sodatea:
Fixed in #3331

PR #3331, opened 2019-01-20
fix(cli-ui): folder list when root windows

jtommy:
Issue: https://github.com/vuejs/vue-cli/issues/3253

It's a node issue on Windows when using drive path without backslash.


sodatea:
Sidenote:
According to https://github.com/nodejs/node-v0.x-archive/issues/4688 It's not a bug. A more accurate fix is to ensure the argument passed to `list` has an following path seperator. Nonetheless, it's an acceptable fix.

PR #3372, opened 2019-01-26
fix: genCacheConfig return different result because linebreak

WilsonLiu95:
when I build in different plat   generate different hash because Linebreak is different.


sodatea:
Do you have a reproduction? I don't think cache config will affect the content hash

WilsonLiu95:
> Do you have a reproduction? I don't think cache config will affect the content hash
cacheIdentifier will be insert to file in comments. so affect the content hash. 

```
// CONCATENATED MODULE: ./node_modules/cache-loader/dist/cjs.js?{"cacheDirectory":"node_modules/.cache/vue-loader","cacheIdentifier":"64bd234e-vue-loader-template"}
```

sodatea:
But comments are removed in production, and in development we don't add hashes.

WilsonLiu95:
This is the same to https://github.com/vuejs/vue-loader/pull/1477.
webpack generate hash before remove comments

sodatea:
Got it. Thanks for pointing out.

One last thing before merging it: could you please run `npx eslint --fix packages/@vue/cli-service/lib/PluginAPI.js` in the project root? There're several style issues in the code.

WilsonLiu95:
😄Ok.Wait for tommorow morning~
I am not in office now.

I think webpack should generate hash by the last file to confirm hash last more time.

Users can get assets from cache.

WilsonLiu95:
> Got it. Thanks for pointing out.
> 
> One last thing before merging it: could you please run `npx eslint --fix packages/@vue/cli-service/lib/PluginAPI.js` in the project root? There're several style issues in the code.

I have run eslint~

PR #3373, opened 2019-01-26
fix(cli-service): Vue external add amd config

tweet:
If one module builded as a lib mode(umd style), and use `SystemJs` to load in browser, it would be treated as an amd module, so Vue external need add amd config.

### DIFF
--- a/packages/@vue/cli-service/lib/commands/build/resolveLibConfig.js
+++ b/packages/@vue/cli-service/lib/commands/build/resolveLibConfig.js
@@ -50,6 +50,7 @@ module.exports = (api, { entry, name, formats }, options) => {
         vue: {
           commonjs: 'vue',
           commonjs2: 'vue',
+          amd: 'Vue',
           root: 'Vue'
         }
       })


sodatea:
The UMD bundle of vue does not come with a module ID, see reasons here: https://github.com/vuejs/vue/pull/8371#issuecomment-398105340

So users are expected to configure the ID themselves.

TimYi:
Can I set a module id by passing a parameter to the build command? I didn't see any document about this. @sodatea

sodatea:
@TimYi 
```js
// vue.config.js
module.exports = {
  configureWebpack: {
    externals: {
      vue: {
        commonjs: 'vue',
        commonjs2: 'vue',
        root: 'Vue',
        amd: 'vue'
      }
    }
  }
}
```

TimYi:
@sodatea 
Appreciate it a lot!

NikitaIT:
@sodatea could you mention this in the documentation for https://cli.vuejs.org/guide/build-targets.html#library
I think it's worth it, thanks.

PR #3388, opened 2019-01-29
chore(plugin/nightwatch): bump nightwatch dependency

darrenjennings:
The yarn.lock file changes fairly drastically when I run yarn so not committing it in this PR. I was able to run the test and it ran successfully. I also ran this in the context of a new project and everything seems to work well with boilerplate generators.

```bash
$ yarn -v
1.13.0
```

image


darrenjennings:
@sodatea you may also want to bump chromedriver for the next major release since it has been bumped a couple major versions. Chrome is at version 74 and this chromedriver expects versions 70-73.

sodatea:
So now chromedriver's major version number is strictly in line with Chrome, whose release cycle is 6 weeks? In that case I think we'll need to hoist the chromedriver dependency to project root.

PR #3403, opened 2019-02-02
Add note about comments in .env files

dflock:
You can put "comments" in .env files, because lines that don't start with letters or numbers are ignored - see [here](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-service/lib/util/loadEnv.js#L17) for the code that reads these files.

I'm not sure if this constitutes "official" support for comments in .env files or not - or if it's just taking advantage of an implementation detail.

Either way, I wanted to put comments into this file, so this was useful information to me - and something that I had to figure out myself.

This commit adds a note about this to the docs, using the `#` to denote a comment line - although any non-alphanumeric char would work.

### DIFF
--- a/docs/guide/mode-and-env.md
+++ b/docs/guide/mode-and-env.md
@@ -9,10 +9,12 @@ You can specify env variables by placing the following files in your project roo
 .env.[mode].local   # only loaded in specified mode, ignored by git
 ```
 
-An env file simply contains key=value pairs of environment variables:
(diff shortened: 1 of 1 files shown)


sodatea:
Thanks for your insight. After looking into this issue deeper, I found much of our `.env` parsing logic were copied from `dotenv`. So I decided to refactor that part of code to use `dotenv` directly and link the documentation to its README https://github.com/vuejs/vue-cli/pull/3462, considering that their documentation is more comprehensive.

Issue #3433, opened 2019-02-07
Verbose console output in docker container logs


Soviut:
### Version
3.4.0

### Environment info
```
INSIDE DOCKER CONTAINER: Node:10.15.1-slim

  System:
    OS: Linux 4.9 Debian GNU/Linux 9 (stretch) 9 (stretch)
    CPU: (2) x64 Intel(R) Core(TM) i7-6850K CPU @ 3.60GHz
  Binaries:
    Node: 10.15.1 - /usr/local/bin/node
    Yarn: 1.13.0 - /usr/local/bin/yarn
    npm: 6.4.1 - /usr/local/bin/npm
  Browsers:
    Chrome: Not Found
    Firefox: Not Found
  npmPackages:
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0-beta.2
    @vue/babel-plugin-transform-vue-jsx:  1.0.0-beta.2
    @vue/babel-preset-app:  3.4.0
    @vue/babel-preset-jsx:  1.0.0-beta.2
    @vue/babel-sugar-functional-vue:  1.0.0-beta.2
    @vue/babel-sugar-inject-h:  1.0.0-beta.2
    @vue/babel-sugar-v-model:  1.0.0-beta.2
    @vue/babel-sugar-v-on:  1.0.0-beta.2
    @vue/cli-overlay:  3.4.0
    @vue/cli-plugin-babel: ^3.4.0 => 3.4.0
    @vue/cli-plugin-eslint: ^3.4.0 => 3.4.0
    @vue/cli-plugin-unit-jest: ^3.4.0 => 3.4.0
    @vue/cli-service: ^3.4.0 => 3.4.0
    @vue/cli-shared-utils:  3.4.0
    @vue/component-compiler-utils:  2.5.2
    @vue/eslint-config-airbnb: ^4.0.0 => 4.0.0
    @vue/eslint-config-standard: ^4.0.0 => 4.0.0
    @vue/preload-webpack-plugin:  1.1.0
    @vue/test-utils: ^1.0.0-beta.20 => 1.0.0-beta.29
    @vue/web-component-wrapper:  1.2.0
    eslint-plugin-vue: ^5.0.0 => 5.1.0
    jest-serializer-vue:  2.0.2
    vue: ^2.5.22 => 2.6.2
    vue-eslint-parser:  2.0.3
    vue-hot-reload-api:  2.3.1
    vue-jest:  3.0.2
    vue-loader:  15.6.2
[...]

PR #3440, opened 2019-02-09
fix(cli-service): #3433 reduced verbose logging in containers when recompiling

Soviut:
Each time a recompile happens, ~10 lines of logs are generated. When running on a native CLI this isn't a problem since the console clears and redraws in place. However, inside a container, the linear logs become very verbose. See below:

image

This patch only shows the messages on first compile. After that, only the `Compiling` and `Compiling complete` messages appear on each subsequent recompile (after the dotted red line below).

image


cedricdelpoux:
3 yers later, still not merged 🙁

PR #3444, opened 2019-02-10
Add notes for multi-page apps on references to 'prefetch' plugin

bengallienne:
I have edited the code examples to reference the plugin by the name 'prefetch-app', as this seems to be the correct name the webpack config requires from inspecting the generated config. I have tested and can confirm this works with the example that deletes the plugin altogether, but not the blacklisting example.

### DIFF
--- a/docs/guide/html-and-static-assets.md
+++ b/docs/guide/html-and-static-assets.md
@@ -39,8 +39,7 @@ By default, a Vue CLI app will automatically generate prefetch hints for all Jav
 
 The hints are injected using [@vue/preload-webpack-plugin](https://github.com/vuejs/preload-webpack-plugin) and can be modified / deleted via `chainWebpack` as `config.plugin('prefetch')`.
 
-::: tip 
-Note for multi page setups
+::: tip Note for multi page setups
 When using a multipage setup, the plugin name above should be changed to match the structure 'prefetch-{pagename}', for example 'prefetch-app'.
 :::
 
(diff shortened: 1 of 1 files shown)


bengallienne:
Ah ok that makes sense. Shall I just cancel the pull request then or do you think it is worth adding a comment to that effect somewhere in the docs?

docs/guide/html-and-static-assets.md
```diff
@@ -39,6 +39,9 @@ By default, a Vue CLI app will automatically generate prefetch hints for all Jav
 
 The hints are injected using [@vue/preload-webpack-plugin](https://github.com/vuejs/preload-webpack-plugin) and can be modified / deleted via `chainWebpack` as `config.plugin('prefetch')`.
 
+### Note for multi page setups
+When using a multipage setup, the plugin name above should be changed to match the structure 'prefetch-{pagename}', for example 'prefetch-app'.
```
sodatea: We can use the `::: tip` container here instead of adding another header (https://vuepress.vuejs.org/guide/markdown.html#custom-containers), like:

```
::: tip Note for Multi-Page Setups
When using a multipage setup, the plugin name above should be changed to match the structure 'prefetch-{pagename}', for example 'prefetch-app'.
:::
```

bengallienne: No problem, nice idea. I have done this now.

docs/guide/html-and-static-assets.md
```diff
@@ -39,6 +39,11 @@ By default, a Vue CLI app will automatically generate prefetch hints for all Jav
 
 The hints are injected using [@vue/preload-webpack-plugin](https://github.com/vuejs/preload-webpack-plugin) and can be modified / deleted via `chainWebpack` as `config.plugin('prefetch')`.
 
+::: tip 
+Note for multi page setups
```
sodatea: The header line should be placed right after `::: tip`, instead of starting a new line
See the difference in format:
image
image

bengallienne: Ok no problem, all done

sodatea: Thanks! 😄
I'll merge it soon.

PR #3463, opened 2019-02-17
Update deployment.md

shanestillwell:
For surge, you need to copy your `index.html` to `200.html` if you are using HTML5 history mode.


shanestillwell:
Now I can see that this is referenced above and pointing to vue-router... I'll close this.

Issue #2983, opened 2018-11-20
Nondeterministic builds with multipage app and default babel polyfills


jcushman:
### Version
3.1.1

### Reproduction link
[https://github.com/jcushman/vue_cli_polyfills_bug](https://github.com/jcushman/vue_cli_polyfills_bug)

### Node and OS info
Node 11.1.0 / npm 6.4.1 / OS X 10.12.6

### Steps to reproduce
- Install package.json requirements (I used `npm i`)
- Repeatedly run `sudo rm -r node_modules/.cache/; npx vue-cli-service build; git status`

### What is expected?
`git status` should print `On branch master; nothing to commit, working tree clean`, as dist/ files are in source control and src/ has not changed.

### What is actually happening?
**About half the time**, `git status` reports files modified:

	modified:   dist/js/page1.js
	modified:   dist/js/page1.js.map
	modified:   dist/js/page2.js
	modified:   dist/js/page2.js.map
	modified:   dist/js/page3.js
	modified:   dist/js/page3.js.map

The changes look like the adding or removing of three lines of whitespace, and the adding or removing of these lines:

```
/* harmony import */ var core_js_modules_es6_array_iterator__WEBPACK_IMPORTED_MODULE_0__ = __webpack_require__("cadf");
/* harmony import */ var core_js_modules_es6_array_iterator__WEBPACK_IMPORTED_MODULE_0___default = /*#__PURE__*/__webpack_require__.n(core_js_modules_es6_array_iterator__WEBPACK_IMPORTED_MODULE_0__);
/* harmony import */ var core_js_modules_es6_promise__WEBPACK_IMPORTED_MODULE_1__ = __webpack_require__("551c");
/* harmony import */ var core_js_modules_es6_promise__WEBPACK_IMPORTED_MODULE_1___default = /*#__PURE__*/__webpack_require__.n(core_js_modules_es6_promise__WEBPACK_IMPORTED_MODULE_1__);
/* harmony import */ var core_js_modules_es7_promise_finally__WEBPACK_IMPORTED_MODULE_2__ = __webpack_require__("097d");
/* harmony import */ var core_js_modules_es7_promise_finally__WEBPACK_IMPORTED_MODULE_2___default = /*#__PURE__*/__webpack_require__.n(core_js_modules_es7_promise_finally__WEBPACK_IMPORTED_MODULE_2__);
```

---
[...]

PR #3470, opened 2019-02-19
fix(babel-preset-app): multiPage repo add option entryFileList

WilsonLiu95:
To inject polyfills to all entry file.
Multi page repo now will be only inject polyfills to the first entry file.And the queue of entry file list may be different.So the hash may change in different machine.
We should inject polyfills to all entry file.

### DIFF
--- a/packages/@vue/babel-preset-app/README.md
+++ b/packages/@vue/babel-preset-app/README.md
@@ -97,3 +97,9 @@ Set to `false` to disable JSX support.
 - Default: `false`.
 
 Setting this to `true` will generate code that is more performant but less spec-compliant.
+
+### entryFiles
+
+- Default: `[]`
+
+Multi page repo use entryFiles to ensure inject polyfills to all entry file.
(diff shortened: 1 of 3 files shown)


sodatea:
Thank you so much for this pull request!
So I finally found out what caused this issue: https://github.com/vuejs/vue-cli/issues/2983

Actually, due to the presence of `thread-loader` it is not possible to reliably determine the real entry file. I think we can just remove the `entryFile` logic and only rely on the `entryFileList` option, and then automatically set this field in `@vue/cli-plugin-babel` so that users don't need to be aware of this change (this can be done in a separate pull request though).

packages/@vue/babel-preset-app/README.md
```diff
@@ -97,3 +97,9 @@ Set to `false` to disable JSX support.
 - Default: `false`.
 
 Setting this to `true` will generate code that is more performant but less spec-compliant.
+
+### entryFileList
```
sodatea: Let's rename it to `entryFiles` to keep align with `polyfills`
```suggestion
### entryFiles
```

WilsonLiu95:
yes,  due to the presence of thread-loader and cache-loader.When JS file is cache then the first file will be change.

PR #3472, opened 2019-02-19
test(cli-service): add test for multi page history api fallback

jkzing:
Add test case for https://github.com/vuejs/vue-cli/pull/3181, it was merged just before I pushed to original branch....😂


sodatea:
😂

PR #3485, opened 2019-02-20
Add .mjs files as transpileable files

mnugter:
I'm not 100% sure but the option transpileDependencies in vue.config.js did not work for dependencies that use .mjs files. This small regex change fixes that problem.


mnugter:
Failed CI AppVeyor seems a random failure on a test that took too long?

kuoruan:
@sodatea What about the `cjs` files? can it be included in the regex? `/\.(c|m)?jsx?$/`

When I transpile some lib with cjs files, I found it can not go through babel-loader.

PR #3487, opened 2019-02-21
Simplify lint-staged config in `@vue/cli-plugin-eslint`

fengyuanchen:


fengyuanchen:
The corresponding test case had been updated.

PR #3495, opened 2019-02-21
docs: fix example changing main file

disitec:
In the giving example
```js
  fs.writeFileSync(api.entryFile, contentMain, { encoding: 'utf-8' })
```
it was being written back the original content instead of the modified lines. I've changed the approach adding the new line to the array and then joining it back.

```js
  lines.splice(renderIndex, 0, `  router,`)
  const newContent = lines.join('\r\n')
  fs.writeFileSync(api.entryFile, newContent, { encoding: 'utf-8' })
```


sodatea:
Thanks for this pull request and sorry for the late reply.
But we found https://github.com/vuejs/vue-cli/pull/3656 to be a better fix for this example.
Thanks again though, your help is much appreciated.

Issue #3476, opened 2019-02-19
"vue-cli-service serve --copy" throws in WSL due to clipboardy.write()


jonaskuske:
### Version
3.4.0

### Environment info
```
Environment Info:

  System:
    OS: Linux 4.4 Ubuntu 16.04.5 LTS (Xenial Xerus)
    CPU: (8) x64 Intel(R) Core(TM) i7-4790 CPU @ 3.60GHz
  Binaries:
    Node: 10.13.0 - /usr/bin/node
    Yarn: 1.12.3 - /usr/bin/yarn
    npm: 6.8.0 - ~/.npm-global/bin/npm
  Browsers:
    Chrome: Not Found
    Firefox: Not Found
  npmPackages:
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0-beta.2
    @vue/babel-plugin-transform-vue-jsx:  1.0.0-beta.2
    @vue/babel-preset-app:  3.4.0
    @vue/babel-preset-jsx:  1.0.0-beta.2
    @vue/babel-sugar-functional-vue:  1.0.0-beta.2
    @vue/babel-sugar-inject-h:  1.0.0-beta.2
    @vue/babel-sugar-v-model:  1.0.0-beta.2
    @vue/babel-sugar-v-on:  1.0.0-beta.2
    @vue/cli-overlay:  3.4.0
    @vue/cli-plugin-babel: ^3.4.0 => 3.4.0
    @vue/cli-plugin-eslint: ^3.4.0 => 3.4.0
    @vue/cli-plugin-pwa: ^3.4.0 => 3.4.0
    @vue/cli-plugin-unit-jest: ^3.4.0 => 3.4.0
    @vue/cli-service: ^3.4.0 => 3.4.0
    @vue/cli-shared-utils:  3.4.0
    @vue/component-compiler-utils:  2.5.2
    @vue/eslint-config-prettier: ^4.0.1 => 4.0.1
    @vue/preload-webpack-plugin:  1.1.0
    @vue/test-utils: ^1.0.0-beta.20 => 1.0.0-beta.29
    @vue/web-component-wrapper:  1.2.0
    eslint-plugin-vue: ^5.0.0 => 5.1.0
    jest-serializer-vue:  2.0.2
    vue: ^2.5.22 => 2.6.5
    vue-eslint-parser:  2.0.3
    vue-hot-reload-api:  2.3.1
    vue-jest:  3.0.3
    vue-loader:  15.6.2
[...]

PR #3503, opened 2019-02-23
fix(cli-service): catch exception if "copy to clipboard" fails (issue #3476)

jonaskuske:
As explained in issue #3476 , `clipboardy` throws an exception on some platforms, notably Windows Subsystem for Linux. This results in an `unhandledPromiseRejection` when using the `--copy` option.
As of right now this is just very annoying, but those rejections are deprecated and will completely stop the process in the future and so forth – you know the drill.

To prevent this I changed the copying to use `writeSync()` and wrapped it in a `try catch`-block. This also means we only log `(copied to clipboard)` now if copying was actually successful.

I also added a tip to the docs, it looks like this:

___

__copy notice

___

If you think this is unnecessary and we can just fail silently without mentioning it somewhere, we can also revert this commit of course :)

### DIFF
--- a/docs/guide/cli-service.md
+++ b/docs/guide/cli-service.md
@@ -52,6 +52,11 @@ Options:
   --https   use https (default: false)
 ```
 
+::: tip --copy
+Copying to clipboard might not work on a few platforms.  
+If copying was successful, `(copied to clipboard)` is displayed next to the local dev server URL.
+:::
+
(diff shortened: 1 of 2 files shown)


jonaskuske:
By the way, it should be possible to get support for WSL upstream into `clipboardy` (by detecting WSL and piping to `clip.exe`) and I'll try to land that once I have time. Still, the error handling on our side is needed imo.

PR #3510, opened 2019-02-24
fix: add missing package keys to the sort array (issue #3509)

### LINKED ISSUE #3509: extendPackage changes the order of the keys
### Version
3.4.1

### Reproduction link
[https://github.com/elevatebart/test-vue-lib](https://github.com/elevatebart/test-vue-lib)

### Environment info
```
MacOS, node 10, yarn 1.13
```

### Steps to reproduce
When using the followin code in my plugin, I get changes on the order of keys in packages.json. It is very unsetling for users.
```js
api.extendPackage({
    scripts: {
      styleguide: 'vue-cli-service styleguidist',
      'styleguide:build': 'vue-cli-service styleguidist:build'
    }
  })
```

### What is expected?
I expect the extendPackages touches only what it is intended to and leves the rest alone.

The folloing keys are moved to the end of the file:
main, files, unpkg

---
Found while inverstigating https://github.com/vue-styleguidist/vue-cli-plugin-styleguidist/issues/34

elevatebart:
closes #3509


packages/@vue/cli/lib/Generator.js
```diff
@@ -181,9 +181,13 @@ module.exports = class Generator {
       'private',
       'description',
       'author',
+      'main',
+      'unpkg',
```
LinusBorg: We could also inlcude the `jsDelivr` field if we add `unpkg`
```suggestion
      'jsDelivr',
      'unpkg',
```

packages/@vue/cli/lib/Generator.js
```diff
@@ -181,9 +181,13 @@ module.exports = class Generator {
...
+      'files',
       'scripts',
```
LinusBorg: I would prefer for the scripts to be above the file-related fields (main, files etc.), because as a developer I have to look for those quite a lot.

This is of course subjective - other's feelings about this?

packages/@vue/cli/lib/Generator.js
```diff
@@ -181,9 +181,13 @@ module.exports = class Generator {
       'private',
       'description',
       'author',
+      'main',
```
LinusBorg: While we are at it, there are some other related fields:
```suggestion
      'main',
      'module',
      'js:next',
      'browser',
```

packages/@vue/cli/lib/Generator.js
andreasvirkus: Agreed. This is also a well-established best practise imo

elevatebart: You guys are absolutely right. 
Great remarks. 
Hope my  fixes are AOK ;)

PR #3523, opened 2019-02-26
docs: Document jest's `transformIgnorePatterns` option to avoid pitfall

LinusBorg:
I found that many people have issues running tests with jest, receiving an error like `"unexpected token "import"`.

The most common reason - as far as I can tell - are packages in their node_modules that contain untranspiled ES6 imports.

While we offer the `transpileDependencies` option, jest doesn't respect it, and rightfully so, since the issues solved by that option are often different ones (transpiling for browser compatibility).

Instead, people have to adjust jest's `transformIgnorePatters` option when one of the packages they use has to be transformed by `babel-jest` or `vue-jest`.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -57,7 +57,7 @@ Deprecated since Vue CLI 3.3, please use [`publicPath`](#publicPath) instead.
   ``` js
   module.exports = {
     publicPath: process.env.NODE_ENV === 'production'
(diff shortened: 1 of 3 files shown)


LinusBorg:
Will undo the unnecessary formatting changes that my editor did.

LinusBorg:
Should be good now. Jeeeeez....

PR #3525, opened 2019-02-26
fix: proxy should warn when array is passed.

### LINKED ISSUE #3524: Proxy settings do not accept an array but array passes validations and fails silently
### Version
3.4.1

### Steps to reproduce
in vue config set proxy config to:
```
devServer: {
  overlay: {
    warnings: true,
    errors: true
  },
  host: '0.0.0.0',
  disableHostCheck: true,
  index: '',
  proxy:[{
    context: ['/**', '!/app.js', '!/sockjs-node/**'],
    target: 'http://localhost:8000'
  }]
}
```

### What is expected?
The http-proxy-middleware package accepts an array as a config option (https://github.com/chimurai/http-proxy-middleware#context-matching) but this is not supported by vue-cli.  Vue cli should warn the user that the array syntax is unsupported.

No error is raised because in the prepareProxy function in prepareProxy.js typeof proxy !== 'object' is used to check that the proxy config is an object, but an array passes this test.  The typeof an Array is 'object'.

LinusBorg:
close #3524

### DIFF
--- a/packages/@vue/cli-service/lib/util/prepareProxy.js
+++ b/packages/@vue/cli-service/lib/util/prepareProxy.js
@@ -27,7 +27,7 @@ module.exports = function prepareProxy (proxy, appPublicFolder) {
   if (!proxy) {
     return undefined
   }
-  if (typeof proxy !== 'object' && typeof proxy !== 'string') {
+  if (Array.isArray(proxy) || (typeof proxy !== 'object' && typeof proxy !== 'string')) {
     console.log(
       chalk.red(
         'When specified, "proxy" in package.json must be a string or an object.'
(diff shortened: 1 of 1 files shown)


packages/@vue/cli-service/lib/util/prepareProxy.js
```diff
@@ -27,7 +27,7 @@ module.exports = function prepareProxy (proxy, appPublicFolder) {
   if (!proxy) {
     return undefined
   }
-  if (typeof proxy !== 'object' && typeof proxy !== 'string') {
+  if (Array.isArray(proxy) || (typeof proxy !== 'object' && typeof proxy !== 'string')) {
```
sodatea: There's a `\u00A0` before `(typeof proxy`

LinusBorg: Thanks, fixing ...

PR #3530, opened 2019-02-27
fix: Fixes "Changing main file"-example

ChristianKienle:


sodatea:
Thanks for spotting this and sorry for the late reply.
But we found https://github.com/vuejs/vue-cli/pull/3656 to be a better fix for this example.
Thanks again though, your help is much appreciated.

Issue #761, opened 2018-02-01
Use tslint-config-prettier for typescript


alexjoverm:
### What problem does this feature solve?
Similar to how the new vue-cli 3 has [eslint-config-prettier](https://github.com/vuejs/vue-cli/tree/dev/packages/%40vue/eslint-config-prettier) in order to use both without conflicts, I think it can be useful to add [tslint-config-prettier](https://github.com/alexjoverm/tslint-config-prettier) to achieve the same thing when using typescript.

### What does the proposed API look like?
It should be transparent to the user as long as it selects typescript and a formatter on the cli questions

yyx990803:
Does `tsling-config-prettier` support `--fix` via TSLint? Or does it require the user to setup prettier themselves?

alexjoverm:
Right now it disables conflicting rules, so yeah it assumes prettier is already configured. In that way, it works just as [eslint-config-prettier](https://github.com/vuejs/vue-cli/tree/dev/packages/%40vue/eslint-config-prettier).

eladcandroid:
That would be great (TSLint + Prettier option in the CLI)

backbone87:
you can simply `npm i -D prettier tslint-plugin-prettier tslint-config-prettier` and setup your tslint.json:
```json
{
  "extends": [
    "tslint-config-prettier"
  ],
  "rulesDirectory": ["tslint-plugin-prettier"],
  "rules": {
    "prettier": true
  }
}
```

the only problem i had with this is that when running `npm run lint` the prettier rule reports alot of consecutive newline errors. (and doesnt seem to be able to autofix...).

the problem is here:
https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-typescript/lib/tslint.js#L46

this pads the extracted script with newlines (see https://www.npmjs.com/package/vue-template-compiler#pad )
while the tslint rule for consecutive newlines is disabled by `npm run lint`, prettier has its own rules.

wminshew:
Is there a temporary workaround for this or are you just ignoring those error messages?

backbone87:
There is no workaround. There needs to be an option in vue-template-compilers parseComponent to comment out newlines (like vue-parser does) when `lang="ts"` is specified. After that, the option can be used downstream (if not enabled by default).

backbone87:
i have opened a PR which fixes script content padding inside the vue-cli tslint command

sodatea:
[...]

PR #3561, opened 2019-03-03
fix(tslint): correctly pad the script content

backbone87:
this avoids manipulating the tslint ruleset, which is only necessary
because the padding of vueCompiler does not work for lang="ts" scripts.
this also makes vue-cli tslint work with tslint-plugin-prettier

prereq for: https://github.com/vuejs/vue-cli/issues/761
unexpected upstream behavior:
https://github.com/vuejs/vue/blob/dev/src/sfc/parser.js#L119

### DIFF
--- a/packages/@vue/cli-plugin-typescript/lib/tslint.js
+++ b/packages/@vue/cli-plugin-typescript/lib/tslint.js
@@ -26,15 +26,21 @@ module.exports = function lint (args = {}, api, silent) {
 
   const patchWriteFile = () => {
     fs.writeFileSync = (file, content, options) => {
-      if (isVueFile(file)) {
-        const parts = vueFileCache.get(path.normalize(file))
-        if (parts) {
-          parts.content = content
-          const { before, after } = parts
-          content = `${before}\n${content.trim()}\n${after}`
-        }
+      const parts = vueFileCache.get(path.normalize(file))
+      if (!parts) {
+        return writeFileSync(file, content, options)
       }
-      return writeFileSync(file, content, options)
+
+      // update cached content
+      parts.content = content
+
+      content = content.slice(parts.paddingOffset)
+      // remove one leading and trailing linebreak, if any
+      content = content.replace(/^\r?\n/, '').replace(/\r?\n$/, '')
+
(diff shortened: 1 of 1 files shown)


backbone87:
notes on leading and trailing linebreaks inside script content:

first leading linebreak (imediately after the script tag) is preserved:
```
<script>
import Vue from 'vue';
</script>
```
-> will become
```
<script>
import Vue from 'vue';
</script>
```

no leading linebreak is preserved:
```
<script>import Vue from 'vue';
</script>
```
-> will become
```
<script>import Vue from 'vue';
</script>
```


trailing linebreak is added, if not present (after linter fixing)
```
<script>
import Vue from 'vue';</script>
```
-> will become
```
<script>
import Vue from 'vue';
</script>
```

therefore this is not possible:
```
<script>import Vue from 'vue';</script>
```
-> will become
```
<script>import Vue from 'vue';
</script>
```

backbone87:
after experimenting more with the single line scripts, i noticed that vscode + vetur dont correctly syntax color them anyway, so i dropped the preservation of single line scripts completely.

there is just one of caveat:
`<script>import Vue from 'vue';`: no leading linebreak in script content wont be reported as a lint failure (at least for prettier, i guess its the some for consecutive newlines tslint rule). however when the file is written (because of another lint failure), this leadine linebreak is added.`

so basically with prettier the following would be reported as error:
```
<script>

import Vue from 'vue';
```
while
```
<script>
import Vue from 'vue';
```
and
```
<script>import Vue from 'vue';
```
causes no error.
if the file gets written (because of any lint error in the file), all result in the follow:
```
<script>
import Vue from 'vue';
```

this could be solved by creating a custom tslint rule for vue files, where this special case is handled

backbone87:
any update on this? its really annoying not being able to use the lint command especially in CI

sodatea:
Thanks for the PR.

As TSLint has been [officially deprecated](https://github.com/palantir/tslint/issues/4534), we do not plan to add any new feature to this functionality and it will be removed in the next major.
So I'm closing this PR.

Thanks anyway!

PR #3574, opened 2019-03-05
fix(docs): update zeit now deployment process considering v2

svtfrancisco:
Now is currently on version 2, and it use that version by default when someone want to make a deploy, then it let us two choices:
- A) Maintain current docs about v1 deployment but set version on example JSON.
- B) Update docs considering the the advent of v2 (11/2018)

I choose option B, then i made the changes and just commit it.

### Before
captura de tela 2019-03-05 as 00 24 51

### After
captura de tela 2019-03-05 as 00 25 06

#### Reasons to try to contribute
- Docs need to be always updated :)

#### Refs
[ZEIT - Now about Vue.js deployment (v2)](https://zeit.co/guides/deploying-vuejs-to-now/)

Sorry my english, only a Brazilian servant.

### DIFF
--- a/docs/guide/deployment.md
+++ b/docs/guide/deployment.md
@@ -223,6 +223,7 @@ npm install -g now
     {
       "version": 2,
       "name": "my-vue-project",
+      "alias": "my-vue-project",
       "builds": [{ "src": "package.json", "use": "@now/static-build" }],
       "routes": [
         { "src": "^/js/(.*)", "dest": "/js/$1" },
--- a/docs/guide/deployment.md
+++ b/docs/guide/deployment.md
@@ -221,30 +221,25 @@ npm install -g now
 
     ```json
     {
-      "name": "my-example-app",
-      "type": "static",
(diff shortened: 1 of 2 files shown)


LinusBorg:
Possible duplicate with #3560

svtfrancisco:
> Possible duplicate with #3560

To work right must exists a "now-build" script on `package.json` #3560 don't have this warn/instruction.

Virgiel:
I think you are right to use the option B as I see no reason not to get the advantage of now 2 with a vue project

Virgiel:
Now CLI has just reached version 14.2.0 and the aliasing commands has changed

`now` build and give a default alias
`now --target production` build and give alias from the config

more information at https://zeit.co/blog/automatic-aliasing

NataliaTepluhina:
Thank you for working on this! Unfortunately, need to close the PR in favor of https://github.com/vuejs/vue-cli/pull/4232

Issue #3671, opened 2019-03-19
Build fails with @babel/preset-env 7.4.0


AegirLeet:
### Version
3.5.1

### Environment info
```

Environment Info:

  System:
    OS: Linux 4.18 Debian GNU/Linux 9 (stretch) 9 (stretch)
    CPU: (12) x64 Intel(R) Xeon(R) CPU D-1531 @ 2.20GHz
  Binaries:
    Node: 11.12.0 - /usr/bin/node
    Yarn: 1.15.2 - /usr/bin/yarn
    npm: 6.7.0 - /usr/bin/npm
  Browsers:
    Chrome: Not Found
    Firefox: Not Found
  npmPackages:
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0-beta.2
    @vue/babel-plugin-transform-vue-jsx:  1.0.0-beta.2
    @vue/babel-preset-app:  3.5.1
    @vue/babel-preset-jsx:  1.0.0-beta.2
    @vue/babel-sugar-functional-vue:  1.0.0-beta.2
    @vue/babel-sugar-inject-h:  1.0.0-beta.2
    @vue/babel-sugar-v-model:  1.0.0-beta.2
    @vue/babel-sugar-v-on:  1.0.0-beta.2
    @vue/cli-overlay:  3.5.1
    @vue/cli-plugin-babel: ^3.5.0 => 3.5.1
    @vue/cli-plugin-eslint: ^3.5.0 => 3.5.1
    @vue/cli-service: ^3.5.0 => 3.5.1
    @vue/cli-shared-utils:  3.5.1
    @vue/component-compiler-utils:  2.6.0
    @vue/preload-webpack-plugin:  1.1.0
    @vue/web-component-wrapper:  1.2.0
    eslint-plugin-vue: ^5.0.0 => 5.2.2
    vue: ^2.6.6 => 2.6.9
    vue-eslint-parser:  5.0.0
    vue-hot-reload-api:  2.3.3
    vue-loader:  15.7.0
    vue-style-loader:  4.1.2
    vue-template-compiler: ^2.5.21 => 2.6.9
    vue-template-es2015-compiler:  1.9.1
  npmGlobalPackages:
    @vue/cli: Not Found


```

### Steps to reproduce
```
yarn global add @vue/cli
vue create -d vue-cli-test
cd vue-cli-test
yarn build
```

### What is expected?
A successful build.

### What is actually happening?
[...]

PR #3673, opened 2019-03-19
Fix import path for isPluginRequired and use correct built-ins

perkola:
Relates to #3671

### DIFF
--- a/packages/@vue/babel-preset-app/index.js
+++ b/packages/@vue/babel-preset-app/index.js
@@ -14,8 +14,8 @@ const defaultPolyfills = [
 ]
 
 function getPolyfills (targets, includes, { ignoreBrowserslistConfig, configPath }) {
-  const { isPluginRequired } = require('@babel/preset-env')
-  const builtInsList = require('@babel/preset-env/data/built-ins.json')
+  const { isPluginRequired } = require('@babel/preset-env/lib/filter-items')
+  const builtInsList = require('@babel/preset-env/data/corejs2-built-ins.json')
   const getTargets = require('@babel/preset-env/lib/targets-parser').default
   const builtInTargets = getTargets(targets, {
     ignoreBrowserslistConfig,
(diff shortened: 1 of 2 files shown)


andreineculau:
I guess you'd also want to mention in package.json that this change requires `@babel/preset-env` at `^7.4.0`

perkola:
> I guess you'd also want to mention in package.json that this change requires `@babel/preset-env` at `^7.4.0`

Good catch! I was a bit too quick. I have updated the PR now.

packages/@vue/babel-preset-app/index.js
```diff
@@ -14,8 +14,8 @@ const defaultPolyfills = [
 ]
 
 function getPolyfills (targets, includes, { ignoreBrowserslistConfig, configPath }) {
-  const { isPluginRequired } = require('@babel/preset-env')
-  const builtInsList = require('@babel/preset-env/data/built-ins.json')
+  const { isPluginRequired } = require('@babel/preset-env/lib/filter-items')
+  const builtInsList = require('@babel/preset-env/data/corejs2-built-ins.json')
```
nicolo-ribaudo: Hi, Babel maintainer here 👋

We just released v7.4.1, which should fix the issue. The `builtInsList` change is correct, but I'd ask you to revert line 17 (it's better not to rely on internal files, when possible). You'll need to update the `preset-env` version in package.json.

perkola:
This PR becomes obsolete with the release of `@babel/preset-env` `v.7.4.1`.

packages/@vue/babel-preset-app/index.js
perkola: Hi! Thank you for the quick reply. I noticed that you also aliased `cosejs2` so both my changes are unnecessary right now. Perhaps we still need to change the path of `built-ins.json` if the alias is removed in the future.

nicolo-ribaudo:
Please update `package.json` to `^7.4.1` to avoid accidentally downloading `7.4.0`

PR #3674, opened 2019-03-19
Use correct built-ins path and require babel-preset 7.4.1

perkola:
See #3671.


sodatea:
Fixed by #3696 and #3710

PR #3687, opened 2019-03-20
perf(ui): improve get folder list to use Promises instead of sync

pikax:
Use Promises to get if the file `isHidden` and `isDirectory`, this improves the performance when doing `project/select`.

On windows was taking around 1\~2seconds to fetch a directory (using SSD), with this I manage to get 300\~600ms.


packages/@vue/cli-ui/apollo-server/connectors/folders.js
```diff
@@ -13,7 +13,7 @@ const pkgCache = new LRU({
 
 const cwd = require('./cwd')
 
-function isDirectory (file) {
+function isDirectorySync (file) {
```
pikax: Don't know if replacing this function to fully sync will break at other place, please advice if is save to use the async version

sodatea (approved):

PR #3688, opened 2019-03-20
feat(ui): add copyContent to clipboard button on terminal component

pikax:
Add button to copy all the terminal content or only selection to the clipboard

Is quite annoying select all the text when doing an `webpack inspect`, because `CTRL+ COMMAND` is not supported in terminal


Akryum:
@sodatea Seems like a good workaround until we find why Ctrl+C is not working on xterm like in VS Code.

Issue #3678, opened 2019-03-20
core-js module error


kobe990:
### Version
3.5.1

### Environment info
```
mac os 10.14.3
```

### Steps to reproduce
vue-cli-service serve --mode development

### What is expected?
run the project correctly

### What is actually happening?
With useBuiltIns option, required direct setting of corejs option
98% after emitting CopyPlugin
ERROR Failed to compile with 36 errors 10:20:25
These dependencies were not found:

core-js/modules/es6.array.iterator in ./src/main.js, ./src/webapp/main.js
core-js/modules/es6.function.name in ./src/main.js
core-js/modules/es6.number.constructor in ./src/assets/js/utils.js
core-js/modules/es6.object.assign in ./src/main.js, ./src/webapp/main.js
core-js/modules/es6.object.to-string in ./src/main.js, ./src/assets/js/utils.js and 2 others
core-js/modules/es6.promise in ./src/main.js, ./src/webapp/main.js
core-js/modules/es6.regexp.match in ./src/assets/js/utils.js
core-js/modules/es6.regexp.replace in ./src/main.js, ./src/assets/js/gt.js and 3 others
core-js/modules/es6.regexp.split in ./src/assets/js/utils.js
core-js/modules/es6.regexp.to-string in ./src/main.js, ./src/assets/js/utils.js and 1 other
core-js/modules/es6.string.includes in ./src/store.js, ./src/main.js and 2 others
core-js/modules/es6.string.iterator in ./node_modules/_cache-loader@2.0.1@cache-loader/dist/cjs.js??ref--12-0!./node_modules/_babel-loader@8.0.5@babel-loader/lib!./node_modules/_cache-loader@2.0.1@cache-loader/dist/cjs.js??ref--0-0!./node_modules/_vue-loader@15.7.0@vue-loader/lib??vue-loader-options!./src/views/assets/Index.vue?vue&type=script&lang=js&
core-js/modules/es6.typed.uint8-array in ./src/assets/js/utils.js
core-js/modules/es7.array.includes in ./src/store.js, ./src/main.js and 3 others
core-js/modules/es7.promise.finally in ./src/main.js, ./src/webapp/main.js
[...]

PR #3692, opened 2019-03-22
Update package.json (@babel/core@7.4.0 issue)

oakromulo:
please check these resources below on the reasoning behind this problematic babel bug:

- https://stackoverflow.com/questions/55251983/what-does-this-error-mean-with-usebuiltins-option-required-direct-setting-of

- https://github.com/parcel-bundler/parcel/issues/2819
- https://github.com/parcel-bundler/parcel/commit/53c09b3cfa3a14bc0dc9875950b31e43d4a9a2e4

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,7 @@
     "yorkie": "^2.0.0"
   },
   "resolutions": {
-    "@babel/core": "^7.0.0",
+    "@babel/core": "^7.0.0 <7.4.0",
     "@babel/preset-env": "^7.0.0",
     "babel-core": "7.0.0-bridge.0",
     "ps-tree": "^1.1.1",


oakromulo:
see also:
https://github.com/vuejs/vue-cli/issues/3678

oakromulo:
directing users to this workaround is also a solid temporary solution:
`"resolutions": { "@vue/cli-plugin-babel/**/@babel/core": "^7.0.0 <7.4.0" }`

oakromulo:
p.s. the next babel/core minor release might fix it too, see https://github.com/babel/babel/pull/7646

sodatea:
Thanks for this PR but we prefer to use the latest version of babel unless there's a serious bug that can't be circumvented. And #3678 was fixed in #3710

PR #3716, opened 2019-03-27
fix: downgrade @babel/preset-env temporarily

sodatea:
seems there're still problems with it, e.g. when core-js 3 is hoisted as the root level dependency.

### DIFF
--- a/packages/@vue/babel-preset-app/index.js
+++ b/packages/@vue/babel-preset-app/index.js
@@ -123,8 +123,7 @@ module.exports = (context, options = {}) => {
     include,
     exclude: polyfills.concat(exclude || []),
     shippedProposals,
-    forceAllTransforms,
-    corejs: 2
+    forceAllTransforms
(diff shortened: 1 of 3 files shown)


jacekkarczmarczyk:
@sodatea Can you describe the problem? I'm not sure if it doesn't affect my app or I just don't see it

sodatea:
@jacekkarczmarczyk @babel/preset-env 7.4+ introduced an indirect dependency of core-js@3, in some cases, core-js@3 is incorrectly hoisted to `node_modules` root. Meanwhile, `@babel/preset-env` injects code like `import "core-js/..."`, which expects core-js 2 directory structure, that will cause problem.

I'm not sure how it happened as I personally didn't encounter this problem, but I guess it may happen when you use `transpileDependencies`.

To ultimately solve this problem we may need to add core-js@2 as direct dependency to the generated projects.

PR #3730, opened 2019-03-29
fix: should not add polyfills from transform-runtime plugin

sodatea:
It is due to a misunderstanding when upgrading babel from 7.0.0-beta.47
to 7.x.
In Vue CLI, usage-based polyfills are **only** added by @babel/preset-env
so that it avoids overheads. `transform-runtime` plugin adds polyfills
regardless of env targets so we must disable its polyfill feature.
Though, there're edge cases where runtime-helpers' polyfill dependency
are not correctly added (https://github.com/babel/babel/issues/7597),
so we also need to whitelist `@babel/runtime` for webpack to transpile.


aasailan:
i use vuecli 3.9.2  but the project add polyfills from transform-runtime plugin。what can i do?

PR #3736, opened 2019-03-31
fix(generator): add core-js as direct dependency of generated projects

sodatea:


cyrilf:
Hey @sodatea,

Could you explain me what is this change doing?
I'm a bit confused if I have to keep `core-js` as a `dependency` of my project or if I could move it to the `dev-dependencies` (or get rid of it all in all).

Thanks for your time! :v:

sodatea:
@cyrilf 
If you are using babel to transpile your code, babel may add some polyfill imports to the transpiled code for it to function correctly.
E.g.
```js
const a = new Map()
```
will be transpiled into
```js
import "core-js/modules/es6.map";
var a = new Map();
```
as map is not natively supported in IE9.

So technically speaking the `core-js` package **is a dependency** of your project.

In the past we utilized the [package hoisting feature of npm](https://npm.github.io/how-npm-works-docs/npm3/how-npm3-works.html) and hide it behind `@vue/cli-plugin-babel`.
But after core-js 3 came out:
1. the internal directory structure of `core-js` changed a lot, we need to make sure users only get the correct version of `core-js` so as not to break their app
2. other packages in the project's dependency chain may have already depended on `core-js@3`, so we can't be sure if the root level `core-js` dependency is of the right version (as npm package hoisting is non-deterministic) without explicitly listing it in `package.json`.

To conclude:
you should keep it as a dependency.

cyrilf:
Got it! :+1: 
Thanks a lot for this clear and explicit explanation. :ok_hand:

zWingz:
Thanks for explanation @sodatea

PR #3787, opened 2019-04-10
just use portfinder when user not set port

### LINKED ISSUE #3719: export the port
In most cases, the port is our user-defined port.
But sometimes users will start multiple services. At this time vue-cli will use portfinder to start a new port, but the port is not exposed.

```js
proxy: {
 target: `http://localhost:${port}/mock`,
}
```
Since the port has changed, the request cannot be proxy correctly.

process.env.VUE_CLI.port 
Or any other way

xsbchen:
fix #3719


LinusBorg:
This will result in a cryptic error if the user-specified port is not free.

Which is why we still run portfinde even if the user provided a port.

This PR would at least have to provide a decent error handling & console message.

xsbchen:
@LinusBorg added!

PanJiaChen:
image

Maybe this kind of interaction is better?

IndexXuan:
any further ? It should give options to not use find porter

Issue #2697, opened 2018-10-09
It always uses `index-default.html` when give `template` property


fengxinming:
### Version
3.0.4

### Reproduction link
[https://github.com/fengxinming/vuecli3-demo](https://github.com/fengxinming/vuecli3-demo)

### Node and OS info
node v8.11.1, npm 6.4.1, macOS

### Steps to reproduce
npm i
npm run build

### What is expected?
`public/index.pug` is compiled to `dist/index.html`

### What is actually happening?
`index-default.html` is compiled to `dist/index.html`

---
190  // resolve page index template
191  const hasDedicatedTemplate = fs.existsSync(api.resolve(template))
192  if (hasDedicatedTemplate) {
193         publicCopyIgnore.push(template)
194  }

if I use pug template as property `template`, then `hasDedicatedTemplate` still can be `false`

KancerEzeroglu:
Hi, is somebody working on this issue? If not, I would like to work on it.

LinusBorg:
It seems to be free, so go ahead!

Thanks! :)

KancerEzeroglu:
Hi @LinusBorg ,

Do you have any suggestion for a starting point?
Today I have a chance to have a look.

fengxinming:
@KancerEzeroglu 看这个文件 `packages/@vue/cli-service/lib/config/app.js` 里的第148行代码，有两部分配置，一部分是多页配置；另一部分是单页配置。

fengxinming:
@KancerEzeroglu You may find the line 148 of that file `packages/@vue/cli-service/lib/config/app.js` which starts to resolve HTML files.

jeneser:
Any progress? @fengxinming Can you update the reproduction link?

Issue #3061, opened 2018-12-01
Add prop in vue.config.js to change vue src directory (copy)


cthames:
### What problem does this feature solve?

I apologize. The [previous issue](https://github.com/vuejs/vue-cli/issues/1134#issuecomment-384856754) had comments disabled as I was in the middle of responding to the comments by @sodatea and @yyx990803 . There is confusion that I'm trying to address. I'm talking about **existing projects** see my comment below. I think it's in the best interest and for increased adoption of Vue JS that a thorough discussion from both sides is had instead of ending it prematurely before I even had a chance to respond. I understand wanting to limit for **new** projects, but **existing projects** is where greater adoption for Vue JS is. It will also allow for existing projects to transition from another front end to Vue JS slowly without having to re-write an entire app. Businesses aren't going to spend a bunch of money upfront to rewrite an entire app. A slow transition is more feasible.

### Feature

Having the Vue src files and directories in the /src directory works for simple projects, but for more complex projects it doesn't work as well. If you're including the Vuejs project in with another project with other programming languages that follow a different or same src dir causes problems. This feature will separate the vue code from the other project code (js/PHP/C#/Go/python/etc.)

Currently if you want to modify the vue src directory you need to modify vue.config.js:

1.  change the alias @ path.
```
configureWebpack: {
        resolve: {
            alias: {
            '@': path.join(__dirname, vueSrc)
            }
        }
    }
```

2. Use the pages option to change their path:

```
pages: {
        app_main: {
            entry: path.join(vueSrc, 'main.ts'),
            template: path.join(vueSrc, 'public/index.html'),
            filename: 'index.html',
            title: 'Index'
        }
}
```

There might be other locations or properties that need to be updated with the change in src directory, but these 2 are the ones I've noticed so far.

There is the public directory that should also be considered. I assume it would also be moved into the specified Vue src directory or made into a separate web.config.js option.

If this is already doable without all the customization mentioned above, let me know.

### What does the proposed API look like?

1. Add an option to vue.config.js called `src` or `vueSrc` or `vueDir` or `srcDir` or `vueSrcDir` or `sourceDir` or whatever is deemed appropriate.
2. Either assume the public folder is also in the new vueSrc path or create a separate publicSrc (or publicDir) option in vue.config.js.

The value would be starting the current directory so you don't need to include the whole path.

Example: if you wanted to change the src vue directory and files from src to webclient/vue.

[...]

Issue #3040, opened 2018-11-29
Add prop in vue.config.js to change vue src directory


cthames:
### What problem does this feature solve?
Having the Vue src files and directories in the /src directory works for simple projects, but for more complex projects it doesn't work as well. If you're including the Vuejs project in with another project with other programming languages that follow a different or same src dir causes problems. This feature will separate the vue code from the other project code (js/PHP/C#/Go/python/etc.)

Currently if you want to modify the vue src directory you need to modify `vue.config.js`:

1. change the alias `@` path.
```
configureWebpack: {
        resolve: {
            alias: {
            '@': path.join(__dirname, vueSrc)
            }
        }
    }
```

2. Use the [pages](https://cli.vuejs.org/config/#pages) option to change their path:
```
pages: {
        app_main: {
            entry: path.join(vueSrc, 'main.ts'),
            template: path.join(vueSrc, 'public/index.html'),
            filename: 'index.html',
            title: 'Index'
        }
}
```

There might be other locations or properties that need to be updated with the change in `src` directory, but these 2 are the ones I've noticed so far.

There is the `public` directory that should also be considered. I assume it would also be moved into the specified Vue `src` directory or made into a separate `web.config.js` option.

If this is already doable without all the customization mentioned above, let me know.

### What does the proposed API look like?
1. Add an option to `vue.config.js` called `src` or `vueSrc` or `vueDir` or `srcDir` or `vueSrcDir` or whatever is deemed appropriate.
2. Either assume the `public` folder is also in the new `vueSrc` path or create a separate `publicSrc` (or `publicDir`) option in `vue.config.js`.

The value would be starting the current directory so you don't need to include the whole path.

**Example:** if you wanted to change the src vue directory and files from `src` to `webclient/vue`. 
> Note: The developer would manually create the new `webclient` directory in the project root. Then rename the `src` directory to `vue` and move it into the `webclient` folder. The `public` directory would also be moved into `webclient/vue`.

**vue.config.js**
```
module.exports = {
    vueSrcDir: 'webclient/vue', // Default: /src
    publicDir: 'webclient/vue' // Default: /public
}
```

> This is NOT something that would make cli-service / webpack move the files automatically.

The `vue-cli` would then change the vue src path where used by the cli-service.

The 3 places I've seen so far are:

1. resolve: { alias: { '@': pathToVueSrc } }
2. pages / entry / template
3. The `public` directory would also be located in the `pathToVueSrc` or another `vue.config.js` option.

[...]

Issue #2436, opened 2018-09-05
Add a configuration to specify "source" public folder.


cobysy:
### What problem does this feature solve?
Presently `npm run build` will copy all files in the folder `./public` (relative to the project folder) to the destination specified in `outputDir`. Which is fine if you stick to the default scaffolded folder structure (created by `vue create [project_name]`).

However in my case my entire vue project resides in slightly quirky folder structure like this (the ```public``` folder is in a 'sub folder' called ```vue-project```). The 'sub folder' sort of 'encapsulates' the vue project), except for the babel, tsconfig.json, vue.config.js, etc. which all still reside in the root of the 'project folder.'

```.
├── data
├── vue-project
│   ├── assets
│   ├── components
│   ├── dist
│   ├── public
│   └── src
├── lib-src
└── lib-tests
```

The code which copies the 'public' folder assets 
```
    // copy static assets in public/
    const publicDir = api.resolve('public')
    if (!isLegacyBundle && fs.existsSync(publicDir)) {
      webpackConfig
        .plugin('copy')
         ....
```

```api.resolve('public')``` resolves to either the `[project folder]/public` or `[environment variable VUE_CLI_CONTEXT]/public`.

Unfortunately you can't workaround it by executing ```cd vue-project && npm run build``` nor by ```VUE_CLI_CONTEXT=vue-project npm run build``` since other parts of the cli rely on the context being correct. 

EDIT:
A workaround is possible through `vue.config.js` using `CopyWebpackPlugin`, it's less pretty through:
```
    outputDir: 'example/dist',
    pages: {
        index: {
            entry: ...,
            template: 'example/public/index.html',
            filename: 'index.html',
        },
    },
    ...
    configureWebpack: config => {
        config.plugins.push(new CopyWebpackPlugin([{
            from: 'example/public/',
            to: '.',
            ignore: ['index.html', '.DS_Store']
        }]));
    }
```

### What does the proposed API look like?
Add a setting called `sourcePublicDir` in `vue.config.js` which value is resolved in places where `api.resolve('public')` (and `api.resolve('public/index.html')` etc.) are called.

**Comment by LinusBorg** [comment]
This is one of the sitations where i'm unsure what makes sense to add as *another* config option and what should be seen as a usecase for `chainWebpack` / `configureWebpack`.

Instead of pushing a new plugin, you could adjust the existing one:

```
chainWebpack: config => {
  config.plugin('copy').tap(([options]) => {
    options[0].from = 'example/public'
[...]

PR #3836, opened 2019-04-17
feat: support specifying directory for static files and template sources via a new publicDir option

kbtzr:
Adds a `publicDir` option to specify an alternative name or path to the public directory. Doc entry added and wording adjusted is some places to avoid confusion from the deterministic mentions to `public/`.

Tested locally both single and multi-page mode. If everything went alright I plan to work at #2697 later on.

### DIFF
--- a/docs/config/README.md
+++ b/docs/config/README.md
@@ -62,6 +62,18 @@ Deprecated since Vue CLI 3.3, please use [`publicPath`](#publicPath) instead.
   }
   ```
 
+### publicDir
+
+- Type: `string`
+- Default: `'public'`
+
+  If present, the contents of this directory will be copied to `outputDir` to be served as public static files.
+
+  ::: tip
+  Vue CLI tries to find template sources in this directory by default, it checks for `<publicDir>/index.html` or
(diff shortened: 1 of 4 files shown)


sodatea:
Thank you for your enthusiastic help but we have decided not to add this option. See more discussions here: https://github.com/vuejs/vue-cli/issues/2436#issuecomment-457961734 https://github.com/vuejs/vue-cli/issues/3040#issuecomment-443460624 https://github.com/vuejs/vue-cli/issues/3061#issuecomment-443856018

kbtzr:
Thank you for gathering these comments, now I not only understand the decision but also agree with it!

I'm considering making those tweaks via cli plugin, but I couldn't find information regarding plugin interoperability. Could you help me get started and point what could I do to ensure that my plugin is the last one extending the webpack configuration?

kbtzr:
@sodatea ping

sodatea:
Currently there's no way to do so.
Though, we do have plans to add a plugin ordering mechanism. We just haven't figured out a good API interface…

PR #3852, opened 2019-04-19
feat!: require ESLint as a peer dependency

sodatea:
BREAKING CHANGE:
Users will have to explicitly list their required ESLint version in the
project's package.json.

The major concern behind this change is that ESLint release schedule is
never aligned with Vue CLI's.
See https://eslint.org/blog/2019/04/eslint-v6.0.0-alpha.0-released
So even if we upgraded our built-in ESLint version to v5, we'll see a
new major release of ESLint in the coming months. Meanwhile the built-in
ESLint version can't be easily upgraded again without bumping Vue CLI's
major version number. This could become a maintenance nightmare.


sodatea:
Note:
If I recall correctly, the `babel-eslint` dependency in `@vue/cli-plugin-eslint` is only used for support of dynamic import parsing in ESLint.

https://github.com/babel/babel-eslint/pull/711/
Bad news: `babel-eslint` v11 requires a peer dependency of `@babel/core`, which, will definitely confuses those who don't need babel support in their projects.

Good news:
ESLint does not support dynamic import only because it is still a stage-3 proposal.
According to the TC39 meeting notes in March https://github.com/rwaldron/tc39-notes/blob/master/es10/2019-03/mar-28.md#dynamic-import-for-stage-4-in-june
The proposal is likely to move to stage 4 in June. Hopefully ESLint will follow up soon and then we can get rid of this extraneous dependency (Well, will need to move it to `@vue/cli-plugin-babel`, actually).

PR #3864, opened 2019-04-22
feat: allow `parallel` option to be an integer

### LINKED ISSUE #3850: Parallelism: set used cpu cores manually
Dockerized services limit the amount of CPU cores that can be used and, the service itself does not report correctly how many you can use from the actual total amount.

Example: https://github.com/vuejs/vue-cli/issues/2208

Global config file has the property `parallel` as a boolean. It could be just a number, the number of cores that are gonna be used, being 0 as parallelism deactivated. Or just some other way that allows us to set manually the amount of cores.

sodatea:
closes #3850

### DIFF
--- a/packages/@vue/cli-plugin-babel/index.js
+++ b/packages/@vue/cli-plugin-babel/index.js
@@ -59,9 +59,13 @@ module.exports = (api, options) => {
           .end()
 
     if (useThreads) {
-      jsRule
+      const threadLoaderConfig = jsRule
         .use('thread-loader')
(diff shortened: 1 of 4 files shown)


DRoet:
This doesn't seem to apply the actual worker/core count for the `thread-loader` when an integer is given, no?

JFGHT:
No, it doesn't. A bit more of work is required for that. Right now is only taking the integer and activating/deactivating the parallelism.

Issue #2024, opened 2018-08-01
--name must contain a hyphen when building a single web component.


crazyNoah:
IMPORTANT: Please use the following link to create a new issue:

https://new-issue.vuejs.org/?repo=vuejs/vue-cli

If your issue was not created using the app above, it will be closed immediately.

中文用户请注意：
请使用上面的链接来创建新的 issue。如果不是用上述工具创建的 issue 会被自动关闭。
build with component methods always come out this bug.
using vue-cli 3.0+ 
build with vue ui

bivainis:
In case someone stumbles on this, the error is received when Vue CLI assumes you're building a single file component, and thus requires you to hyphenate component name:
### Bad
`npx vue-cli-service build --target wc --name mycomponent './src/components/Button.vue'`

### Good
`npx vue-cli-service build --target wc --name my-component './src/components/Button.vue'`

Personally, I was receiving this error when I forgot to wrap my `[entry]` path in quotes (which was not that apparent when the command was in `package.json` scripts), when creating a bundle that registers multiple components. Which confused me, because I was sure I don't need to hyphenate the name in this case. But what probably happened was that vue-cli discared entry value as invalid and then just took first `.vue` file from `components` dir as entry, thus assuming it was supposed to be a single file component.

### Error (notice missing quotes)
`npx vue-cli-service build --target wc --name companyname ./src/components/*.vue`

### Success
`npx vue-cli-service build --target wc --name companyname './src/components/*.vue'`

PR #3886, opened 2019-04-25
Fix wc entry to allow multiple file patterns

manico:
This PR allows `globby` to get files from multiple different directories by using comma separated list of file patterns. This way we can also list specific components.

Examples:

`vue-cli-service build --target wc --name myprefix src/components/core*.vue,src/components/basic*.vue`

`vue-cli-service build --target wc --name myprefix src/components/myFirst.vue,src/components/mySecond.vue`

### DIFF
--- a/packages/@vue/cli-service/lib/commands/build/resolveWcConfig.js
+++ b/packages/@vue/cli-service/lib/commands/build/resolveWcConfig.js
@@ -15,7 +15,7 @@ module.exports = (api, { target, entry, name }) => {
   const isAsync = /async/.test(target)
 
   // generate dynamic entry based on glob files
-  const resolvedFiles = require('globby').sync([entry], { cwd: api.resolve('.') })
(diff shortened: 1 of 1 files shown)


bytedance-misaka:
Hey it's a major enhancement why no body merge this?

theboredfish:
any particular reason this wasn't also applied to 'lib' components?
need a way of independently compiling so that, if one failed, the rest would continue, so chunks and a common entry file isn't really an option for me i don't think :-/

YamiOdymel:
Leaving a note here, remember to use the `'` single quote to quote your path to get everything right.

Error (notice missing quotes)
`npx vue-cli-service build --target wc --name companyname ./src/components/*.vue`

Success
`npx vue-cli-service build --target wc --name companyname './src/components/*.vue'`

Source: [#2024 --name must contain a hyphen when building a single web component.](https://github.com/vuejs/vue-cli/issues/2024)

PR #3888, opened 2019-04-25
docs: add deployment instructions for Heroku

pinodex:
This method uses Express to serve the `dist` directory to Heroku's load balancer.


pinodex:
Will commit again following the naming conventions

PR #3897, opened 2019-04-26
feat(cli): Generator support async

xierenyuan:
In some cases, asynchronous operations in the generator will cause the build to fail, so I hope that the generator can support async

demo

```js
const packageJson = require('package-json')

const pkgConfig = {
  registryUrl: 'todo'
}

module.exports = async (api, options, rootOptions) => {
   const pkg1 = await packageJson('todo pkg', pkgConfig)
   api.render('./template', rootOptions)
   
    // todo...

    api.extendPackage({
	devDependencies: {
	     'todo pkg': pkg1.version
	}
    })
}
```


packages/@vue/cli/lib/Generator.js
```diff
@@ -97,17 +97,31 @@ module.exports = class Generator {
...
-    plugins.forEach(({ id, apply, options }) => {
-      const api = new GeneratorAPI(id, this, options, rootOptions)
-      apply(api, options, rootOptions, invoking)
+
+    this.rootOptions = rootOptions
+  }
+
+  initPlugins () {
+    const { rootOptions, invoking } = this
+    return new Promise((resolve, reject) => {
```
Akryum: Maybe use `async` with a `for` loop?

rikbrowning: Can we not do something like:
```suggestion
    return Promise.all(this.plugins.map(({ id, apply, options }) => {
        const api = new GeneratorAPI(id, this, options, rootOptions)
        return apply(api, options, rootOptions, invoking)
      }))
```

PR #3909, opened 2019-04-29
chore: prefer .tsx? files over .jsx? file extensions

### LINKED ISSUE #3898: .ts files should be preferred over .js files
### Version
3.6.0

### Reproduction link
[https://github.com/prigaux/vue-cli-prefer-ts-over-js](https://github.com/prigaux/vue-cli-prefer-ts-over-js)

### Steps to reproduce
Have both `foo.js` and `foo.ts` files in the same directory. Use `import ... from './foo'`

### What is expected?
`foo.ts` should be used

`foo.js` is used instead

---
This is an issue when foo.ts is shared between server (tsc generates foo.js) and client (vue).

#### Suggested fix

ensure '.ts' and '.tsx' are added in front of `config.resolve.extensions` in file `@vue/cli-plugin-typescript/index.js`

#### Workaround
add this to vue.config.js

```
module.exports = {
  ...
  chainWebpack: config => {
    config.resolve.extensions.prepend('.ts')
  },
}
```

LinusBorg:
When using Typescript, the webpack resolve options should prefer ts? file extensions over js? ones.

Strictly speaking, this is a breaking change, so we can ship this with v4 I think.

close #3898

### DIFF
--- a/packages/@vue/cli-plugin-typescript/index.js
+++ b/packages/@vue/cli-plugin-typescript/index.js
@@ -15,7 +15,8 @@ module.exports = (api, options) => {
 
     config.resolve
       .extensions
-        .merge(['.ts', '.tsx'])
+        .prepend('.ts')
+        .prepend('.tsx')
 
     const tsRule = config.module.rule('ts').test(/\.ts$/)
     const tsxRule = config.module.rule('tsx').test(/\.tsx$/)


LinusBorg:
@sodatea I think this is a small optimization that we can still add to v4, no?

sodatea:
Yeah I think so

PR #3921, opened 2019-04-30
chore!: default to dart sass for `sass` option of `cssPreprocessor`

sodatea:
BREAKING CHANGE:
Preset generated before v3.4.0 may contain a `"cssPreprocessor": "sass"`
field. It now means dart-sass rather than node-sass.

For generators, `rootOptions.cssPreprocessor === 'sass'` now also means dart sass.


jinweijie:
nice, finally get rid of node-sass. :D

Issue #212, opened 2016-10-26
How to update vue-cli?


xrr2016:

egoist:
Reinstall `npm i -g vue-cli` will do the favor.

CharlesOkwuagwu:
@egoist how do you update a project built with a lower version of the cli?

egoist:
@CharlesOkwuagwu you mean templates or `vue build`

CharlesOkwuagwu:
I mean a project already set up with an older version of vue, how do we get the project updated via the cli

LinusBorg:
You don't, currently. The templates+CLI don't support any kind of update process. Depending on what you changed in the template you are currently using, it might be easier to generate a new project, copy over `/src` & `/static`, and re-apply any changes you made to the webpack config & package.json scripts.

posva:
I'd add to what @LinusBorg said that git is your friend 😄

LinusBorg:
...and maybe that we are thinking about how to improve upgrades in the future, but that's nothing that will happen very soon.

iamdubx:
Is it happening a year ago?

LinusBorg:
Yes, actually. You can find an Alpha Version of vue-cli 3 in the `next` branch.

bitbytebit1:
@LinusBorg, so can we update our projects now? or do we still have to do it the "manual way"

LinusBorg:
Nothing changed for old projects, but new projects created with vue-cli 3 can be updated by simply updating the dependencies.

muntac:
Note that if you're already on 3.0 and are trying to upgrade, `npm i -g vue-cli` will install 2.* (2.9.6 in my case). Instead, try `npm i -g @vue-cli`.

rhkina:
Hi, @muntac, I suppose the command is npm i -g @vue/cli... I've tried this in Ubuntu, but when I type vue -V, the response is 2.9.3

rhkina:
Hi @muntac,  I suppose the command is `npm i -g @vue/cli`... But after installing it I am getting 2.9.3 when input `vue --version` command...

Akryum:
This means you have the old `vue-cli` package installed. You need to uninstall it.

mak2014:
This worked for me:  

Uninstall first: 

`npm uninstall -g vue-cli`

Install:

`npm install -g @vue/cli`

ZadaWu:
seems we need to install the latest, I worked when I add @latest:

[...]

PR #3926, opened 2019-05-01
chore: better upgrade messages

phanan:
Adding an instruction into our "Update available" message, similar to how npm and yarn handle the case.

Before:

After:

Related: #212.


packages/@vue/cli/lib/util/clearConsole.js
```diff
@@ -16,12 +16,18 @@ exports.generateTitle = async function (checkUpdate) {
...
     if (process.env.VUE_CLI_API_MODE) {
-      title += chalk.green(` 🌟️ Update available: ${latest}`)
+      title += chalk.green(` 🌟️ New version available: ${latest}`)
     } else {
-      title += chalk.green(`
-┌────────────────────${`─`.repeat(latest.length)}──┐
-│  Update available: ${latest}  │
-└────────────────────${`─`.repeat(latest.length)}──┘`)
+      const upgradeMessage = `New version available ${chalk.magenta(current)} → ${chalk.green(latest)}\n` +
+        `Run ${chalk.yellow(`npm i -g ${require('../../package.json').name}`)} to update!`
```
Akryum: It should take into account if it's installed using `npm`, `yarn` or `pnpm`, otherwise it will actually cause trouble.

phanan: Good point. How do we differentiate the two cases though (I don't think we should care for `pnpm`, as it's not even listed on the official docs).

phanan: Hmm, actually I think something like "…or yarn global add @vue/cli` will do just fine.

Akryum: > as it's not even listed on the official docs

It's officially supported though.

> Hmm, actually I think something like "…or yarn global add @vue/cli` will do just fine.

I think having a tiny help message that doesn't really help (things might go wrong if the user picks an incorrect "choice") is bad. We may display a link to a section in the docs explaining how to upgrade vue-cli though (I think that's the way `now` is doing it).

phanan: Do we have such a section?

pksunkara: @phanan You can now use this. https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli/lib/util/ProjectPackageManager.js#L66. Let's get this merged once you fix this issue.

sodatea: @pksunkara The result may be incorrect. Better to test `__dirname` against `npm config get prefix` / `yarn global dir`

sodatea:
Hi, I've implemented the installation detection feature on my local clone.
But I can't push it directly to this branch because of permission issues. Could you edit this PR to allow edit from maintainers?	https://help.github.com/en/articles/allowing-changes-to-a-pull-request-branch-created-from-a-fork#enabling-repository-maintainer-permissions-on-existing-pull-requests

PR #3930, opened 2019-05-02
fix(babel): ignore core-js sources

Akryum:
See https://github.com/babel/babel/issues/8731

### DIFF
--- a/packages/@vue/babel-preset-app/index.js
+++ b/packages/@vue/babel-preset-app/index.js
@@ -184,6 +184,7 @@ module.exports = (context, options = {}) => {
 
   return {
     presets,
-    plugins
+    plugins,
+    ignore: [/\/core-js/]
   }
 }


sodatea:
The tests are failing.
Honestly I've never encountered such error before…

Akryum:
I encountered it while transpiling a linked npm package when importing a raw `.vue` component which has a prop with `type: Number`, which triggers the core-js `NumberConstructor` polyfill.

Akryum:
> The tests are failing.

Looks like it's not allowed inside presets 😿, maybe `exclude` would work...

Akryum:
Or maybe we can just add it to the generated babel config?

sodatea:
Hmmmm…
I think the actual problem here is that linked modules does not work well with webpack `exclude`. We already excluded `node_modules` by default,

Akryum:
> We already excluded node_modules by default,

The issue arises when manually telling vue-cli to actually transpile the components package

sodatea:
oh then it makes sense to fix the exclude pattern

PR #3939, opened 2019-05-04
feat(HtmlPwaPlugin): adds 'manifestCrossorigin' option

### LINKED ISSUE #3933: [@vue/pwa] add option to configure crossorigin tag in manifest link in generated index.html
I work with an apache server with sspi auth, so I've been having problems with the manifest.json include returning 401, with the text:

    This server could not verify that you are authorized to access the URL "[...]/manifest.json". You either supplied the wrong credentials (e.g. bad password), or you browser doesn't understand how to supply the credentials required.

Adding `crossorigin=use-credentials` to the manifest link tag fixes it, but there is no automatic generation, it's something I have to do every time I build a new version. This is something that could be a plugin setting

add `pwa.crossorigin` seeting option to vue.config.js file to specify if crossorigin tag should be added and if so, it's value

BigsonLvrocha:
close #3933

### DIFF
--- a/packages/@vue/cli-plugin-pwa/lib/HtmlPwaPlugin.js
+++ b/packages/@vue/cli-plugin-pwa/lib/HtmlPwaPlugin.js
@@ -8,25 +8,26 @@ const defaults = {
   appleMobileWebAppStatusBarStyle: 'default',
   assetsVersion: '',
   manifestPath: 'manifest.json',
-  manifestOptions: {}
+  manifestOptions: {},
+  manifestCrossorigin: undefined
 }
 
 const defaultManifest = {
   icons: [
     {
-      "src": "./img/icons/android-chrome-192x192.png",
-      "sizes": "192x192",
-      "type": "image/png"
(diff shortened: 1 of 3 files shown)


bluebrown:
Thanks for implementing this! I am little confused how to use it.

BigsonLvrocha:
In some specific cases the PWA is served behind an authenticated proxy (corporate network), so the request for the manifest.json file might return `409 - Forbidden` error. The solution to this is to add the attribute `crossorigin=use-credentials` in the include tag. [Info about it](https://developer.mozilla.org/pt-PT/docs/Web/HTML/CORS_settings_attributes). Currently there is no setup for that in the plugin, so I added that setting. So, if you need this, you need to setup this in your vue.config.js as below
```js
module.exports = {
  pwa: {
    manifestCrossorigin: 'use-credentials',
  },
};
```
And it will add the attribute to the tag in the generated index.html file

bluebrown:
Thanks so much for taking your time to  clarify. I will see if I can find a good place in the documentation, and update the info via pull request. (If its not already there.).

Issue #3941, opened 2019-05-05
checkyarn fails on Solus


omarcresp:
### Version
3.7.0

### Environment info
```
Environment Info:

  System:
    OS: Linux 5.0 Solus 4.0
    CPU: (4) x64 Intel(R) Celeron(R) CPU  N2930  @ 1.83GHz
  Binaries:
    Node: 10.15.3 - ~/.nvm/versions/node/v10.15.3/bin/node
    Yarn: 1.16.0 - /usr/bin/yarn
    npm: 6.9.0 - ~/.nvm/versions/node/v10.15.3/bin/npm
  Browsers:
    Chrome: Not Found
    Firefox: 66.0.3
  npmGlobalPackages:
    @vue/cli: 3.7.0
```

### Steps to reproduce
- Create a vue project with yarn
- Use vue add anyPackage

### What is expected?
That vue detect yarn without problem during the plugin installation

### What is actually happening?
vue add quasar@beta

📦  Installing vue-cli-plugin-quasar@beta...

 ERROR  Error: The project seems to require yarn but it's not installed.
Error: The project seems to require yarn but it's not installed.
    at checkYarn (/home/jackcres/.nvm/versions/node/v10.15.3/lib/node_modules/@vue/cli/node_modules/@vue/cli-shared-utils/lib/env.js:46:43)
    at exports.hasProjectYarn (/home/jackcres/.nvm/versions/node/v10.15.3/lib/node_modules/@vue/cli/node_modules/@vue/cli-shared-utils/lib/env.js:42:10)
    at add (/home/jackcres/.nvm/versions/node/v10.15.3/lib/node_modules/@vue/cli/lib/add.js:30:59)
    at module.exports.args (/home/jackcres/.nvm/versions/node/v10.15.3/lib/node_modules/@vue/cli/lib/add.js:66:10)
    at Command.program.command.description.option.allowUnknownOption.action (/home/jackcres/.nvm/versions/node/v10.15.3/lib/node_modules/@vue/cli/bin/vue.js:86:26)
    at Command.listener (/home/jackcres/.nvm/versions/node/v10.15.3/lib/node_modules/@vue/cli/node_modules/commander/index.js:315:8)
    at Command.emit (events.js:189:13)
    at Command.parseArgs (/home/jackcres/.nvm/versions/node/v10.15.3/lib/node_modules/@vue/cli/node_modules/commander/index.js:651:12)
[...]

PR #3942, opened 2019-05-05
Detect yarn in Solus os

omarcresp:
https://github.com/vuejs/vue-cli/issues/3941

### DIFF
--- a/packages/@vue/cli-shared-utils/lib/env.js
+++ b/packages/@vue/cli-shared-utils/lib/env.js
@@ -27,15 +27,13 @@ exports.hasYarn = () => {
     execSync('yarnpkg --version', { stdio: 'ignore' })
     return (_hasYarn = true)
   } catch (e) {
-    _hasYarn = false
   }
   try {
     execSync('yarn --version', { stdio: 'ignore' })
     return (_hasYarn = true)
   } catch (e) {
-    _hasYarn = false
+    return (_hasYarn = false)
   }
-  return _hasYarn
 }
 
 exports.hasProjectYarn = (cwd) => {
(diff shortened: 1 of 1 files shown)


packages/@vue/cli-shared-utils/lib/env.js
```diff
@@ -27,8 +27,15 @@ exports.hasYarn = () => {
     execSync('yarnpkg --version', { stdio: 'ignore' })
     return (_hasYarn = true)
   } catch (e) {
-    return (_hasYarn = false)
+    _hasYarn = false
```
pksunkara:
```suggestion
```

packages/@vue/cli-shared-utils/lib/env.js
```diff
@@ -27,8 +27,15 @@ exports.hasYarn = () => {
...
   }
+  try {
+    execSync('yarn --version', { stdio: 'ignore' })
+    return (_hasYarn = true)
+  } catch (e) {
+    _hasYarn = false
+  }
+  return _hasYarn
```
pksunkara:
```suggestion
```

packages/@vue/cli-shared-utils/lib/env.js
```diff
@@ -27,8 +27,15 @@ exports.hasYarn = () => {
...
     return (_hasYarn = true)
   } catch (e) {
-    return (_hasYarn = false)
+    _hasYarn = false
   }
+  try {
+    execSync('yarn --version', { stdio: 'ignore' })
+    return (_hasYarn = true)
+  } catch (e) {
+    _hasYarn = false
```
pksunkara:
```suggestion
    return (_hasYarn = false)
```

pksunkara:
The main reason for using `yarnpkg` is that most of OS's give preference to Hadoop YARN when you call `yarn`

sodatea:
Hadoop YARN will throw for `yarn --version` (it only supports `yarn --version`) so won't pass our check.

Since we run `yarn` rather than `yarnpkg` everywhere else, I think it makes sense to only check `yarn`.

PR #3955, opened 2019-05-07
Feat rename project

### LINKED ISSUE #3793: Option to nickname projects in the vue-ui Dashboard
I have a whole list of SPA's within parent applications that are all called "frontend", so I have no idea what I'm looking at when I pull up the ui dashboard. I just accidentally made a bunch of changes to one that I wasn't even intending to work on.

It would be great if Vue UI projects could be "nicknamed" by something besides the last dir in path.

// shrug

I'm just an end user, I have no idea what the API looks like currently.

Akryum:
closes #3793


packages/@vue/cli-ui/src/components/app/ProjectQuickDropdown.vue
```diff
@@ -1,84 +1,98 @@
...
+
+        <VueDropdownButton
+          :label="$t('org.vue.components.project-select-list-item.tooltips.open-in-editor')"
+          icon-left="open_in_browser"
+          @click="openInEditor(projectCurrent)"
+        />
+
+        <VueDropdownButton
+          :label="$t('org.vue.components.project-rename.title')"
+          icon-left="local_offer"
```
sodatea: This icon typically indicates "label" rather than "rename" IMO.
Actually I failed to recognize it in the first glance…

Akryum: Maybe a pen then? https://material.io/tools/icons/?icon=create&style=baseline

sodatea: Looks good

Akryum: Could you replace the two `local_offer` with `edit` then? Thanks!

PeterDKC:
🤙 That's a sanity saver. Thanks folks!

PR #3966, opened 2019-05-08
fix(cli-ui): fix detectLanguage function

maple3142:
This is because [vue-cli-locales/locales](https://unpkg.com/vue-cli-locales@3.7.0/locales/) use `_` to separate language and region name, while some browser use `-` to do so.
image


LizhenQiao:
Same problem happens to me now. Vue/cli@4.5.9  .  What should I do to run the vue ui?

LizhenQiao:
[
image
](url)

Issue #3726, opened 2019-03-29
Customize or disable serve console message


csicky:
### What problem does this feature solve?
I would like to be able to see only if there is a problem, if it's green, yellow or red. I don't want to give so much screen to the console, but in order to see what I want I have to see also the entire message which I know already:

```
 DONE  Compiled successfully in 24499ms      
  App running at:
  - Local:   http://localhost:8080/
  - Network: http://192.168.100.10:8080/

  Note that the development build is not optimized.
  To create a production build, run npm run build.
```

### What does the proposed API look like?
I need a way to have the console say only:

```
 DONE  Compiled successfully in 24499ms  
```

### Not duplicate of https://github.com/vuejs/vue-cli/issues/2115, that issue was a different topic.

LinusBorg:
...but that message only appears on first compile?

csicky:
Yes, the "development build not optimized" message appears only once, but still the "app running at" is taking up vertical space. If I try to make the console height to take only one line, I don't see anymore if it's green red or yellow. I actually only care about the colors.

![image](https://user-images.githubusercontent.com/3931907/55241979-f862f780-5244-11e9-97d3-af879b83711d.png)


What I do to is I go to serve.js in `node_modules` folder and comment out the noisy `console.log` calls. This is problematic when I update packages because I have to go and do it again.

Thank you for your help.

PR #3970, opened 2019-05-09
Added a parameter to disable serve console message (#3726)

matteotomasoni:
I added a new parameter `noinstructions` to the serve command.
This parameter disables the boring instructions logged on the console on every compilation, so that you can have a very small console open with just the important stuff:
`DONE  Compiled successfully in XXms`

I think it fixes issue #3726


csicky:
Thank you @matteotomasoni
I hope this gets accepted.

PR #3988, opened 2019-05-13
feat: add --extra-env to load more env configs.

novbing:
I wanna add --extra-env  to load more env configs. So, it can support .env files more freely, maybe.

Exp:
.env files:
.env.development.main
.env.extra1
.env.extra2

cli:
vue-cli-service serve --mode=development.main --extra-env=extra1+extra2

### DIFF
--- a/packages/@vue/cli-service/lib/Service.js
+++ b/packages/@vue/cli-service/lib/Service.js
@@ -55,7 +55,7 @@ module.exports = class Service {
     }
   }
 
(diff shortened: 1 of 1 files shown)


sodatea:
I don't feel like adding this option.
I think it makes the usage of `.env` files too obscure. The inter-dependency between each `.env` file is not clear. Would become a maintenance burden in the future.

PR #4000, opened 2019-05-15
feat: implement version-related APIs for `GeneratorAPI`

sodatea:
To discuss: is the name `serviceVersion` / `assertServiceVersion` ideal? Any other naming ideas?


novbing:
> To discuss: is the name `serviceVersion` / `assertServiceVersion` ideal? Any other naming ideas?

`serviceVersion` / `exactServiceVersion`
`serviceVersion` / `lockedServiceVersion`
`serviceVersion` / `workServiceVersion`
...

sodatea:
Well, I think the `assert` part is accurate. But is `serviceVersion` clear enough to indicate that it is actually a per-project version number, in contrast to the `version` field as a global version number?

novbing:
currentVersion
localVersion
?

sodatea:
`currentVersion` does not distinguish it from the global `version` field.

`localVersion` may be a good alternative but I'm not sure…
While `@vue/cli-service` is local, it is not depended on by `GeneratorAPI`…

pksunkara:
I would name them as following:

* `cliVersion`
* `requireCliVersion`
* `cliServiceVersion`
* `requireCliServiceVersion`

IMO easier to understand what they mean

sodatea:
🤔prefer `assert` because it's in line with `PluginAPI` and babel's API naming.

So,

- `cliVersion`
- `assertCliVersion`
- `cliServiceVersion`
- `assertCliServiceVersion`

WDYT?

pksunkara:
Sounds good. We should probably backport this to v3 and let `assert*` just return `false`.

sodatea:
Yeah, all non-breaking changes would be backported to v3.

PR #4034, opened 2019-05-21
fix: fix dev mode compatibility with workers

### LINKED ISSUE #3539: Compiling with  `vue-cli-service build --mode development --modern` produces output that fails to run
### Version
3.4.1

### Environment info
```
MacOS/Chrome 72

FYI: The instructions on this page to run `vue info` does not work anymore.
```

### Steps to reproduce
Steps are so basic, no project is required.
1. Create a new project with `vue create hello-world`
2. Use all defaults when creating.
3. In new project directory, execute `vue-cli-service build --mode development --modern`
4. Try to run outputted vue app.
Result: Produced project will error with 
    Uncaught TypeError: Cannot read property 'webpackHotUpdate' of undefined
        at app.js:5
        at app.js:794

### What is expected?
The issue is that globalObject should be `self` not `this`

The dev version of the code is using 'this' which fails because `this` is not defined. It should be `self`.

A fix can clearly be seen in resolveLibConfig.js that probably just never made it into dev.js

---
`globalObject` in dev.js should probably have same fix as found in resolveLibConfig.js

See:
<https://github.com/vuejs/vue-cli/blob/94ff7bf33c4936be2e080a36ea54889bbac81fe2/packages/%40vue/cli-service/lib/config/dev.js#L16>
vs
<https://github.com/vuejs/vue-cli/blob/94ff7bf33c4936be2e080a36ea54889bbac81fe2/packages/%40vue/cli-service/lib/commands/build/resolveLibConfig.js#L105>

My fix to dev.js was:
```
      webpackConfig
        .output
          .globalObject(`(typeof self !== 'undefined' ? self : this)`)
```

sodatea:
fixes #3539

### DIFF
--- a/packages/@vue/cli-service/lib/config/dev.js
+++ b/packages/@vue/cli-service/lib/config/dev.js
@@ -11,9 +11,10 @@ module.exports = (api, options) => {
           .use(require('webpack/lib/HotModuleReplacementPlugin'))
 
       // https://github.com/webpack/webpack/issues/6642
+      // https://github.com/vuejs/vue-cli/issues/3539
       webpackConfig
         .output
-          .globalObject('this')
+          .globalObject(`(typeof self !== 'undefined' ? self : this)`)
 
       if (!process.env.VUE_CLI_TEST && options.devServer.progress !== false) {
         webpackConfig


alexander-elgin:
I met the same issue in my project built using vue-cli. What should I do to fix it?

PR #4070, opened 2019-05-28
fix: support custom Global Vue alias

shaydoc:
Adding support for customizing the Global Vue alias in WebComponent builds.
The reason for this is to avoid conflict with other components that may bundle Vue as part of their library. I am having this issue at present.

### DIFF
--- a/packages/@vue/cli-service/__tests__/buildWc.spec.js
+++ b/packages/@vue/cli-service/__tests__/buildWc.spec.js
@@ -89,6 +89,57 @@ test('build as single wc', async () => {
   expect(h1Text).toMatch('Welcome to Your Vue.js App')
 })
 
+test('build as wc with custom Vue alias', async () => {
+  const project = await create('build-wc-custom-alias', defaultPreset)
+
+  const { stdout } = await project.run(`vue-cli-service build --alias $MyVueAlias --target wc **/*.vue --name wc-custom-alias`)
+  expect(stdout).toMatch('Build complete.')
+
+  expect(project.has('dist/demo.html')).toBe(true)
+  expect(project.has('dist/wc-custom-alias.js')).toBe(true)
+  expect(project.has('dist/wc-custom-alias.min.js')).toBe(true)
+
+  const port = await portfinder.getPortPromise()
+  server = createServer({ root: path.join(project.dir, 'dist') })
+
+  await new Promise((resolve, reject) => {
+    server.listen(port, err => {
+      if (err) return reject(err)
+      resolve()
+    })
+  })
+
+  const launched = await launchPuppeteer(`http://localhost:${port}/demo.html`)
+  browser = launched.browser
+  page = launched.page
+
+  const customAlias = await page.evaluate(() => {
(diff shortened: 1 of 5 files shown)


shaydoc:
@linusborg I want externalize Vue as Vue$ or some other name to avoid a collision.

I have a bundled vue application that uses a another component that is bundled but also includes Vue.

When I set `window.Vue = Vue` in my main.js of my core application, it breaks the other code split component which had Vue bundled as part of its code.

I have to set this in my main.js because I bring in webcomponents built within Vue.

The reason I bring Vue Web components is specifically for use inside a tinymce editor.

Do you understand what I mean?

shaydoc:
Hey @LinusBorg 
So I am revisiting this because I really need it :)
I have made a change to add a command line alias operation, I have updated the tests to reflect this.

`  '--alias': "use with --target 'wc' or 'wc-async' to externalize Vue under a custom alias, e.g. $MyVue"`

Within demo-wc.html
```
<meta charset="utf-8">
<title><%- htmlWebpackPlugin.options.libName %> demo</title>
<script src="https://unpkg.com/vue"></script>
<% if  (htmlWebpackPlugin.options.alias !== 'Vue') { %>
<script>
    window[<%= JSON.stringify(htmlWebpackPlugin.options.alias) %>] = Vue;
</script>
<% } %>
<script src="./<%- htmlWebpackPlugin.options.libName %>.js"></script>

<% for (const comp of htmlWebpackPlugin.options.components) { %>
<<%= comp %>></<%= comp %>>
<% } %>
```

LinusBorg:
Hey @shaydoc 

Thanks for working on this and pushing it. I definitely understand the use case now as well, so thanks for that.

I'll try and review this before the end of the week. Want to take some time in order to think about possible implications.

LinusBorg:
Okay so I think the code looks fine technically. I'm still thinking about the use cases.

In your specific case, I'd like to understand how the 3rd party component, which has *not* externalized Vue, but instead comes with it's own version bundled inside, creates issues when a global Vue instance is present? If they use their internally bundle version, I don't see how their code could care about the globally available version.

Do they assign their bundled version to `window.Vue`?

I ask about this as I want to make sure this actually solves a use case that is not super-rare and caused by bad practices in some other place...

shaydoc:
Screenshot 2019-07-29 at 12 03 30

So in my app when I set `window.Vue = Vue`  I get this error from the 3rd party component

shaydoc:
@LinusBorg any further thoughts on this one ?

xcxooxl:
Actually I am building a library for my website which both are built in vue, And I'm experiencing the same error (Cannot redefine property.. $router), Would like an update on this also

shaydoc:
@LinusBorg any updates on when this might be able to be merged ?
How could i use this in the immediate term, can I npm install this branch globally ?

sodatea:
It's already possible with the following config in `vue.config.js`:
```js
module.exports = {
  chainWebpack: (config) => {
    config.externals({
      vue: 'global $MyVueAlias'
    })
  }
}
```

So no need to add a new flag.

PR #4080, opened 2019-05-30
V3 after install @vue/cli I got 'vue' is not recognized as an internal or external command, operable program or batch file.

govarthananve:
vue

and I got another error my friend system 
vue1

Please help to fix this issue


sodatea:
Please do not abuse pull requests. Next time, open an issue with our [issue helper](https://new-issue.vuejs.org/?repo=vuejs/vue-cli).

In your case, the problem is that the binary path that npm installs the global packages to (`C:\Users\gvenunathan\'~\.npm-packages'\`) is not in the System Variable Paths. It's not related to Vue CLI. See https://www.thewindowsclub.com/is-not-recognized-as-an-internal-or-external-command

In your friend's case, he/she needs to run the CMD program with Administrative privileges.

govarthananve:
yes thanks;)
 I just move the C:\Users\gvenunathan\'~\.npm-packages'\ vue.cmd and vue file into ..\AppData\Roaming\npm here now working fine

PR #4094, opened 2019-06-03
fix: set useTaobaoRegistry to true when user configured custom regist…

KuangPF:


sodatea:
Thanks for your help, but with the refactoring done in this PR: #4256 I think this logic is no longer necessary.

PR #4109, opened 2019-06-05
Upgrade fiber to support node@12

renfeng:
See https://github.com/laverdet/node-fibers/issues/409


sodatea:
Thanks for this PR. But I don't think this change makes any difference.
And since we may need to test on Node.js v8, it's better to continue allowing fibers@3 as a devDependency.

renfeng:
It breaks the build process for projects that depend on it! Can you read this, and reopen it? Thanks

sodatea:
Could you provide a reproduction?
I don't think it will break any build.

1. It's in `devDependencies`, won't be installed for those who depend on it.
2. The range `>= 3.1.1 <5.0.0` does resolve to `4.0.1` in environments that support fibers@4.x. See the lockfile: https://github.com/vuejs/vue-cli/blob/ab97f31a17282d1096060609b7e33bef22c0b21b/yarn.lock#L7018-L7023 Your PR won't make any difference here.

PR #4143, opened 2019-06-13
type error; sercet => secret

lee-zhi:
type error; sercet => secret


LinusBorg:
Values from .env files are stringified, you don't have to do it yourself.

PR #4171, opened 2019-06-20
Explain when .env files are loaded.

rimutaka:
It appears that .env files are only loaded at the start of `vue-cli-service` and are not watched for hot reloads. 
**I am actually not sure this is the case for everyone, but is so for me.**

I think it is worth noting because the expectation is that a file at the root of the project would be watched for hot reloads.


StefanFeederle:
Is it possible to watch for changes on .env files?

Issue #4051, opened 2019-05-24
Cannot read property 'android' of undefined


yusidi:
### Version
3.7.0

### Environment info
```
mac
```

### Steps to reproduce
    ['@vue/app', {
      polyfills: [
        // all polyfills list -> https://github.com/zloirock/core-js
        'es6.object.assign',
        'es6.array.find',
        'es6.array.iterator',
        'es6.function.name',
        'es7.object.values',
        'web.dom.iterable'
      ]
It  seem caused by'web.dom.iterable'

### What is expected?
build is running

### What is actually happening?
Cannot read property 'android' of undefined
---
    ['@vue/app', {
      polyfills: [
        // all polyfills list -> https://github.com/zloirock/core-js
        'es6.object.assign',
        'es6.array.find',
        'es6.array.iterator',
        'es6.function.name',
        'es7.object.values',
        'web.dom.iterable'
      ]
It  seem caused by'web.dom.iterable'

LinusBorg:
Hi  

Thanks for filing this issue! Could you please provide a small reporduction on jsfiddle.net so we can debug this without trying to replicate this ourselves?`

That would be very helpful, thanks!

yusidi:
vue create demo。
show tip ‘These dependencies were not found: core-js/modules/web.dom.iterable’，after the code in my project。
we add 'web.dom.iterable' in babel.config.js after that。
 vue-cli-service serve, throw error  'Cannot read property 'android' of undefined '
```
import Vue from 'vue'
import hljs from 'highlight.js/lib/highlight'

import 'highlight.js/styles/github-gist.css'

import shell from 'highlight.js/lib/languages/shell'
import css from 'highlight.js/lib/languages/css'
import markdown from 'highlight.js/lib/languages/markdown'
import js from 'highlight.js/lib/languages/javascript'
import json from 'highlight.js/lib/languages/json'
import xml from 'highlight.js/lib/languages/xml'

hljs.registerLanguage('css', css)
hljs.registerLanguage('markdown', markdown)
hljs.registerLanguage('javascript', js)
hljs.registerLanguage('json', json)
hljs.registerLanguage('shell', shell)
hljs.registerLanguage('xml', xml)

Vue.directive('highlight', (el) => {
  let blocks = el.querySelectorAll('pre code')
  blocks.forEach(block => {
    hljs.highlightBlock(block)
  })
})

```

johndab:
[...]

PR #4193, opened 2019-06-24
ci: running some test cases with a local registry to better reflect real-world use cases

sodatea:
For example:
https://github.com/vuejs/vue-cli/issues/4051#issuecomment-504298130
We'll have to set up a new project alongside the monorepo (rather than inside it) to test such cases

Also, test cases of `vue upgrade` (https://github.com/vuejs/vue-cli/pull/4090) need to be built upon this PR.


sodatea:
This seems more difficult than I thought…
I'll drop this for now, so as to focus more on v4 core features and migrations. Will revisit it later

Issue #2335, opened 2018-08-23
Make router and vuex proper packages


pksunkara:
### What problem does this feature solve?
There's a lot of careful manipulation being done in @vue/cli code because of the above plugins not being proper packages, which we could avoid. Since the implementation of #1202, the format of the above plugins is in a proper plugin format anyway. So, why don't we move them to their own packages in the repo? That way, we could remove all the specialty code written just for these packages.

### What does the proposed API look like?
Have 2 new packages @vue/cli-plugin-router, @vue/cli-plugin-vuex

sodatea:
I'm reluctant to add two new generator-only plugins.
Even if we do, we should not mess up users' `package.json`s; co-existence of `@vue/cli-plugin-vuex` and `vuex` would definitely confuse them.
So we'll still need to add some special logic to remove the plugins after generating the necessary files. Does that worth the refactoring?

pksunkara:
> So we'll still need to add some special logic to remove the plugins after generating the necessary files. Does that worth the refactoring?

I think yes, because right now the generation is not very consistent and also having these as internal options in `cli-service` generator is messing with other 3rd party plugins when they are doing [this](https://cli.vuejs.org/dev-guide/plugin-dev.html#editing-existing-templates)

> co-existence of @vue/cli-plugin-vuex and vuex would definitely confuse them.

I actually have no issue about this as an end-user if the cli-plugin-vuex has more helpful commands (like maybe generating a store module etc..)

sodatea:
> I actually have no issue about this as an end-user if the cli-plugin-vuex has more helpful commands (like maybe generating a store module etc..)

Yeah, definitely worth it if such features implemented.
It's been on my backlog for a while actually. Just haven't got the time to think about the design.

pksunkara:
I will work on this and send a PR for the initial packages. Then we can discuss on the features to be implemented. I am always up for a chat on discord if you need someone to brainstorm with.

sodatea:
😀Cool, much appreciated.

PR #4196, opened 2019-06-25
Make router a separate plugin

pksunkara:
Part of #2335. We remove the `router` option from root options and use it as a plugin as you can see [here](https://github.com/vuejs/vue-cli/pull/4196/files#diff-33aaf09ea1e85f41cdf5050ca1493dd5). [This](https://github.com/vuejs/vue-cli/pull/4196/files#diff-b236e80985559dfeb311568d7c39ff5e) is the actual App.vue template for router which now makes it more consistent over different invoking methods.


packages/@vue/cli-ui/ui-defaults/suggestions.js
```diff
@@ -73,16 +73,19 @@ async function install (api, id) {
...
   const context = api.getCwd()
 
   let error
 
   try {
-    await invoke.runGenerator(context, {
-      id: `core:${name}`,
-      apply: loadModule(`@vue/cli-service/generator/${name}`, context)
-    })
+    if (id === 'router') {
```
pksunkara: This `if` condition will exist until I make the `cli-plugin-vuex` PR.

pksunkara:
@LinusBorg @sodatea This is ready for review and merge.

docs/ru/guide/plugins-and-presets.md
```diff
@@ -112,15 +105,15 @@ vue add vuex
 ``` json
 {
   "useConfigFiles": true,
-  "router": true,
```
sodatea: Though we now have a standalone router plugin, the legacy preset format (`router` and `routerHistoryMode` option) can and should still be supported, so as to reduce the migration costs.

packages/@vue/cli-plugin-router/README.md
```diff
@@ -0,0 +1,9 @@
+# @vue/cli-plugin-router
+
+> router plugin for vue-cli
+
+## Installing in an Already Created Project
+
+``` sh
+vue add @vue/router
```
sodatea: This seems too verbose. I'm thinking about fixing the `add` command to check for `@vue/cli-plugin-foo` if a shorthand id is provided. I'll do it in a separate PR though.

packages/@vue/cli/lib/util/inferRootOptions.js
```diff
@@ -8,11 +8,6 @@ module.exports = function inferRootOptions (pkg) {
   // projectName
   rootOptions.projectName = pkg.name
 
-  // router
-  if ('vue-router' in deps) {
-    rootOptions.router = true
```
sodatea: We don't have to remove this. Router is now a generator-only plugin anyway. Checking for `vue-router` is fine and we need to support the legacy rootOptions for now.

docs/ru/guide/plugins-and-presets.md
pksunkara: I added support for legacy router option but have it removed from docs completely.

pksunkara:
This is now ready for another review.

packages/@vue/cli/lib/Creator.js
```diff
@@ -92,10 +92,16 @@ module.exports = class Creator extends EventEmitter {
...
       projectName: name
     }, preset)
+
     if (cliOptions.bare) {
       preset.plugins['@vue/cli-service'].bare = true
     }
 
+    // legacy support for router
+    if (preset.router) {
+      preset.plugins['@vue/cli-plugin-router'] = {}
```
sodatea: Need to support `routerHistoryMode` here.

packages/@vue/cli/lib/options.js
```diff
@@ -10,8 +10,6 @@ const rcPath = exports.rcPath = getRcPath('.vuerc')
 const presetSchema = createSchema(joi => joi.object().keys({
   bare: joi.boolean(),
   useConfigFiles: joi.boolean(),
-  router: joi.boolean(),
```
sodatea: Though not preferred, they should remain in the schema, otherwise the legacy preset format cannot pass the schema validation.

Note: you may leave a TODO: comment here, once `@hapi/joi` v16 released, we can update the schema to utilizing its `.warn` feature. https://github.com/hapijs/joi/commit/59e60aa851f674d63c9ac1bad4b0f59c0fa24033

sodatea:
As I tested out, the router plugin does not work well with `@vue/cli-service` 3.x (maybe due to the `REPLACE` part?).
This may be an issue for users who have global `@vue/cli` v4.x installed and try to call `vue add router` on a v3 project. We can add a preflight check in `generator/index.js` to prompt users the incompatibility.

pksunkara:
> As I tested out, the router plugin does not work well with @vue/cli-service 3.x (maybe due to the REPLACE part?).

I wanted to add the assert version stuff we recently added to vue but kinda forgot about it. It fixes this issue.

pksunkara:
Addressed your comments. This is now ready for another review.

packages/@vue/cli/lib/util/features.js
```diff
@@ -3,9 +3,6 @@ const { toShortPluginId } = require('@vue/cli-shared-utils')
 
 exports.getFeatures = (preset) => {
   const features = []
-  if (preset.router) {
```
sodatea: Don't delete this line. It's for displaying preset features.
image

packages/@vue/cli-plugin-router/generator/template/src/App.vue
```diff
@@ -0,0 +1,65 @@
+---
+extend: '@vue/cli-service/generator/template/src/App.vue'
+replace:
+  - !!js/regexp /<template>[^]*?<\/template>/
+  - !!js/regexp /<script>[^]*?<\/script>/
```
sodatea: This will still introduce 2 extraneous blank lines. Could be eliminated by changing the regexp to
```suggestion
  - !!js/regexp /\n<script>[^]*?<\/script>\n/
```

sodatea:
Only 2 small modifications suggested.
The rest of this PR looks great to me.
Thanks for your efforts! 😄

pksunkara:
Updated

PR #4204, opened 2019-06-27
feat(chore): Minor refactor

jamesgeorge007:
Remove redundant method names.


jamesgeorge007:
cc @sodatea

sodatea:
Thanks for your enthusiasm. But the extra function name is necessary here. Otherwise, we may lose track of the function name in the error stack.

For more details, see https://eslint.org/docs/rules/func-names#further-reading

Also next time please exclude `package-lock.json` when committing.

PR #4208, opened 2019-06-28
feat(cli-service-global): change ‘resolveEntry’ to look in src/ aswell

### LINKED ISSUE #4207: "vue serve | build" looks in scr/ for valid entry files aswell
When I open a vue project in my ide the terminal points on the root directory of the project. When I type in "vue serve" it tells that no valid entry point is found an I should rather use "npm run serve". A little change in the "cli-service-global" package could solve the problem.

No change on the api.

bmacher:
Serve or build the project from the root project directory.

close #4207


sodatea:
See the issue comment: https://github.com/vuejs/vue-cli/issues/4207#issuecomment-507007003

PR #4212, opened 2019-06-30
feat(typescript): add an option to not rename all .js files to .ts and to set `allowJs` to true

### LINKED ISSUE #2676: Do not rename all *.js files to *.ts after running `vue add @vue/typescript` - give users a choice to enable `allowJs:true` mode.
This feature makes the JS -> TypeScript migration easier

At the moment when we run `@vue/typescript` then vue CLI adds all required dependencies and creates all required configuration files - this is fine. But it also renames all *.js files to *.ts. IMO it doesn't help with anything. Such an automation only makes sense if we have a fresh Vue project, but if we already have an existing application, then it doesn't make the migration any easier. It's very unlikely that existing files will compile successfully. Missing types and different import syntax will cause lots of compilation failures.

So I suggest that after the `vue add @vue/typescript` command is executed:
- the CLI should as whether to rename all the *.js files to *.ts
- the CLI should ask whether to enable the `allowJs: true` mode.
- the only file that should be renamed is the `main.js` file.

 [`allowJs: true`](https://www.typescriptlang.org/docs/handbook/compiler-options.html) is the best way to migrate a JS project to TypeScript progressively. Thanks to this we can still use existing JS modules as they are, write new modules in TypeScript, and migrate existing ones one by one.

It's not my personal opinion. Setting `allowJs` to `true` is recommended in the official [Migration from Javascrtipt](https://www.typescriptlang.org/docs/handbook/migrating-from-javascript.html) guide.

NataliaTepluhina:
## What does this PR do?

Close #2676 

At the moment when we run `vue add`, Vue CLI renames all `*.js` files to `*.ts`. This PR introduces two new prompts on `@vue/cli-plugin-typescript` late-invokation:

1) `convertJsToTs`: if set to `true`, renames all `.js` files to `.ts`. Otherwise renames only `main.js` -> `main.ts`;
2) `allowJs`: if set to `true`, adds `allowJs: true` to TSConfig compiler options.

### DIFF
--- a/packages/@vue/cli-plugin-typescript/generator/convert.js
+++ b/packages/@vue/cli-plugin-typescript/generator/convert.js
@@ -21,8 +21,11 @@ module.exports = (api, { tsLint = false } = {}, convertAllFiles = true) => {
       }
     } else {
       // rename only main file to main.ts
-      const content = files[api.entryFile]
       const tsFile = api.entryFile.replace(jsRE, '.ts')
+      let content = files[api.entryFile]
+      if (tsLint) {
+        content = convertLintFlags(content)
+      }
       files[tsFile] = content
       delete files[api.entryFile]
     }
--- a/packages/@vue/cli-plugin-typescript/generator/convert.js
+++ b/packages/@vue/cli-plugin-typescript/generator/convert.js
@@ -1,4 +1,4 @@
-module.exports = (api, { tsLint = false } = {}, convertAllFiles = true) => {
+module.exports = (api, { tsLint = false, convertAllFiles = true } = {}) => {
   const jsRE = /\.js$/
   const excludeRE = /^tests\/e2e\/|(\.config|rc)\.js$/
   const convertLintFlags = require('../lib/convertLintFlags')
--- a/packages/@vue/cli-plugin-typescript/generator/convert.js
+++ b/packages/@vue/cli-plugin-typescript/generator/convert.js
@@ -1,22 +1,30 @@
-module.exports = (api, { tsLint = false } = {}) => {
-  // delete all js files that have a ts file of the same name
-  // and simply rename other js files to ts
+module.exports = (api, { tsLint = false } = {}, convertAllFiles = true) => {
   const jsRE = /\.js$/
   const excludeRE = /^tests\/e2e\/|(\.config|rc)\.js$/
   const convertLintFlags = require('../lib/convertLintFlags')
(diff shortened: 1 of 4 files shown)


NataliaTepluhina:
@Akryum @sodatea could you please take a look? 🙇‍♀

packages/@vue/cli-plugin-typescript/prompts.js
```diff
@@ -36,6 +36,18 @@ const prompts = module.exports = [
         value: 'commit'
       }
     ]
+  },
+  {
+    name: `convertJsToTs`,
+    type: `confirm`,
+    message: `Convert all .js files to .ts?`,
+    default: false
```
pksunkara: I think the default should be true.

packages/@vue/cli-plugin-typescript/generator/convert.js
```diff
@@ -1,22 +1,30 @@
...
           }
-          files[tsFile] = content
+          delete files[file]
         }
-        delete files[file]
       }
+    } else {
+      // rename only main file to main.ts
+      const content = files[api.entryFile]
+      const tsFile = api.entryFile.replace(jsRE, '.ts')
```
pksunkara: I might be completely wrong here but don't we need to `convertLintFlags` here too?

packages/@vue/cli-plugin-typescript/prompts.js
NataliaTepluhina: Agree, let's change it to true

packages/@vue/cli-plugin-typescript/generator/convert.js
NataliaTepluhina: Yes, you're right 😅 Added a `convertLintFlags`

packages/@vue/cli-plugin-typescript/generator/convert.js
```diff
@@ -1,22 +1,33 @@
-module.exports = (api, { tsLint = false } = {}) => {
-  // delete all js files that have a ts file of the same name
-  // and simply rename other js files to ts
+module.exports = (api, { tsLint = false } = {}, convertAllFiles = true) => {
```
sodatea: What about merging the options into the second argument?

NataliaTepluhina: Done! 👍

PR #4230, opened 2019-07-03
fix: make webpack `use` values consistent

shadow-light:
The use key is `pug-plain` despite a similar item above it using the key `pug-plain-loader`. This just makes them consistent.

This has a small chance of breaking projects that already override that particular loader config. So feel free to ignore this if it isn't worth it.


LinusBorg:
@sodatea Technically, this is a breaking change as people might hook into this config setting in existing projects. 

It's a simple enough fix to put it into v4 though, I think.

@sodatea you agree?

sodatea:
Yeah I think so.

PR #4232, opened 2019-07-03
Updated for Zeit Now V2

ivansieder:


ivansieder:
Thank you Natalia for the hint!

Sadly I couldn't find any better way than this one and I'm not quiet sure if that's something that should be added directly to the docs/guide? As it might make it very long. What do you think?
```json
{
    "name": "my-example-app",
    "version": 2,
    "builds": [
        {
            "src": "dist/**",
            "use": "@now/static"
        }
    ],
    "routes": [
        {
            "src": "/(js|css|img)/(.*)",
            "dest": "/dist/$1/$2"
        },
        {
            "src": "/favicon.ico",
            "dest": "/dist/favicon.ico"
        },
        {
            "src": "/(.*)",
            "dest": "/dist"
        }
    ],
    "alias": "vue-example"
}
```

NataliaTepluhina:
@ivansieder I think I've seen the same approach on other PRs for Now 2. It seems to be the only possible way for now (please correct me if I am wrong)

ivansieder:
@NataliaTepluhina I've added it now as an additional config, but that way it's clearer what can be left out, if the router mode is not set to history mode.

docs/guide/deployment.md
```diff
@@ -222,35 +224,57 @@ npm install -g now
...
+        },
+        {
+          "src": "/(.*)",
+          "dest": "/dist"
+        }
+      ],
+      "alias": "vue-example"
+    }
+   ```
+   This additional config is required in order to avoid issues when directly deep-linking to a specific page (e.g. when opening my-example-app.now.sh/some-subpage, you would presented with a 404 error otherwise).
```
NataliaTepluhina:
```suggestion
   This additional config is required in order to avoid issues when directly deep-linking to a specific page (e.g. when opening `my-example-app.now.sh/some-subpage`, you would be presented with a 404 error otherwise).
```

docs/guide/deployment.md
```diff
@@ -222,35 +224,57 @@ npm install -g now
...
+      "routes": [
+        { "src": "/(.*)", "dest": "dist/$1" }
+      ],
+      "alias": "vue-example"
     }
     ```
-
-    You can further customize the static serving behavior by consulting [Now's documentation](https://zeit.co/docs/deployment-types/static).
+    
+   In case you want to deploy an application with router mode set to history, the config file should look like the following (if you have different folder names, update your config accordingly):
```
NataliaTepluhina:
```suggestion
   In case if you want to deploy an application with router mode set to `history`, the config file should look like the following (if you have different folder names, update your config accordingly):
```

ivansieder:
Thank you Natalia, I didn't even see that error 🙊

Issue #4024, opened 2019-05-18
apply is not a function at Generator.Plugins.forEach


CodeSpent:
### Version
3.7.0

### Reproduction link
[https://github.com/CodeSpent/CodeSpent.dev](https://github.com/CodeSpent/CodeSpent.dev)

### Environment info
```
Environment Info:

  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
  Binaries:
    Node: 10.15.3 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.17763.1.0
  npmPackages:
    @fortawesome/vue-fontawesome: ^0.1.6 => 0.1.6
    @sagalbot/vueable: 0.0.3 => 0.0.3
    vue: ^2.5.11 => 2.6.10
    vue-chartjs: ^3.4.2 => 3.4.2
    vue-cli-webpack:  1.0.0
    vue-github-api: ^0.1.7 => 0.1.7
    vue-hot-reload-api:  2.3.3
    vue-loader: ^13.0.5 => 13.7.3
    vue-resource:  1.5.1
    vue-style-loader:  3.1.2
    vue-template-compiler: ^2.4.4 => 2.6.10
    vue-template-es2015-compiler:  1.9.1
    vuetify: ^1.5.14 => 1.5.14
    vuex: ^3.1.1 => 3.1.1
  npmGlobalPackages:
    @vue/cli: Not Found
```

### Steps to reproduce
use `vue add router`

### What is expected?
vue add router

### What is actually happening?
```C:\.......\......\Development\CodeSpent.dev> vue add router
? Use history mode for router? (Requires proper server setup for index fallback in production) No
(node:4328) UnhandledPromiseRejectionWarning: TypeError: apply is not a function
    at Generator.plugins.forEach (C:\Users\hanfo\AppData\Roaming\npm\node_modules\@vue\cli\lib\Generator.js:103:7)
    at Array.forEach (<anonymous>)
    at new Generator (C:\Users\hanfo\AppData\Roaming\npm\node_modules\@vue\cli\lib\Generator.js:101:13)
    at Function.runGenerator (C:\Users\hanfo\AppData\Roaming\npm\node_modules\@vue\cli\lib\invoke.js:123:21)
[...]

PR #4242, opened 2019-07-05
feat!: make vuex a separate plugin

### LINKED ISSUE #2335: Make router and vuex proper packages

pksunkara:
Fixes #2335, #4024

### DIFF
--- a/packages/@vue/cli-plugin-vuex/generator/template/src/store/index.js
+++ b/packages/@vue/cli-plugin-vuex/generator/template/src/store/index.js
@@ -5,12 +5,11 @@ Vue.use(Vuex)
 
 export default new Vuex.Store({
   state: {
-
   },
   mutations: {
-
   },
   actions: {
-
+  },
+  modules: {
   }
 })
--- a/docs/guide/plugins-and-presets.md
+++ b/docs/guide/plugins-and-presets.md
@@ -105,7 +105,6 @@ Here's an example preset:
 ``` json
 {
   "useConfigFiles": true,
-  "vuex": true,
   "cssPreprocessor": "sass",
   "plugins": {
     "@vue/cli-plugin-babel": {},
@@ -113,7 +112,8 @@ Here's an example preset:
       "config": "airbnb",
       "lintOn": ["save", "commit"]
     },
-    "@vue/cli-plugin-router": {}
+    "@vue/cli-plugin-router": {},
+    "@vue/cli-plugin-vuex": {}
   }
 }
 ```
--- a/docs/ru/guide/plugins-and-presets.md
+++ b/docs/ru/guide/plugins-and-presets.md
@@ -105,7 +105,6 @@ vue add @vue/eslint --config airbnb --lintOn save
 ``` json
 {
   "useConfigFiles": true,
-  "vuex": true,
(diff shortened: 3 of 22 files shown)


packages/@vue/cli-plugin-vuex/generator/template/src/store/index.js
```diff
@@ -4,13 +4,6 @@ import Vuex from 'vuex'
...
 export default new Vuex.Store({
-  state: {
-
-  },
-  mutations: {
-
-  },
-  actions: {
-
+  modules: {
```
sodatea: `modules` field with no actual modules may be confusing for newbies?

For now, we can stick to the original structure and introduce modules later with the planned snippets generator feature.

packages/@vue/cli-plugin-vuex/prompts.js
```diff
@@ -0,0 +1,5 @@
+// these prompts are used if the plugin is late-installed into an existing
```
sodatea: Not necessary if empty.

pksunkara: I wanted to add some prompts for snippets feature after I finish the hooks PR.

pksunkara: Also, I want to make a PR where there's a single source of truth for the prompts. Right now, there are many duplicates.

packages/@vue/cli-plugin-vuex/generator/template/src/store/index.js
pksunkara: How about the state, actions and also modules? codemods will work better if the `modules` is present here.

sodatea: Yeah that seems fine.

pksunkara:
Addressed review comments and force pushed the commit.

packages/@vue/cli-plugin-vuex/prompts.js
Akryum: What do you want to refactor precisely? I'm doing a refactor on `@vue/cli` already.

pksunkara: Some prompts in current repo are duplicated, once for `vue create` and once for `vue add @vue/{plugin}`. I want to move all those into cli-shared-utils and take them from there wherever necessary. Single Source of truth.

Akryum: It will be in the upcoming `@vue/cli-utils` package then.

pksunkara: Will you take over that task as part of your refactor then?

Akryum: Yeah I can do it

PR #4253, opened 2019-07-09
feat: upgrade webpack

DRoet:
Now that https://github.com/webpack/webpack/issues/8656 has been resolved in the latest webpack release, I believe the version range restriction can finally be removed.

Should the updated `.lock` file be commited aswell?


sodatea:
Better use `^4.0.0` as there might be other plugins in the ecosystem also has this version restriction, forcing the latest version may result in failing dependency deduping.

We do lockfile maintenance regularly so no need to commit it in PRs.

DRoet:
@sodatea done

sodatea:
Will merge this after fixing the failing tests.
For now, I've narrowed it down to this commit in webpack: https://github.com/webpack/webpack/pull/8917

sodatea:
😅Just found out that both webpack & preload-webpack-plugin are doing it right. It's html-webpack-plugin that adds the unused `chunk-common` chunk to the html file…
I'll fix the test cases.

Issue #767, opened 2018-02-02
Env variables don't seem to load from a .env file on the new cli


alexjoverm:
### Version
3.0.0-alpha.5

### Reproduction link
[https://github.com/alexjoverm/vue-cli-env-reproduce](https://github.com/alexjoverm/vue-cli-env-reproduce)

### Steps to reproduce
Clone, run `npm install` and `npm run serve` on the repo provided.

### What is expected?
According to the code of the [Service](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-service/lib/Service.js#L61), some env variables should be loaded taken from a path, which are parsed on the [loadEnv](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-service/lib/util/loadEnv.js) util.

By adding some logs, it seems that indeed the path it takes when running `npm run serve` is `${rootDirectory}/.env.development` and `${rootDirectory}/.env.development.local`. They're indeed loaded. 

So, by the `console.log(process.env)` line in the main.js file, those variables should be printed in the console.

### What is actually happening?
They're not listed in the console.

Note: if this is not supposed how is it possible to add env variables?

yyx990803:
Only variables that starts with `VUE_APP_` are loaded: https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-service/lib/util/resolveClientEnv.js#L1

irufus:
Why is this not part of the documentation?

kispi:
> Why is this not at the very top part of the documentation?
> 
> Especially where you give your example
> 
> > FOO=bar
> > VUE_APP_SECRET=secret
> 
> FOO should be VUE_APP_FOO
> 
> Edited:
> It is documented but easily overlooked.

Yes I also overlooked

I was yelling on myself 'why it's not working!' like 10 minutes and eventually came here



*READ THE DOCUMENTATION*
"EACH AND EVERY CHARACTER"

3zzy:
For those who arrived here looking for a solution, but already use have your variables prefixed with `VUE_APP_`, just restart.

gregfr:
> just restart.
This should be stressed out in the docs, I spent two hours figuring out why the documented example was not working...

malandles:
I understand that the prefix is a good thing to have when your Vue app is coupled with another framework like Laravel, but when the app is standalone, the prefix requirement is a bit exagerated no?

[...]

PR #4257, opened 2019-07-10
Update mode-and-env.md

TechieForFun:
Fixed example of keys, which they should start with `VUE_APP_`, related to #767

### DIFF
--- a/docs/guide/mode-and-env.md
+++ b/docs/guide/mode-and-env.md
@@ -82,7 +82,7 @@ In both cases, the app is built as a production app because of the `NODE_ENV`, b
 
 ### Using Env Variables in Client-side Code
 
-You can access them in your application code:
+You can access env variables in your application code:
 
 ``` js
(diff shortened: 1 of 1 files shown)


TechieForFun:
Thanks for reviewing.

Indeed. Because it can be easily overlooked currently.
I changed it as you suggested.

docs/guide/mode-and-env.md
```diff
@@ -80,7 +82,7 @@ In both cases, the app is built as a production app because of the `NODE_ENV`, b
 
 ### Using Env Variables in Client-side Code
 
-Only variables that start with `VUE_APP_` will be statically embedded into the client bundle with `webpack.DefinePlugin`. You can access them in your application code:
+You can access them in your application code:
```
NataliaTepluhina:
```suggestion
You can access env variables in your application code:
```

NataliaTepluhina:
Thank you for the changes!

PR #4261, opened 2019-07-10
Add `--inline-vue` flag for `build` command to avoid externalization of Vue in lib & wc mode

### LINKED ISSUE #4055: Include Vue instance when Build Target is Lib mode
We are building multiple micro frontend applications that are then stitched together into singular spa experience at run-time. As such, building these applications in using a build target "--target app" is not a viable approach. We need a single file that contains the application source code to load up as we are stitching applications together. For this reason, using a build target "--target lib" seemed like a ideal solution for us. However, we've noticed the build tool contains logic when it is building the application that it excludes Vue instances automatically, which you folks have clearly indicated in your documentation. We weren't able to override this behavior using either the configureWebpack object nor the chainWebpack() callback function. 

As simple fix would be to move the process of externalizing Vue up prior to applying configureWebpack config object or executing chainWebpack() callback. That way, you are giving programmers the flexibility to include it if it suits their architectural scenario without it being a breaking change.

romansp:
This PR resolves an issue when build is targeting Lib, but for whatever reason Vue instance should be preserved in final bundle.

I followed @seteam1 original suggestion. And also added test case and updated docs. With recommendations by @sodatea.

Option is only used for `--target lib`, but might be useful for `--target wc` as well. Let me know and I update this PR.

Would like to hear some feedback is if this change might be accepted to upstream.

Resolves #4055


romansp:
Rebased on top of `dev` branch. WIP on reworking this PR from `vue.config.js` option into `--inline-vue` command line flag.

packages/@vue/cli-service/lib/commands/build/resolveWcConfig.js
```diff
@@ -100,7 +100,9 @@ module.exports = (api, { target, entry, name }) => {
     rawConfig.externals = [
       ...(Array.isArray(rawConfig.externals) ? rawConfig.externals : [rawConfig.externals]),
       {
-        vue: 'Vue'
+        ...(!inlineVue ? {
+          vue: 'Vue'
+        } : {})
```
sodatea: This can be written in a more concise way:
```
{ ...inlineVue || { vue: 'Vue' } }
```

packages/@vue/cli-service/lib/commands/build/resolveLibConfig.js
```diff
@@ -97,11 +97,13 @@ module.exports = (api, { entry, name, formats, filename }, options) => {
...
-          commonjs2: 'vue',
-          root: 'Vue'
-        }
+        ...(!inlineVue ? {
+          vue: {
+            commonjs: 'vue',
+            commonjs2: 'vue',
+            root: 'Vue'
+          }
+        } : {})
```
sodatea:
```
...(inlineVue || {
  vue: {
    commonjs: 'vue',
    commonjs2: 'vue',
    root: 'Vue'
  }
})
```

romansp:
Pushed code style fixes and also improved consistency for related docs.

PR #4275, opened 2019-07-11
feat!: confirm before invoke/add/upgrade if project has uncommitted changes

sodatea:


Akryum:
Can you move the function to `shared-utils`?

davidsickmiller:
I've now learned how vue plugin generators work, so I no longer need additional warnings, but I wanted to suggest an idea to be more helpful to new users:

It's pretty intuitive that adding a module (e.g. "vue add router") will add packages to package.json, package-lock.json, and node_modules.  It's not as obvious, but probably OK, that it will add new files under src/.  It's comparatively most surprising that it will also overwrite existing files under src/.

I think it would be nice to change this warning to:
1. Only be displayed if it's going to overwrite an existing file, and if so
2. List which files will be overwritten.  Also,
3. Show the warning even if all files have been committed to git.

Regarding #3, even if I have the previous version in my repo, I think it's nice to know up front that I'll need to manually merge certain files, rather than running `git diff` after each vue add to find out what happened.

ianharrd69 (commented):
I own this

PR #4302, opened 2019-07-17
chore!: move dev configs into serve command

sodatea:
BREAKING CHANGE:
These configs are only meaningful when used with a running dev server.

Technically, it is a breaking change.
For example, this breaks use cases in which a user runs their own dev
server instead of calling `vue-cli-service serve`.


kvakazyambra:
So what I have to do if I need to use my own dev server?

PR #4309, opened 2019-07-18
Update `cypress` to version `3.4.0` due to `lodash` high severity vulnerability (#4296)

### LINKED ISSUE #4296: E2E Cypress Plugin - Lodash prototype polution
### Version
3.9.2

### Environment info
```
Environment Info:
```

### Steps to reproduce
Install cli-plugin-e2e-cypress plugin
Run npm audit

### What is expected?
Cypress plugin using version of lodash higher than 4.17.12

cli-plugin-e2e-cypress plugin uses lodash package vulernable to Prototype Pollution: https://www.npmjs.com/advisories/1065

jclaudan:
for fix #4296

### DIFF
--- a/packages/@vue/cli-plugin-e2e-cypress/package.json
+++ b/packages/@vue/cli-plugin-e2e-cypress/package.json
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "@vue/cli-shared-utils": "^3.9.0",
-    "cypress": "^3.2.0",
+    "cypress": "^3.4.0",
     "eslint-plugin-cypress": "^2.2.1"
   }
 }


sodatea:
1. Don't open multiple PRs with the same content
2. It does not solve the problem that you claimed. With `^3.2.0` package manager will automatically resolve to the latest 3.x version available unless a cached version is preferred somehow, so it works basically the same as `^3.4.0`. And, most importantly, cypress didn't update their dependency at 3.4.0, they only updated lodash version **after** 3.4.0 release, and a new version is not available yet.

jclaudan:
> 1. Don't open multiple PRs with the same content
> 2. It does not solve the problem that you claimed. With `^3.2.0` package manager will automatically resolve to the latest 3.x version available unless a cached version is preferred somehow, so it works basically the same as `^3.4.0`. And, most importantly, cypress didn't update their dependency at 3.4.0, they only updated lodash version **after** 3.4.0 release, and a new version is not available yet.

Oh ok sorry for my mistake and thank you for reply me

Issue #4311, opened 2019-07-18
Default value of the --filename parameter results in broken build if scoped package name is used


MewesK:
### Version
3.9.3

### Reproduction link
[https://github.com/MewesK/vue-cli-scoped-lib-bug](https://github.com/MewesK/vue-cli-scoped-lib-bug)

### Environment info
```
This bug can be found in all versions (3.0 - 4.0 alpha) and all environments.

  System:
    OS: Windows 10
    CPU: (8) x64 Intel(R) Core(TM) i7-6820HQ CPU @ 2.70GHz
  Binaries:
    Node: 11.14.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.15.2 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.7.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.18362.1.0
  npmPackages:
    @fortawesome/vue-fontawesome: ^0.1.6 => 0.1.6
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0
    @vue/babel-plugin-transform-vue-jsx:  1.0.0
    @vue/babel-preset-app:  3.8.0
    @vue/babel-preset-jsx:  1.0.0
    @vue/babel-sugar-functional-vue:  1.0.0
    @vue/babel-sugar-inject-h:  1.0.0
    @vue/babel-sugar-v-model:  1.0.0
    @vue/babel-sugar-v-on:  1.0.0
    @vue/cli-overlay:  3.8.0
    @vue/cli-plugin-babel: ^3.7.0 => 3.8.0
    @vue/cli-plugin-eslint: ^3.7.0 => 3.8.0
    @vue/cli-plugin-unit-jest: ^3.7.0 => 3.8.0
    @vue/cli-service: ^3.7.0 => 3.8.0
    @vue/cli-shared-utils:  3.8.0
    @vue/component-compiler-utils:  2.6.0
    @vue/eslint-config-airbnb: ^4.0.0 => 4.0.0
    @vue/preload-webpack-plugin:  1.1.0
    @vue/test-utils: 1.0.0-beta.29 => 1.0.0-beta.29
    @vue/web-component-wrapper:  1.2.0
    bootstrap-vue: ^2.0.0-rc.26 => 2.0.0-rc.21
    eslint-plugin-vue: ^5.2.2 => 5.2.2
    jest-serializer-vue:  2.0.2
[...]

PR #4320, opened 2019-07-19
Fixed: "Default value of the --filename parameter results in broken build if scoped package name is used #4311"

MewesK:
See #4311 for more information


MewesK:
Looks better now. Sorry I wasn't able to get your test suite running at work. That's why I used only your circleci which only runs after creating a PR.

PR #4335, opened 2019-07-22
Recommend npx instead of global env corruption

rayfoss:
There is no downside to using 
```
npm init
npm install @vue/cli
npx vue create
```

### DIFF
--- a/docs/guide/prototyping.md
+++ b/docs/guide/prototyping.md
@@ -8,7 +8,7 @@ npm install -g @vue/cli-service-global
 yarn global add @vue/cli-service-global
 ```
 
-The drawback of `vue serve` is that it relies on globally installed dependencies which may be inconsistent on different machines. Therefore this is only recommended for rapid prototyping.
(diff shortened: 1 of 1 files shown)


Akryum:
I don't understand what's the point of your PR 😅 
`vue serve` only purpose is rapid prototyping, for a real project you should use `vue create` which will install `@vue/cli-service` locally in your project.

rayfoss:
For a real project you should also probably use `@vue/cli` which the docs also recommend installing, globally. Globally installing dependencies is the source of 50% of my headaches... global corruption is also hard-ish to debug. We should really not ever recommend installing any dependency globally, aside from maybe nvm and nodejs.

sodatea:
A more succinct command: `npx @vue/cli create my-project`

But it serves a different purpose than `vue serve` and is not **rapid** at all.

---

> We should really not ever recommend installing any dependency globally

I think it's only a design tradeoff.

PR #4367, opened 2019-07-26
chore(cli-service): word spelling in comments

yiliang114:


yiliang114:
sorry, it was my fault. Now there are only  necessary changes.

Issue #2525, opened 2018-09-14
Allow custom publicPath to be set alongside baseUrl


lsapan:
### What problem does this feature solve?
Right now any attempt to set `publicPath` prevents the bundle from compiling with the following error:
```
Configuration Error: Avoid modifying webpack output.publicPath directly. Use the "baseUrl" option instead.
```

I understand why that validation is there, as it prevents accidental misconfiguration when users aren't aware of the `baseUrl` option. That said, there should be some way to bypass this error (or make it a warning).

My bundle gets served from my webapp which is running on a different localhost port (let's say 4000). As such, HMR attempts to load from `http://localhost:4000/xxx` instead of `http://localhost:8080/xxx`. I can't put the `http://localhost:8080/` in the `baseUrl` because it simply tacks that onto itself (`http://localhost:8080/http://localhost:8080/`).

I really just need to override the `publicPath` for webpack, but keep the `baseUrl` exactly as it is. I've tried a thousand different things but there doesn't seem to be any other way around this.

### What does the proposed API look like?
Either make it a warning, or add some kind of flag that says "hey, I know about baseUrl and I've set it, I just also want to set publicPath too."

Frankly, if both `baseUrl` and `publicPath` are set, it can probably be assumed that it's intentional. Maybe only display the error if `publicPath` is set without `baseUrl`.

LinusBorg:
Isn't `devServer.public` the better solution to your problem of HMR contacting the wrong backend?

lsapan:
Ah, shoot, I tried changing `devServer.publicPath` but I completely missed `devServer.public`. Judging from the docs that’s exactly what I need. Thanks!

lsapan:
I stand corrected! It really seems like that should work, however, for some reason `public` only seems to affect the initial request (i.e. `http://localhost:8080/sockjs-node/info?t=1537018718015`).

After that, subsequent requests are still made to my app server (`http://localhost:4000/09018356538331e29b1e.hot-update.json`).

I've tried combining `devServer.public` and `devServer.publicPath` but it does not change the fact that it is still looking for the files relative to the window's location.

That said, I did try commenting out the validation code in `validateWebpackConfig.js` and it works with the (non dev-server) `publicPath`!

So this does not work:
```
module.exports = {
    devServer: {
        public: 'localhost:8080',
        publicPath: 'http://localhost:8080/',
        headers: {
            'Access-Control-Allow-Origin': '*',
        },
    },
}
```

But this does:
```
[...]

PR #4388, opened 2019-08-01
chore!: remove the already-deprecated `baseUrl` option

sodatea:


Aymkdn:
So what would be the workaround for https://github.com/vuejs/vue-cli/issues/2525 ?

PR #4416, opened 2019-08-09
chore: use a gitflow-like workflow for the project

sodatea:
Now that we do v3 & v4 releases in parallel (even after v4 stable release, v3 will continue to receive bug fixes), I'd like to tweak the git workflow a little bit to make it easier.

So, inspired by Gitflow, I propose that we have two production-ready branches - `master` & `next`. `master` is for the `latest` `dist-tag` version on npm, and `next` for `next` `dist-tag` (pre-release versions). It's not only used for v4 betas, but we can also have prereleases for feature releases in the future.

I've also set up Netlify to deploy https://next.cli.vuejs.org from the `next` branch so that users can preview documentation for the new features.

In the past, the website is deployed from the `docs` branch. It seems not so obvious for many users. Many documentation PRs are targeting the `dev` branch. So I now explicitly clarified this in the contributing guide and PR template. Also, the target branch is changed to `master` or `next` so that the code and documentation are in sync, no longer need to cherry-pick & merge across multiple branches.

Though our documentation currently covers features in both v3 & v4, with version-related notes, it may change over time. So I'd also like to set up version-specific documentation, deployed directly from the version archive branch (e.g. `v3.cli.vuejs.org` deployed from `v3` branch).

PR #4420, opened 2019-08-09
Update parallel config description

danielwaltz:
You are able to pass a number to this configuration option to set the amount of workers, but it was not documented. Hopefully this change will help save someone else some headache!


NataliaTepluhina:
Thank you for this change and for targeting the correct branch according to https://github.com/vuejs/vue-cli/pull/4416 😉

PR #4447, opened 2019-08-16
fix(eslint): eslint should override env

LinusBorg:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

The unit test packages support tests in /tests/unit and \*/\_\_tests\_\_/ folders, but eslint config for jest/mocha  was only applied to the former

This PR adds overrides to the eslint config for files in the latter type of folder so eslint can successfully detect and lint jest/mocha globals

### DIFF
--- a/packages/@vue/cli-plugin-unit-jest/generator/index.js
+++ b/packages/@vue/cli-plugin-unit-jest/generator/index.js
@@ -82,7 +82,7 @@ module.exports = (api, _, __, invoking) => {
       eslintConfig: {
         overrides: [
           {
-            files: ['**/__tests__/*.js', '**/__tests__/*.jsx', '**/__tests__/*.ts', '**/__tests__/*.tsx'],
+            files: ['**/__tests__/*.{j,t}s?(x)'],
             env: {
               jest: true
             }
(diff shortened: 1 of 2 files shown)


packages/@vue/cli-plugin-unit-jest/generator/index.js
```diff
@@ -77,6 +77,19 @@ module.exports = (api, _, __, invoking) => {
 
   if (api.hasPlugin('eslint')) {
     applyESLint(api)
+
+    api.extendPackage({
+      eslintConfig: {
+        overrides: [
+          {
+            files: ['**/__tests__/*.js', '**/__tests__/*.jsx', '**/__tests__/*.ts', '**/__tests__/*.tsx'],
```
sodatea: Seems too verbose. What about `'**/__tests__/*.{j,t}s?(x)'`?

LinusBorg: I initially tried that but it seems that eslint only supports basic wildcards with *

I may be wrong but couldn't get it to work with glob/regexp matching syntax

sodatea: ESLint uses `minimatch` underlyingly. As I tested, the abovementioned glob works well.

LinusBorg: I see. Will adjust.

LinusBorg: Done

PR #4448, opened 2019-08-16
feat: --skip-plugins

### LINKED ISSUE #3830: Option to turn off PWA generation
If building for multiple targets, some of them might not support PWA. It would be great to have an option to turn off the functionality

Currently the only way is to use a env var `VUE_CLI_BUILD_TARGET` different to `app`: https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-pwa/index.js#L4

This line could e.g. be changed to:

```js
if (target && target !== 'app' || options.pwa.off === true) {
      return
}
```

### LINKED ISSUE #4262: Option to disable vue-cli plugins through vue.config.js
I have a project with multiple entry files and multiple vue.config.js, or a dynamic vue.config.js. Depending on the entry or target I would like to enable/disable certain vue-cli plugins.

config.disablePlugins = []

LinusBorg:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Adds a new cli flag named `--skip-plugins` that works with all commands and allows to skip a plugin, which means it will not be applied for that run.

close #4262
close #3830


sodatea:
Looks good to me. But there's now a conflict to resolve…

LinusBorg:
conflict resolved :)

packages/@vue/cli-service/lib/options.js
```diff
@@ -8,6 +8,7 @@ const schema = createSchema(joi => joi.object({
   filenameHashing: joi.boolean(),
   runtimeCompiler: joi.boolean(),
   transpileDependencies: joi.array(),
+  skipPlugins: joi.array(),
```
sodatea: Seems redundant?

LinusBorg: Oh right, that's a leftover from my initial attempt to solve it through a config option instead of a flag.

PR #4451, opened 2019-08-17
use the specified port and not a ramdom port in range[port,highest port]

mpromonet:

- [ x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
This PR allow to fix the port running "vue serve" related in https://stackoverflow.com/questions/57536785/vue-npm-run-serve-starts-on-random-port

portfinder give a ramdom port in the range [basePort,highestPort], so asking for port 8080 result to a random port between [8080 and 40000]

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -101,7 +101,7 @@ module.exports = (api, options) => {
     const useHttps = args.https || projectDevServerOptions.https || defaults.https
     const protocol = useHttps ? 'https' : 'http'
     const host = args.host || process.env.HOST || projectDevServerOptions.host || defaults.host
-     portfinder.basePort = Number(args.port || process.env.PORT || projectDevServerOptions.port || defaults.port)
+    portfinder.basePort = Number(args.port || process.env.PORT || projectDevServerOptions.port || defaults.port)
     portfinder.highestPort  = portfinder.basePort + 1
     const port = await portfinder.getPortPromise()
     const rawPublicUrl = args.public || projectDevServerOptions.public
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -101,7 +101,7 @@ module.exports = (api, options) => {
     const useHttps = args.https || projectDevServerOptions.https || defaults.https
     const protocol = useHttps ? 'https' : 'http'
     const host = args.host || process.env.HOST || projectDevServerOptions.host || defaults.host
-    portfinder.basePort = args.port || process.env.PORT || projectDevServerOptions.port || defaults.port
+     portfinder.basePort = Number(args.port || process.env.PORT || projectDevServerOptions.port || defaults.port)
     portfinder.highestPort  = portfinder.basePort + 1
     const port = await portfinder.getPortPromise()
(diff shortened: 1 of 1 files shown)


packages/@vue/cli-service/lib/commands/serve.js
```diff
@@ -102,6 +102,7 @@ module.exports = (api, options) => {
     const protocol = useHttps ? 'https' : 'http'
     const host = args.host || process.env.HOST || projectDevServerOptions.host || defaults.host
     portfinder.basePort = args.port || process.env.PORT || projectDevServerOptions.port || defaults.port
```
minami-minami: The basePort is passed as string type when using environment variable and dotenv.
Then the result is as follows.

```js
// PORT=3000 yarn serve
portfinder.highestPort  = portfinder.basePort + 1 // 30001
```

So you might explicitly cast it to number type.

```suggestion
     portfinder.basePort = Number(args.port || process.env.PORT || projectDevServerOptions.port || defaults.port)
```

AndreasCag:
I think it is an issue with portfinder and should be fixed there.

This PR has problem - if selected port is busy, dev build would throw an error instead of selecting new port.

mpromonet:
hi @AndreasCag 

There is an issue about the recent behaviour change in 1.0.22
https://github.com/http-party/node-portfinder/issues/84

Sure you are right if the port is used, serve will fails.

But as option `--port` (or env PORT) specify a port not a range of port, so it seems strange to ask for a port and another is chosen. Maybe it could be better to extend the option to specified a range ?

Best Regards,
Michel.

hbenl:
@AndreasCag The issue is not really in `portfinder`, `vue-cli` is relying on `portfinder` to try ports in ascending order, but the `portfinder` documentation doesn't specify the order in which ports are tried, so it is a bit foolish for `vue-cli` to rely on this undocumented behavior. Of course it is also foolish for `portfinder` to change this behavior _in a minor bugfix release_, I'm sure this will break a lot of other packages that also rely on the old behavior.
@mpromonet I agree that it seems strange that `vue-cli` chooses a different port if the specified port is used, but that seems to be the desired behavior, otherwise it would be pointless to use `portfinder` at all.

BarthesSimpson:
There is a fix for this out for review in portfinder: 
https://github.com/http-party/node-portfinder/pull/86

therealshark:
> @AndreasCag The issue is not really in `portfinder`, `vue-cli` is relying on `portfinder` to try ports in ascending order, but the `portfinder` documentation doesn't specify the order in which ports are tried, so it is a bit foolish for `vue-cli` to rely on this undocumented behavior. Of course it is also foolish for `portfinder` to change this behavior _in a minor bugfix release_, I'm sure this will break a lot of other packages that also rely on the old behavior.

But it actually does specify that it will do that:
https://github.com/http-party/node-portfinder/blob/master/README.md
> By default portfinder will start searching from 8000 and scan until maximum port number (40000) is reached. 40000 is the highest safe port across platforms. If you are not on OSX, you can safely override the max port up to 65535.

hbenl:
@therealshark Oops, I missed that. Thanks for pointing it out.

sodatea:
Closing this now that portfinder has fixed their bug.

PR #4533, opened 2019-08-31
fix(cli-plugin-babel): roperly exports the babel preset

arcanis:

- [x] Bugfix

- [x] No

**Other information:**

The current setup relies on the hoisting for `babel.config.js` to be able to access `@vue/babel-preset-app`, since it's not actually part of the package dependencies. This change re-exports it from `@vue/cli-plugin-babel`, which is an actual dependency.

Ref https://github.com/yarnpkg/berry/issues/368


sodatea:
Thanks for the help! But this change failed our tests. Would you please also change the corresponding test case for this feature?

PR #4537, opened 2019-09-01
docs: remove old note indicating the Nightwatch version

beatfactor:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-plugin-e2e-nightwatch/README.md
+++ b/packages/@vue/cli-plugin-e2e-nightwatch/README.md
@@ -6,7 +6,7 @@
 
 - **`vue-cli-service test:e2e`**
 
-  run e2e tests with [NightwatchJS](http://nightwatchjs.org).
+  run e2e tests with [NightwatchJS](https://nightwatchjs.org).
 
   Options:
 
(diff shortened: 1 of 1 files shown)


beatfactor:
@NataliaTepluhina actually I see that in master nightwatch version is still v0.9. Do you think it might be time for an upgrade? I can follow up with that and also fix the config as well.

NataliaTepluhina:
@sodatea what do you think about upgrading Nightwatch?

sodatea:
We've upgraded nightwatch version in v4. So maybe only remove the note in the `next` branch?

beatfactor:
In the meantime we have released a v1.2 version. I would also like to submit some improvements on the default configuration and maybe some more test examples in the template (to include more advanced topics like page objects, several more custom commands, and instructions on how to write those). 

We can remove the Selenium server, or at least offer standalone support for Chromedriver and Geckodriver by default. You can have a look at our example project [here](https://github.com/nightwatchjs/nightwatch-website-tests) - ideally the Vue plugin would be more in sync with this. It would also help us to get more input on how Nightwatch is being used as part of Vue.

What would be the appropriate branch where to send these improvements?

sodatea:
The latest code lives in the `dev` branch.

PR #4541, opened 2019-09-04
Upgrade Nightwatch to v1.2 and update bundled config and generated tests

beatfactor:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### Overview

This PR upgrades the Nightwatch version bundled with the `cli-e2e-nightwatch` plugin to latest v1.2 version. In addition, it adds several new features for running end-to-end tests with Nightwatch in a Vue application a bit more pleasant of an experience.

### Improvements
- switched to using standalone Webdriver (default with [Chromedriver](https://sites.google.com/a/chromium.org/chromedriver/)) instead of with Selenium server
- added support for running tests in Firefox (via [Geckodriver](https://github.com/mozilla/geckodriver/releases))
- added built-in support for running tests in **headless** mode (via `--headless`) with Chrome and Firefox
- updated generated tests and also updated template with example custom commands, page objects, and external globals file

### New CLI flags
- `--use-selenium`: allows using the Selenium standalone server instead of running Webdriver standalone
- `--parallel`: enable parallel mode via test workers (only available when using chromedriver)
- `--headless`: run tests with the browser (chrome/firefox) in headless mode

### Other
- added more unit tests
- to run unit tests independently than the other packages, in the root folder do:
```sh
$ node scripts/test.js -p cli-plugin-e2e-nightwatch
```

### DIFF
--- a/packages/@vue/cli-plugin-e2e-nightwatch/README.md
+++ b/packages/@vue/cli-plugin-e2e-nightwatch/README.md
@@ -6,29 +6,92 @@
 
 - **`vue-cli-service test:e2e`**
 
-  run e2e tests with [NightwatchJS](http://nightwatchjs.org).
+  Run end-to-end tests with [Nightwatch.js](https://nightwatchjs.org).
 
   Options:
 
   ```
-  --url        run e2e tests against given url instead of auto-starting dev server
-  --config     use custom nightwatch config file (overrides internals)
-  -e, --env    specify comma-delimited browser envs to run in (default: chrome)
-  -t, --test   specify a test to run by name
-  -f, --filter glob to filter tests by filename
+  --url                 run the tests against given url instead of auto-starting dev server
+  --config              use custom nightwatch config file (overrides internals)
+  --headless            use chrome or firefox in headless mode
+  --parallel            enable parallel mode via test workers (only available in chromedriver)
+  --use-selenium        use Selenium standalone server instead of chromedriver or geckodriver
+  -e, --env             specify comma-delimited browser envs to run in (default: chrome)
+  -t, --test            specify a test to run by name
+  -f, --filter          glob to filter tests by filename
   ```
 
-  > Note: this plugin currently uses Nightwatch v0.9.x. We are waiting for Nightwatch 1.0 to stabilize before upgrading.
+  Additionally, all [Nightwatch CLI options](https://nightwatchjs.org/guide/running-tests/#command-line-options) are also supported. 
+  E.g.: `--verbose`, `--retries` etc.
+  
 
-  Additionally, [all Nightwatch CLI options are also supported](https://nightwatchjs.org/guide#command-line-options).
+## Project Structure
+
+The following structure will be generated when installing this plugin. There are examples for most testing concepts in Nightwatch available.    
+
+```
+tests/e2e/
+  ├── custom-assertions/
+  |   └── elementCount.js
+  ├── custom-commands/
+  |   ├── customExecute.js
+  |   ├── openHomepage.js
+  |   └── openHomepageClass.js
+  ├── page-objects/
+  |   └── homepage.js
+  ├── specs/
+  |   ├── test.js
+  |   └── test-with-pageobjects.js
+  └── globals.js
+```
+
+#### `specs`
+The main location where tests are located. Can contain sub-folders which can be targeted during the run using the `--group` argument. [More info](https://nightwatchjs.org/guide/running-tests/#test-groups).
+
+#### `custom-assertions`
+Files located here are loaded automatically by Nightwatch and placed onto the `.assert` and `.verify` api namespaces to extend the Nightwatch built-in assertions. See [writing custom assertions](https://nightwatchjs.org/guide/extending-nightwatch/#writing-custom-assertions) for details.
+
+#### `custom-commands`
+Files located here are loaded automatically by Nightwatch and placed onto the main `browser` api object to extend the built-in Nightwatch commands. See [writing custom commands](https://nightwatchjs.org/guide/extending-nightwatch/#writing-custom-commands) for details.
+
+#### `page objects`
+Working with page objects is a popular methodology in end-to-end UI testing. Files placed in this folder are automatically loaded onto the `.page` api namespace, with the name of the file being the name of the page object. See [working with page objects](https://nightwatchjs.org/guide/working-with-page-objects/) section for details.  
+
+#### `globals.js`
+The external globals file which can hold global properties or hooks. See [test globals](https://nightwatchjs.org/gettingstarted/configuration/#test-globals) section.
 
 ## Configuration
 
-We've pre-configured Nightwatch to run with Chrome by default. If you wish to run e2e tests in additional browsers, you will need to add a `nightwatch.config.js` or `nightwatch.json` in your project root to configure additional browsers. The config will be merged into the [internal Nightwatch config](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/cli-plugin-e2e-nightwatch/nightwatch.config.js).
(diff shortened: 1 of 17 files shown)


packages/@vue/cli-plugin-e2e-nightwatch/package.json
```diff
@@ -26,11 +26,12 @@
     "@vue/cli-shared-utils": "^4.0.0-rc.1",
     "deepmerge": "^3.2.0",
     "execa": "^1.0.0",
-    "nightwatch": "^1.1.11",
-    "selenium-server": "^3.141.59"
+    "nightwatch": "^1.2.2",
+    "chromedriver": "^76.0.1",
```
sodatea: For now please make `chromedriver` a peer dependency. It bumps its major version every 6 weeks and Vue CLI certainly cannot do major releases at that pace. So it's better to be managed on the user side.
In later RC (or minor) versions, I'd like to introduce a preflight check, and a prompt interface for users to choose which version of browser drivers they need.

beatfactor:
Is it possible to upgrade Chrome in AppVeyour?

sodatea:
I've added a `patchChromedriver` script for AppVeyor setup in the dev branch.

packages/@vue/cli-plugin-e2e-nightwatch/nightwatch.config.js
```diff
@@ -1,36 +1,98 @@
...
+        browserName: 'firefox',
+        alwaysMatch: {
+          acceptInsecureCerts: true,
+          'moz:firefoxOptions': {
+            args: geckoArgs
+          }
+        }
+      },
+      webdriver: useSelenium ? {} : {
+        server_path: require('geckodriver').path,
```
sodatea: If firefox is supported by default, then we need to install `geckodriver` for users (in `packages/@vue/cli-plugin-e2e-nightwatch/generator/index.js`).
Otherwise, there should be a try catch here.

Personally I think `geckodriver` could be an [optional peer dependency](https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md). As it's not that widely used and its size is not negligible (2MB-5MB depending on OS).

packages/@vue/cli-plugin-e2e-nightwatch/package.json
```diff
@@ -26,14 +26,16 @@
     "@vue/cli-shared-utils": "^4.0.0-rc.1",
     "deepmerge": "^3.2.0",
     "execa": "^1.0.0",
-    "nightwatch": "^1.1.11",
-    "selenium-server": "^3.141.59"
+    "nightwatch": "^1.2.2"
   },
   "devDependencies": {
-    "chromedriver": "^74.0.0"
+    "selenium-server": "^3.141.59",
```
sodatea: It's an optional peer dependency considering it's required in the config file.

packages/@vue/cli-plugin-e2e-nightwatch/README.md
```diff
@@ -6,29 +6,92 @@
...
+
+For now, this is only available in Chromedriver. Read more about [parallel running](https://nightwatchjs.org/guide/running-tests/#parallel-running) in the Nightwatch docs.
+
+```sh
+$ vue-cli-service test:e2e --parallel
+```
+
+**Running with Selenium**
+
+In most cases, running with Selenium is not required since Nightwatch v1.0. It is still possible with the following:
```
sodatea: Should suggest that users need to install `selenium-server` themselves.

packages/@vue/cli-plugin-e2e-nightwatch/nightwatch.config.js
beatfactor: I think I missed that. Regarding distributing gecko by default, I suppose that is your decision to make. I am not entirely up to date with how widely used Firefox is in the wider dev/testing community, but I believe geckodriver is the most W3C Webdriver Api compliant browser driver and also, as far as I can tell, it is being developed by the team who originally built Webdriver. Besides, Firefox is an independent browser with a strong focus on privacy and security, and I think its values align well with Vue's.

sodatea: That makes sense.
We can include this by default for now. Later when we implement the prompt interface, we can check for the user's installed browsers and ask them which webdriver they are going to use.

sodatea:
The rest of this PR looks pretty cool to me. I'll merge it after the dependency issues get fixed.
Thanks for your help to make Nightwatch a breeze to use in Vue CLI!

beatfactor:
> The rest of this PR looks pretty cool to me. I'll merge it after the dependency issues get fixed.

Great, so to summarise, changes required:
- move `selenium-server` to optional peerDependencies and add a note in the Readme regarding the installation
- add geckodriver inside `generator/index.js` as a devDependency

Anything else I might have missed? 

> Thanks for your help to make Nightwatch a breeze to use in Vue CLI!

My pleasure. Feel free to reach out regarding Nightwatch related bugs and/or enhancements, we'd be happy to assist.

sodatea:
Yeah, I think that's all.

packages/@vue/cli-plugin-e2e-nightwatch/package.json
```diff
@@ -29,10 +29,12 @@
     "nightwatch": "^1.2.2"
   },
   "devDependencies": {
-    "selenium-server": "^3.141.59",
     "chromedriver": "^76.0.1",
     "geckodriver": "^1.16.2"
   },
+  "optionalDependencies": {
```
sodatea: Optional peer dependencies are listed in "peerDependencies" (so that they won't be installed when running `npm install`) along with a "peerDependenciesMeta" field (to suppress the "missing peer dependencies" warning).

https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-optional-peer-dependencies.md

```
{
  "peerDependencies": {
    "selenium-server": "^3.141.59"
  },
  "peerDependenciesMeta": {
    "selenium-server": {
      "optional": true
    }
  }
}
```

"optionalDependencies" will still be installed unless users explicitly specify the `--no-optional` flag.

beatfactor: Cool, thanks for the details.

sodatea:
Tests are failing.
So `selenium-server` still needs to be added to `devDependencies`…

packages/@vue/cli-plugin-e2e-nightwatch/generator/template/tests/e2e/specs/test.js
```diff
@@ -1,14 +1,21 @@
...
 module.exports = {
   'default e2e tests': browser => {
     browser
-      .url(process.env.VUE_DEV_SERVER_URL)
-      .waitForElementVisible('#app', 5000)
+      .init()
+      .waitForElementVisible('#app')
       .assert.elementPresent('.hello')
-      .assert.containsText('h1', 'Welcome to Your Vue.js <%- hasTS ? '+ TypeScript ' : '' %>App')
+      .assert.containsText('h1', 'Welcome to Your Vue.js App')
```
sodatea:
```suggestion
      .assert.containsText('h1', 'Welcome to Your Vue.js <%- hasTS ? '+ TypeScript ' : '' %>App')
```

This is required to pass the typescript plugin's e2e test.

packages/@vue/cli-plugin-e2e-nightwatch/__tests__/nightwatchPlugin.spec.js
```diff
@@ -1,13 +1,100 @@
...
+    let results = await project.read('test_results.json')
+    results = JSON.parse(results)
+
+    let testSettings = await project.read('test_settings.json')
+    testSettings = JSON.parse(testSettings)
+
+    expect(testSettings.parallel_mode).toStrictEqual(true)
+    expect(testSettings.test_workers).toStrictEqual(true)
+
+    expect(Object.keys(results.modules)).toEqual([
```
sodatea: Now that the tests are run in parallel, the order of `results.modules` is nondeterministic, which could cause test failures.
```suggestion
    expect(Object.keys(results.modules.sort())).toEqual([
```

PR #4565, opened 2019-09-11
refactor!: [wip] make workbox-webpack-plugin a peer dependency

sodatea:
Update: I'm very hesitant about this refactor. Will give it another look next week.

TODOs:
- [ ] implement a migrator for `vue upgrade` to automate the migration
- [ ] ~~implement a `postinstall` script for prerelease users who don't use `vue upgrade` to upgrade this plugin.~~ Adding a `postinstall` script does not make sense as we've got migrators already. I think adding a strong and explicit is enough.

Considering workbox team is preparing a v5 release.
It's likely that in the early-to-middle stage of Vue CLI v4 lifecycle that a major version bump of workbox may happen.
By making it a peer dep we could have more flexibility on version strategies.

We should do the similar to the unit-jest plugin.
IMO these are our last two remaining issues before an official v4 release.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [x] Yes
- [ ] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-plugin-pwa/generator/index.js
+++ b/packages/@vue/cli-plugin-pwa/generator/index.js
@@ -2,6 +2,9 @@ module.exports = api => {
   api.extendPackage({
     dependencies: {
       'register-service-worker': '^1.6.2'
+    },
+    devDependencies: {
+      'workbox-webpack-plugin': '^4.3.1'
     }
   })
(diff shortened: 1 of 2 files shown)


sodatea:
Closing this because separating such an essential dependency out could be a much larger maintenance burden to us.

PR #4583, opened 2019-09-16
docs: Add common use case scenario for Vue CLI lint command

bencodezen:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [ ] No

**Other information:**

### DIFF
--- a/docs/guide/cli-service.md
+++ b/docs/guide/cli-service.md
@@ -92,6 +92,32 @@ There are a few useful flags:
 
 - `--report` and `--report-json` will generate reports based on your build stats that can help you analyze the size of the modules included in your bundle.
 
+## vue-cli-service lint
+
+```
+Usage: vue-cli-service lint [options] [...files]
+
+Options:
+
+  --format [formatter]   specify formatter (default: codeframe)
+  --no-fix               do not fix errors or warnings
+  --no-fix-warnings      fix errors, but do not fix warnings
+  --max-errors [limit]   specify number of errors to make build failed (default: 0)
+  --max-warnings [limit] specify number of warnings to make build failed (default: Infinity)
+```
+
+### What if my project is not in the same directory as the Vue CLI scaffold?
+
+For those looking to specify direcotry will be impacted by their `vue-cli-service lint` command, you can pass it a regex pattern for whichever directory you desire. For example:
+
+1. Vue CLI projects lives in `/src/webapp/`
+2. `package.json` lives in root directory (`/`)
+
+You will need to define your lint command as:
+
+`"lint": "vue-cli-service lint ./src/webapp/**/*.js ./src/webapp/**/*.vue`
+
+
 ## vue-cli-service inspect
 
 ```


docs/guide/cli-service.md
```diff
@@ -92,6 +92,32 @@ There are a few useful flags:
 
 - `--report` and `--report-json` will generate reports based on your build stats that can help you analyze the size of the modules included in your bundle.
 
+## vue-cli-service lint
```
NataliaTepluhina: **concern**: I think this section of the CLI docs describes only scripts available with the basic setup, i.e. without any CLI plugins included. However, `lint` script is added only with `@vue/cli-plugin-eslint`. My suggestion would be to move this part to respective docs (i.e. `@vue/cli-plugin-eslint` documentation).

Also, PRs to docs should be created using `master` or `next` branch as a base (see [Contributing guide](https://github.com/vuejs/vue-cli/blob/dev/.github/CONTRIBUTING.md)).

In `master` ESLint plugin docs are located [here](https://github.com/vuejs/vue-cli/blob/master/packages/%40vue/cli-plugin-eslint/README.md)

In `next` we changed them to be the part of the [main documentation](https://github.com/vuejs/vue-cli/blob/next/docs/core-plugins/eslint.md) rather than standalone READMEs.

docs/guide/cli-service.md
```diff
@@ -92,6 +92,32 @@ There are a few useful flags:
...
+  --format [formatter]   specify formatter (default: codeframe)
+  --no-fix               do not fix errors or warnings
+  --no-fix-warnings      fix errors, but do not fix warnings
+  --max-errors [limit]   specify number of errors to make build failed (default: 0)
+  --max-warnings [limit] specify number of warnings to make build failed (default: Infinity)
+```
+
+### What if my project is not in the same directory as the Vue CLI scaffold?
+
+For those looking to specify direcotry will be impacted by their `vue-cli-service lint` command, you can pass it a regex pattern for whichever directory you desire. For example:
```
NataliaTepluhina: Small typo 😅 

```suggestion
For those looking to specify directory will be impacted by their `vue-cli-service lint` command, you can pass it a regex pattern for whichever directory you desire. For example:
```

bencodezen: Thanks!

docs/guide/cli-service.md
bencodezen: Whoa that's weird. I just used the "Edit" link on the page to do this. That's interesting that it made the destination vuejs:docs rather than master (which is what I assumed).

Shall I just make the PR for `next` then?

NataliaTepluhina: Oh, that's a great catch! I will fix the Vuepress links to the repo! 🙇‍♀ 

Yes, I think a PR for `next` should work 😉

PR #4597, opened 2019-09-19
refactor: move jest configs into presets

sodatea:
This could significantly reduce the size of default jest boilerplate,
and gives us more control on the internal config details, thus
potentially making future upgrades easier.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [x] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


sodatea:
Instead of adding a standalone `@vue/jest-preset` package, I chose to bundle the presets inside the plugin itself, mostly for easier dependency management, including peer dependency resolutions and compatability with yarn pnp & pnpm.
Unfortunately, the preset paths now become very long: `'@vue/cli-plugin-unit-jest/preset/typescript-and-babel'`. But as most users won't manually type this path, I think it's an acceptable tradeoff.

PR #4635, opened 2019-09-30
feat: names for local plugins

zcuric:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
This PR tries to simplify getting plugin options when developing local plugins, utilising both the`vuePlugins.service` field in the `package.json` and `pluginOptions` in the `vue.config.js`.
Using `Object` instead of an `Array` for plugin registration, for example: 
```json
{
  "vuePlugins": {
    "service": {
      "myCommands": "my-commands.js"
    }
  }
}
```
key `myCommands` becomes the plugin `name`. If that same `name` is used as key in `pluginOptions` field, for example: 
```js
{
...
  "pluginOptions": {
    "myCommands": {
      "foo": "bar",
      "baz": "quz"
    }
  }
...
}
```
options for that plugin will become available as the third argument in the `PluginAPI`, for example: 
```js
// my-commands.js
module.exports = (api, options, { foo, baz }) => {
  console.log(foo) // bar
  console.log(baz) // quz
}
```

Also, this PRs tests now cover registration of local plugins in both cases.

### DIFF
--- a/docs/dev-guide/plugin-dev.md
+++ b/docs/dev-guide/plugin-dev.md
@@ -456,7 +456,7 @@ or
 When using the object configuration, key `myCommands` becomes your local plugin `name` and the plugin options will be directly accessible as the third argument in `PluginAPI` if the plugin `name` corresponds to the key defined in `pluginOptions`.
 
 `pluginOptions` are defined in `vue.config.js` or in `vue` field in `package.json`:
-```json
+```js
 {
   pluginOptions: {
     myPlugin: {
--- a/docs/dev-guide/plugin-dev.md
+++ b/docs/dev-guide/plugin-dev.md
@@ -432,6 +432,49 @@ module.exports.defaultModes = {
 
 This is because the command's expected mode needs to be known before loading environment variables, which in turn needs to happen before loading user options / applying the plugins.
 
+### Local plugins
+
+If you need access to the plugin API in your project and don't want to create a full plugin for it, you can use the `vuePlugins.service` option in your `package.json` file:
+
+```json
+{
+  "vuePlugins": {
+    "service": ["my-local-plugin.js"]
+  }
+}
+```
+or
+```json
+{
+  "vuePlugins": {
+    "service": {
+      "myPlugin": "my-local-plugin.js"
+    }
+  }
+}
+```
+When using the object configuration, key `myCommands` becomes your local plugin `name` and the plugin options will be directly accessible as the third argument in `PluginAPI` if the plugin `name` corresponds to the key defined in `pluginOptions`.
+
+`pluginOptions` are defined in `vue.config.js` or in `vue` field in `package.json`:
+```json
+{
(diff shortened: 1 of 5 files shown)


docs/dev-guide/plugin-dev.md
```diff
@@ -432,6 +432,49 @@ module.exports.defaultModes = {
...
+```json
+{
+  "vuePlugins": {
+    "service": {
+      "myPlugin": "my-local-plugin.js"
+    }
+  }
+}
+```
+When using the object configuration, key `myCommands` becomes your local plugin `name` and the plugin options will be directly accessible as the third argument in `PluginAPI` if the plugin `name` corresponds to the key defined in `pluginOptions`.
```
pksunkara: There's no context for `myCommands` here.

zcuric: You are correct.

zcuric:
> I don't understand why local plugins need a name when you can simply use `local:${file}`. 
To avoid: `local:plugins/my-super-cool-plugin`.

> You also added an unnecessary `options` to PluginAPI. 
Have your read the reasoning I've added this **optional** `options`? It's there to provide easier access to plugin options.

pksunkara:
> To avoid: `local:plugins/my-super-cool-plugin`.

Yeah, could you please expand on that a bit more? What were the issues you are having and how does this fix it? Etc..

zcuric:
> > To avoid: `local:plugins/my-super-cool-plugin`.
> 
> Yeah, could you please expand on that a bit more? What were the issues you are having and how does this fix it? Etc..

It provides easier access to plugins options if plugin is named. I found it cumbersome to extract the plugin options from the `service` when developing plugins. `poi` has this solved nicely by providing directly plugin options as an argument, so this was inspired by it. 

I wasn't having any issues, I just wanted to make things a bit more convenient for plugin developers. 
This doesn't have to be merged, it's just my proposal, although PR has some neat tests that cover existing functionality.

pksunkara:
> It provides easier access to plugins options if plugin is named. I found it cumbersome to extract the plugin options from the service when developing plugins. poi has this solved nicely by providing directly plugin options as an argument, so this was inspired by it.

Yeah, we can provide the pluginOptions as an argument. But, it should be done all kinds of plugins, not only local ones.

It still doesn't change the fact that we don't have to add support for `name` for local plugins. The code can be changed:

```diff
-      const options = pluginOptions[name] || {}
+      const options = pluginOptions[id] || {}
```

zcuric:
> it still doesn't change the fact that we don't have to add support for name for local plugins. The code can be changed

I see your point. I'll update.

zcuric:
@pksunkara Closing this one. I'll make a new PR.

PR #4677, opened 2019-10-11
fix: add pnpm v4 support

B4rtware:
in pnpm v4 the option '--loglevel' is no longer available instead '--reporter' is used.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
`--reporter silent` should replicate `--loglevel error`
and `--shamefully-flatten` was renamed to `--shamefully-hoist`

### DIFF
--- a/packages/@vue/cli-shared-utils/lib/env.js
+++ b/packages/@vue/cli-shared-utils/lib/env.js
@@ -98,7 +98,7 @@ function getPnpmVersion () {
   return _pnpmVersion || '0.0.0'
 }
 
-exports.hasPnpmXOrLater = (version) => {
+function hasPnpmVersionOrLater (version) {
   if (process.env.VUE_CLI_TEST) {
     return true
   }
@@ -106,6 +106,14 @@ exports.hasPnpmXOrLater = (version) => {
     return semver.gte(_pnpmVersion, version)
   }
   return semver.gte(getPnpmVersion(), version)
+}
+
+exports.hasPnpm3OrLater = () => {
+  return hasPnpmVersionOrLater('3.0.0')
+}
+
+exports.hasPnpm4OrLater = () => {
+  return hasPnpmVersionOrLater('4.0.0')
 }
 
 exports.hasProjectPnpm = (cwd) => {
@@ -120,7 +128,7 @@ exports.hasProjectPnpm = (cwd) => {
 }
 
 function checkPnpm (result) {
-  if (result && !exports.hasPnpmXOrLater('3.0.0')) {
+  if (result && !exports.hasPnpm3OrLater()) {
     throw new Error(`The project seems to require pnpm${_hasPnpm ? ' >= 3' : ''} but it's not installed.`)
   }
   return result
--- a/packages/@vue/cli-shared-utils/lib/env.js
+++ b/packages/@vue/cli-shared-utils/lib/env.js
@@ -97,10 +97,11 @@ function checkPnpmVersion () {
     _hasPnpm = true
     _hasPnpm3orLater = semver.gte(pnpmVersion, '3.0.0')
     _hasPnpm4orLater = semver.gte(pnpmVersion, '4.0.0')
-    return [_hasPnpm3orLater, _hasPnpm4orLater]
   } catch (e) {
-    return [_hasPnpm3orLater = false, _hasPnpm4orLater = false]
+    _hasPnpm3orLater = false
+    _hasPnpm4orLater = false
   }
+  return [_hasPnpm3orLater, _hasPnpm4orLater]
 }
 
 exports.hasPnpm3OrLater = () => {
--- a/packages/@vue/cli-shared-utils/lib/env.js
+++ b/packages/@vue/cli-shared-utils/lib/env.js
@@ -84,7 +84,7 @@ const _pnpmProjects = new LRU({
   max: 10,
   maxAge: 1000
 })
-// remove
+
 function getPnpmVersion () {
   if (_pnpmVersion != null) {
     return _pnpmVersion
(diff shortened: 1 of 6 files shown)


packages/@vue/cli-shared-utils/lib/env.js
```diff
@@ -107,6 +108,25 @@ exports.hasPnpm3OrLater = () => {
   }
 }
 
+exports.hasPnpm4OrLater = () => {
+  if (process.env.VUE_CLI_TEST) {
+    return true
+  }
+  if (_hasPnpm3orLater != null) {
```
pksunkara:
```suggestion
  if (_hasPnpm4orLater != null) {
```

B4rtware:
> Maybe it's better if you read pnpm version and set all the flags at once instead of so much duplicated code

You are right! I will try to find a solution.

packages/@vue/cli-shared-utils/lib/env.js
```diff
@@ -101,10 +96,32 @@ exports.hasPnpm3OrLater = () => {
...
+}
+
+exports.hasPnpm3OrLater = () => {
+  if (process.env.VUE_CLI_TEST) {
+    return true
+  }
+  if (_hasPnpm3orLater != null) {
+    return _hasPnpm3orLater
+  }
+  return checkPnpmVersion()[0]
```
pksunkara:
```suggestion
  checkPnpmVersion()
  return _hasPnpm3orLater
```

packages/@vue/cli-shared-utils/lib/env.js
```diff
@@ -101,10 +96,32 @@ exports.hasPnpm3OrLater = () => {
...
+}
+
+exports.hasPnpm4OrLater = () => {
+  if (process.env.VUE_CLI_TEST) {
+    return true
+  }
+  if (_hasPnpm4orLater != null) {
+    return _hasPnpm4orLater
   }
+  return checkPnpmVersion()[1]
```
pksunkara:
```suggestion
  checkPnpmVersion()
  return _hasPnpm4orLater
```

B4rtware: I am not really sure if I now can also remove the return value from checkPnpmVersion(). Will it be used for testing?

pksunkara: @sodatea What do you think?

sodatea: Why not just cache the PNPM version and move the semver logic into the two exported functions?
```js
exports.hasPnpm4OrLater = () => {
  // ...
  return semver.gte(getPnpmVersion() || '0.0.0', '4.0.0')
}
```

B4rtware: Currently I don't see a benefit of caching the version number. Wouldn't that introduce more function calls or am I overseeing something?  What do you think?

sodatea: 1. While you name the function as `checkPnpmVersion`, it only does the job of `checkPnpmVersion3or4`. What if we want to add a `hasPnpm5OrLater` function? The current implementation is a little bit obscure and not very extensible in the long run.
2. The cost of a normal function call is negligible in most cases. What's not negligible is the cost of forking a process, namely the `execSync` call, which may take a few hundred milliseconds to finish. So we need to avoid calling `execSync` multiple times, but no need to worry about other functions.

B4rtware: Indeed good points. I have implemented the caching now. I have also removed the two functions `hasPnpm3/4OrLater` which can be now replicated by using the `hasPnpmXOrLater(version)` function.

packages/@vue/cli-shared-utils/lib/env.js
```diff
@@ -79,32 +79,33 @@ exports.hasProjectGit = (cwd) => {
...
+  } catch (e) {}
+  return _pnpmVersion || '0.0.0'
+}
+
+exports.hasPnpmXOrLater = (version) => {
+  if (process.env.VUE_CLI_TEST) {
+    return true
+  }
+  if (_pnpmVersion != null) {
+    return semver.gte(_pnpmVersion, version)
```
sodatea: This logic should be moved to `getPnpmVersion` because of separation of concerns. The `_pnpmVersion` cache is an implementation detail of `getPnpmVersion()`.

And separation of concerns *is* required here, because, if not for such consideration, the two functions can essentially be merged into one.

packages/@vue/cli-shared-utils/lib/env.js
```diff
@@ -79,32 +79,33 @@ exports.hasProjectGit = (cwd) => {
...
     _hasPnpm = true
-    _hasPnpm3orLater = semver.gte(pnpmVersion, '3.0.0')
-    return _hasPnpm3orLater
-  } catch (e) {
-    return (_hasPnpm3orLater = false)
+  } catch (e) {}
+  return _pnpmVersion || '0.0.0'
+}
+
+exports.hasPnpmXOrLater = (version) => {
```
sodatea: While a more generic function is appreciated, don't remove the old exports…
Otherwise, you are changing the public API of this package, meaning this commit would be a breaking change.

sodatea: Also, the parameter is a little bit confusing here.
If the function name has `X` in it, I would expect it to accept a version number (`hasPnpmXOrLater(3)`), while it requires a full version string (`hasPnpmXOrLater('3.0.0')`)…

B4rtware: Sorry for the inconveniences. I have corrected your concerns. `hasPnpm3OrLater` is now a wrapper for the `hasPnpmVersionOrLater` function with appropriate version number.

B4rtware: I don't know why but somehow github doesn't want to add my latest commit ~6h ago into this merge request.

PR #4681, opened 2019-10-15
chore!: add `@vue/cli` in `--version` output, to avoid confusion

### LINKED ISSUE #4514: Make vue --version print both the CLI version and the current version of Vue in use in the current directory
Currently `vue --version` will show the version of the CLI. It does not show the version of Vue being used in the current directory.

This is potentially confusing if someone ran `vue --version` which told them 3.11.0 causing them to think they are using Vue 3.

NA

sodatea:
Sorry for the last-minute breaking change. But I do think it is important to make the branding clearer.
Including the name in the output is also a common practice in command-line tools,
such as `bash`, `brew`, etc. And this change only requires minimal effort for other tools to get the version number from the new output format.

Closes #4514

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [x] Yes
- [ ] No

**Other information:**

### DIFF
--- a/packages/@vue/cli/bin/vue.js
+++ b/packages/@vue/cli/bin/vue.js
@@ -21,7 +21,7 @@ function checkNodeVersion (wanted, id) {
   }
 }
 
-checkNodeVersion(requiredVersion, 'vue-cli')
+checkNodeVersion(requiredVersion, '@vue/cli')
 
 if (semver.satisfies(process.version, '9.x')) {
   console.log(chalk.red(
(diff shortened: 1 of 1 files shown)


DRoet:
Wouldn't `@vue/cli` be the correct branding?

sodatea:
Yeah you are right 😂
Changed!

PR #4757, opened 2019-10-25
chore: dependency maintenance

janwillemvd:
* chore: dependency maintenance (upgrade cypress to 3.5.x)

**What kind of change does this PR introduce?**
- [x] Underlying tools: upgrade cypress to ^3.5.0

**Does this PR introduce a breaking change?** 
- [ ] Yes
- [x] No


sodatea:
Thanks for the help, but since this is a minor version, there's no need to explicitly bump it alone. (We do periodic batch upgrades)

PR #4787, opened 2019-11-02
types(router): added router array type for Array RouteConfig

manuelojeda:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [x] Other, please describe:
Added a conditional in the generator of router when **TypeScript** is selected, added the routes array as **<Array<RouteConfig>>**, when using the **@vue/composition-api** using the routes without the type may return an error of incompatibility

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-plugin-router/generator/template/src/router/index.js
+++ b/packages/@vue/cli-plugin-router/generator/template/src/router/index.js
@@ -1,9 +1,35 @@
 import Vue from 'vue'
+<%_ if (hasTypeScript) { _%>
+import VueRouter, { RouteConfig } from 'vue-router'
+<%_ } else { _%>
 import VueRouter from 'vue-router'
+<%_ } _%>
 import Home from '../views/Home.vue'
 
-Vue.use(VueRouter)
+<%_ if (hasTypeScript) { _%>
+const routes = <Array<RouteConfig>>[
 
+  {
+    path: '/',
+    name: 'home',
+    component: Home
+  },
+  {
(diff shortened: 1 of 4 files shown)


packages/@vue/cli-plugin-router/generator/template/src/router/index.js
```diff
@@ -1,9 +1,35 @@
...
+<%_ if (hasTypeScript) { _%>
+import VueRouter, { RouteConfig } from 'vue-router'
+<%_ } else { _%>
 import VueRouter from 'vue-router'
+<%_ } _%>
 import Home from '../views/Home.vue'
 
-Vue.use(VueRouter)
+<%_ if (hasTypeScript) { _%>
+const routes = <Array<RouteConfig>>[
```
sodatea: As far as I can see, only this line needs to be modified for typescript projects. So no need to  wrap the whole block in the `if` condition.

sodatea: Better use type annotation than type casting:
```suggestion
const routes: Array<RouteConfig> = [
```

manuelojeda: Adding this type annotation just bug the CLI and an error will popup

manuelojeda:
I'm closing this PR to make a new one with the correct changes without adding the lock files.

PR #4800, opened 2019-11-05
fix(serve): pass devServer sockPath properly to client

AlbertBrand:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

When you change the `devServer.sockPath` setting, one of the required `webpack-devserver-client` ignores the `sockPath` setting. This is visible in your developer console: the local sockjs url connects but the network sockjs fails.

This is caused by incorrectly providing the `sockPath` to the webpack client. It infers the host and port from the provided URL parameter but not the sockPath, which has to be specified as an additional parameter. This PR removes the path part from the `sockjsUrl` (as it does nothing) and adds the `sockPath` as additional param.

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -126,9 +126,10 @@ module.exports = (api, options) => {
 
     // inject dev & hot-reload middleware entries
     if (!isProduction) {
+      const sockPath = projectDevServerOptions.sockPath || '/sockjs-node'
       const sockjsUrl = publicUrl
         // explicitly configured via devServer.public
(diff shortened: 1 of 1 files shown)


McPo:
Whats the normal release schedule for vue-cli, is there anything in particular that would trigger a patch release? As Id like to know how long I can expect until this fix is released.

Cheers,
Emmet

sodatea:
Need to clean up some other stuff.

I'll ship a release (4.4.0, under the `next` dist-tag first) next Monday, and move it to `latest` around Wednesday.

PR #4805, opened 2019-11-07
types(router): added router array type for Array RouteConfig

manuelojeda:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
This PR adds the type annotation to the const routes in router/index.ts when vue-router and TypeScript are selected in the CLI.
❌If you use the @vue/composition-api with both, router and TS selected, and error will pop up in the console as a shown in this image: [https://i.imgur.com/NXugrHy.png](https://i.imgur.com/NXugrHy.png)

✅ After you add the annotation as stated in this PR there is no error shown in the console: 
[https://i.imgur.com/rq0S34U.png](https://i.imgur.com/rq0S34U.png)

**Note:** ❌ If you add the annotation type like `const routes:<Array<RouteConfig>>` an error will occur and won't accept the error:  [https://i.imgur.com/5gytWju.png](https://i.imgur.com/5gytWju.png)

### DIFF
--- a/packages/@vue/cli-plugin-router/generator/template/src/router/index.js
+++ b/packages/@vue/cli-plugin-router/generator/template/src/router/index.js
@@ -1,10 +1,16 @@
 import Vue from 'vue'
+<%_ if (hasTypeScript) { _%>
+import VueRouter, { RouteConfig } from 'vue-router'
+<%_ } else { _%>
 import VueRouter from 'vue-router'
+<%_ } _%>
 import Home from '../views/Home.vue'
 
-Vue.use(VueRouter)
-
-const routes = [
+<%_ if (hasTypeScript) { _%>
+  const routes = <Array<RouteConfig>>[
+<%_ } else { _%>
+  const routes = [
+<%_ } _%>
   {
     path: '/',
     name: 'home',
@@ -25,6 +31,8 @@ const routes = [
     <%_ } _%>
   }
 ]
+
+Vue.use(VueRouter)
 
 const router = new VueRouter({
   <%_ if (historyMode) { _%>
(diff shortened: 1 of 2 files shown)


sodatea:
Sorry but I can't reproduce the error locally.
Do you have an reproduction repo? Did you use `@vue/composition-api`? If so, it should be a bug in that repo.

manuelojeda:
Of course, here is a basic repo builded with the CLI with TypeScript and Vue Router, and only added the @vue/composition-api.
I found out that the erorr occurs when you use the Composition API in a view, in this example just added the API at **views/Home.vue**

Here is the repo:
https://github.com/manuelojeda/vue-ts-router-composition-api

masonk:
https://github.com/vuejs/composition-api/issues/185 is probably a dupe of this

packages/@vue/cli-plugin-router/generator/template/src/router/index.js
```diff
@@ -26,6 +32,8 @@ const routes = [
   }
 ]
 
+Vue.use(VueRouter)
```
sodatea: Thanks for the update!
One last small suggestion here: it might be easier to read if we put the `Vue.use` statement back to the beginning of the file, right after the `import` statements.

manuelojeda: Done, `Vue.use` is in place

sodatea:
Thanks!

PR #4816, opened 2019-11-08
fix: don't prepend publicPath with slash

### LINKED ISSUE #3338: Why use ensureSlash to force slash in front?
We set `publicPath` in the `vue.config.js`. But it will add a slash in the front when there is no `https?:`. In our usage, we will use a publishing tools to handle some STRING LIKE `__CDN__`. In webpack we only need to set publicPath to `__CDN__/`, but now we should set as `./__CDN__/`.

 I don't think the cli should check and auto change personal publicPath settings.

For compatibility, maybe add an option like `forceSlash` to switch the feature is better.

### LINKED ISSUE #4184: The builder prepend '/' on publicPath when using  custom protocol publicPath
### Version
3.8.4

### Environment info
```
System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i5-6200U CPU @ 2.30GHz
  Binaries:
    Node: 9.5.0 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 5.6.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0
    @vue/babel-plugin-transform-vue-jsx:  1.0.0
    @vue/babel-preset-app:  3.8.0
    @vue/babel-preset-jsx:  1.0.0
    @vue/babel-sugar-functional-vue:  1.0.0
    @vue/babel-sugar-inject-h:  1.0.0
    @vue/babel-sugar-v-model:  1.0.0
    @vue/babel-sugar-v-on:  1.0.0
    @vue/cli-overlay:  3.8.0
    @vue/cli-plugin-babel: ^3.8.0 => 3.8.0
    @vue/cli-plugin-eslint: ^3.8.0 => 3.8.0
    @vue/cli-service: ^3.8.0 => 3.8.4
    @vue/cli-shared-utils:  3.8.0
    @vue/component-compiler-utils:  2.6.0
    @vue/preload-webpack-plugin:  1.1.0
    @vue/web-component-wrapper:  1.2.0
    eslint-plugin-vue: ^5.0.0 => 5.2.2 (4.7.1)
    vue: ^2.6.10 => 2.6.10
    vue-eslint-parser:  2.0.3 (5.0.0)
    vue-hot-reload-api:  2.3.3
    vue-loader:  15.7.0
    vue-style-loader:  4.1.2
    vue-template-compiler: ^2.6.10 => 2.6.10
    vue-template-es2015-compiler:  1.9.1
  npmGlobalPackages:
    @vue/cli: Not Found
```

### Steps to reproduce
```js
// vue.config.js
module.exports = {
  publicPath: 'customprotocol://www.my-app.com/'
}

```

### What is expected?
`<link href=customprotocol://www.my-app.com/js/app.3a07a650.js rel=preload as=script>`

`<link href=/customprotocol://www.my-app.com/js/app.3a07a650.js rel=preload as=script>`

---
In our APP, we will intercept the assets with customprotocol and get them from local cache. Please provide an option to allow using publicPath not starting with `https?`.

sodatea:
fixes #3338
fixes #4184

Actually I don't know why the slash was added in the first place, seems
extraneous to me.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-service/__tests__/Service.spec.js
+++ b/packages/@vue/cli-service/__tests__/Service.spec.js
@@ -104,6 +104,16 @@ test('normalize publicPath when relative', () => {
   expect(service.projectOptions.publicPath).toBe('foo/bar/')
 })
 
+test('allow custom protocol in publicPath', () => {
+  mockPkg({
+    vue: {
+      publicPath: 'customprotocol://foo/bar'
+    }
+  })
+  const service = createMockService()
+  expect(service.projectOptions.publicPath).toBe('customprotocol://foo/bar/')
+})
+
 test('keep publicPath when empty', () => {
   mockPkg({
     vue: {
--- a/packages/@vue/cli-service/lib/Service.js
+++ b/packages/@vue/cli-service/lib/Service.js
@@ -395,11 +395,8 @@ module.exports = class Service {
 }
 
 function ensureSlash (config, key) {
-  let val = config[key]
+  const val = config[key]
   if (typeof val === 'string') {
-    if (!/^https?:/.test(val)) {
-      val = val.replace(/^([^/.])/, '/$1')
-    }
     config[key] = val.replace(/([^/])$/, '$1/')
   }
 }


tjam0588:
Was this PR even reviewed?  It broke our app.  Thanks! *smh*

sodatea:
How? Any reproductions?

tjam0588:
How? Something was there. You removed it. Yes, the slash seemed extraneous, but I've got to imagine we aren't the only ones who had a process that took its existence into account. These are the kinds of changes that make people nervous about upgrading.

sodatea:
I'm **very sorry** for the inconvenience caused by this commit.

In theory, the old behavior is undocumented, unintended, and non-intuitive, so I didn't expect it to be depended on by any project.

And we've put this in the beta channel for 18 days before a stable release, so I really didn't intend it to break any production app. I'm deeply sorry for that.

Anyway, in retrospect,
1. Such insignificant fixes should be included in a major release rather than a patch, considering **every bug fix could be a breaking change to some users**. 
2. I'm considering using the `~` version range for generated projects by default in the next major (or maybe minor, I haven't discussed with other team members yet) version of Vue CLI.
  Only fixes for regressions can be added to subsequent patch releases.
  Other bug fixes should be included in minor versions, and users need to run `vue upgrade` manually to upgrade.

What do you think?

tjam0588:
Thankfully, it didn't break the production app.  We caught it in our CI environment. It just caused a bit of panic as we scrambled to understand what had changed.

qiulang:
We accidentally deleted package-lock.json all of sudden our web app stopped to work! We then finally found it was because we didn't add leading '/' our vue.config.js 

My question is what about the '/' in the end ? The document here https://cli.vuejs.org/config/#publicpath seems to say we need the '/' in the end. But thru our test we found we don't.

Issue #4544, opened 2019-09-05
"vue create ." overrides existing package.json


Joe-Kerr:
### Version
3.11.0

### Environment info
```
Environment Info:
  Windows
  Binaries:
    Node: 10.15.1 - node.EXE
    Yarn: Not Found
    npm: 6.1.0 - npm.CMD
  npmPackages:
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0 
    @vue/babel-plugin-transform-vue-jsx:  1.0.0 
    @vue/babel-preset-app:  3.11.0 
    @vue/babel-preset-jsx:  1.1.0 
    @vue/babel-sugar-functional-vue:  1.0.0 
    @vue/babel-sugar-inject-h:  1.0.0 
    @vue/babel-sugar-v-model:  1.0.0 
    @vue/babel-sugar-v-on:  1.1.0 
    @vue/cli-overlay:  3.11.0 
    @vue/cli-plugin-babel: ^3.11.0 => 3.11.0 
    @vue/cli-plugin-pwa: ^3.11.0 => 3.11.0 
    @vue/cli-plugin-unit-mocha: ^3.11.0 => 3.11.0 
    @vue/cli-service: ^3.11.0 => 3.11.0 
    @vue/cli-shared-utils:  3.11.0 
    @vue/component-compiler-utils:  3.0.0 
    @vue/preload-webpack-plugin:  1.1.1 
    @vue/test-utils: 1.0.0-beta.29 => 1.0.0-beta.29 
    @vue/web-component-wrapper:  1.2.0 
    vue: ^2.6.10 => 2.6.10 
    vue-hot-reload-api:  2.3.3 
    vue-loader:  15.7.1 
    vue-style-loader:  4.1.2 
    vue-template-compiler: ^2.6.10 => 2.6.10 
    vue-template-es2015-compiler:  1.9.1 
    vuex: ^3.0.1 => 3.1.1 
  npmGlobalPackages:
    @vue/cli: Not Found 
```

### Steps to reproduce
- npm init
- vue create .

### What is expected?
merge package.json

### What is actually happening?
override package.json

Joe-Kerr:
The contents of the existing nodes_modules folder were removed as well :(

PR #4827, opened 2019-11-11
feat: respect existing package.json

zyy7259:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

This pull request will close https://github.com/vuejs/vue-cli/issues/4544

### DIFF
--- a/packages/@vue/cli-service/lib/Service.js
+++ b/packages/@vue/cli-service/lib/Service.js
@@ -1,14 +1,13 @@
 const fs = require('fs')
 const path = require('path')
 const debug = require('debug')
-const readPkg = require('read-pkg')
 const merge = require('webpack-merge')
 const Config = require('webpack-chain')
 const PluginAPI = require('./PluginAPI')
 const dotenv = require('dotenv')
 const dotenvExpand = require('dotenv-expand')
 const defaultsDeep = require('lodash.defaultsdeep')
-const { chalk, warn, error, isPlugin, resolvePluginId, loadModule } = require('@vue/cli-shared-utils')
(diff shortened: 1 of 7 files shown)


packages/@vue/cli/package.json
```diff
@@ -48,6 +48,7 @@
     "lodash.clonedeep": "^4.5.0",
     "lru-cache": "^5.1.1",
     "minimist": "^1.2.0",
+    "read-pkg": "^5.1.1",
```
pksunkara: You don't need this dep. We have some tools that gets the pkg info.

zyy7259: I'm not sure which tool I should use. Currently I only found that [@vue/cli-service/lib/Service.js#L49](https://github.com/vuejs/vue-cli/blob/v4.1.1/packages/%40vue/cli-service/lib/Service.js#L49) uses `read-pkg` to load the `package.json`.

pksunkara: Yup, so move this resolvepkg function to utils package, and we can use it as common in both places instead of doing it like this.

zyy7259: That makes sense.

zyy7259:
@pksunkara hi, I have modified the code as you requested. Should I rebase the code every time when `dev` has newer commits?

pksunkara:
Yes, please rebase since we need to re trigger CI

PR #4621, opened 2019-09-26
feat: support custom package manager

zyy7259:
**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

**Other information:**

Turn `Unknown package manager` error into a warning and use the npm subcommands as fallbacks.

PR #4828, opened 2019-11-11
feat: add option merge to command create

zyy7259:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

- Usage: `vue create --merge .`
- This option works like the `--force` option.

### DIFF
--- a/packages/@vue/cli/bin/vue.js
+++ b/packages/@vue/cli/bin/vue.js
@@ -63,6 +63,7 @@ program
   .option('-g, --git [message]', 'Force git initialization with initial commit message')
   .option('-n, --no-git', 'Skip git initialization')
   .option('-f, --force', 'Overwrite target directory if it exists')
+  .option('--merge', 'Merge target directory if it exists')
   .option('-c, --clone', 'Use git clone when fetching remote preset')
   .option('-x, --proxy', 'Use specified proxy when creating project')
   .option('-b, --bare', 'Scaffold project without beginner instructions')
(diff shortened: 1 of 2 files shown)


sodatea:
I don't feel like adding this feature. Because the end result is still unpredictable and the directory could still get messed up.
A warning on the overriding might suffice.

zyy7259:
What I want to do with this option is auto generating projects by some tools based on vue-cli.

The tool I used needs to write some files before calling vue-cli.

I just want to skip the prompt, and merge files immediately.

pksunkara:
I agree with @sodatea here. If you want to, you can create a new folder and merge those 2 folders with some other tools locally.

zyy7259:
By default, without any cli options, if the target dir is existing, `@vue/cli` will prompts the user to either delete the target folder or merge with it.

And there is an option `--force` to skip the prompt and delete the target folder. So I just want to add an option `--merge` to skip the prompt and merge with the target folder.

Actually I'm writing a tool which combined `@vue/cli` with [a custom package manager](https://github.com/vuejs/vue-cli/pull/4621). I need to write some info into the `package.json` in the target folder first. And during the creation of the target project, [these info](https://github.com/vuejs/vue-cli/pull/4827) will be used by the custom package manager to installing deps.

So I just want this `--merge` option to skip the prompt and automate the whole creation procedure.

pksunkara:
That actually sounds better. If you are okay, we can merge this @sodatea

pksunkara:
This is still failing. You think there's an issue with your code?

```
FAIL packages/@vue/cli/__tests__/Upgrader.spec.js
```

pksunkara:
Never mind, it has been happening on `dev` too

PR #4842, opened 2019-11-13
Replace chalk.reset with stripAnsi in @vue/cli-shared-utils/lib/logger.js

perakerberg:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
When logging multi-line logs, the logger uses chalk.reset in order to get the length of the tag for the log. This is to achieve an indented layout of multi-line logs. However, chalk.reset only replaces styles such as **bold** and *italic*, not color, see [GitHub issue](https://github.com/chalk/chalk/issues/62). When the tag contains colors, the result is not pretty. Replacing with stripAnsi gets the intended result.

I found this behavior when using the GeneratorApi.exitLog function, when building a custom preset for vue-cli. Any multi-line entry is not well-formatted, and adding multiple exitLog calls gets rather noisy, since the logger adds the preset name as the preceding tag.

**Note**: This introduces a new require of "strip-ansi". If that is okay or not for vue-cli, I am not sure.

### DIFF
--- a/packages/@vue/cli-shared-utils/lib/logger.js
+++ b/packages/@vue/cli-shared-utils/lib/logger.js
@@ -1,4 +1,5 @@
 const chalk = require('chalk')
+const stripAnsi = require('strip-ansi');
 const readline = require('readline')
 const padStart = require('string.prototype.padstart')
 const EventEmitter = require('events')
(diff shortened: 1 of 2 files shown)


packages/@vue/cli-shared-utils/lib/logger.js
```diff
@@ -1,4 +1,5 @@
 const chalk = require('chalk')
+const stripAnsi = require('strip-ansi');
```
sodatea: Please remove the extraneous semicolon here, as we use the semicolon-less style in this codebase

perakerberg: Thanks for the helpful info, the semicolon is now removed!

PR #4843, opened 2019-11-14
fix(cli-service): fix duplicate webpack-dev-server/client

githoniel:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

WebpackDevServer will add client scripts to entry by default, result in duplicate `webpack-dev-server/client`.

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -164,6 +164,8 @@ module.exports = (api, options) => {
     const server = new WebpackDevServer(compiler, Object.assign({
       logLevel: 'silent',
       clientLogLevel: 'silent',
+      // 'Inline mode (set to false to disable including client scripts like livereload)',
+      inline: false,
       historyApiFallback: {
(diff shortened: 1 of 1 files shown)


pksunkara:
Can you describe a bit about why this is needed?

githoniel:
@pksunkara 
see https://github.com/webpack/webpack-dev-server/blob/v3.9.0/lib/utils/addEntries.js#L19

webpack-dev-server will add `dev-entry-client` and `hot-entry` when `options.inline = true`
in 3.9.0. Since vue-cli will add `dev-entry-client` and `hot-entry` manually, I set `options.inline = false` to avoid duplicate.

sodatea:
`inline: false` only switches to iframe mode. The correct fix is to set `injectClient: false`, see https://github.com/vuejs/vue-cli/pull/5069. It's been fixed in v4.2

PR #4887, opened 2019-11-26
warn only production builds output service workers

bseib:
Let people know that a side effect of specifying the mode determines what you'll get for a service worker. I was surprised when I deployed a development build to a private QA server and my `service-worker.js` file was missing.

My suggested changes can use some improvement. I'm hoping someone can help make it more thorough by listing what to expect for service worker output, for each mode.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


bseib:
More context [here](https://forum.vuejs.org/t/how-to-pass-env-var-to-app-build/80640) with respect to what was not understood in the docs.

PR #4910, opened 2019-12-01
fix: comment eslint disable in cypress config

cexbrayat:
The current cypress config has a global eslint-disable:

typescript
```
/* eslint-disable import/no-extraneous-dependencies, global-require, arrow-body-style */
// const webpack = require('@cypress/webpack-preprocessor')
```

But this assumes that `eslint-import-plugin` is present, whereas this is only true when the selected preset is the Airbnb one.
In other cases, this fails `yarn lint` right away with:

```
error: Definition for rule 'import/no-extraneous-dependencies' was not found (import/no-extraneous-dependencies) at tests/e2e/plugins/index.js
```

Even if using the Airbnb preset, the comment does not have any effect, as, by default, the `require` call is commented.

This commit now splits the rules disabling in two:

- one commented for the people using Airbnb and who will uncomment the require call below, as it is only useful for them.
- one activated that disables `arrow-body-style` as it will fail the linting for people using Airbnb preset with the current config.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-plugin-e2e-cypress/generator/template/tests/e2e/plugins/index.js
+++ b/packages/@vue/cli-plugin-e2e-cypress/generator/template/tests/e2e/plugins/index.js
@@ -1,3 +1,4 @@
+/* eslint-disable arrow-body-style */
 // https://docs.cypress.io/guides/guides/plugins-guide.html
 
 // if you need a custom webpack configuration you can uncomment the following import
@@ -5,7 +6,7 @@
 // as explained in the cypress docs
 // https://docs.cypress.io/api/plugins/preprocessors-api.html#Examples
 
-/* eslint-disable import/no-extraneous-dependencies, global-require, arrow-body-style */
+// /* eslint-disable import/no-extraneous-dependencies, global-require */
 // const webpack = require('@cypress/webpack-preprocessor')
 
 module.exports = (on, config) => {


sodatea:
The fix looks good to me. But how did you encounter that error? I can't reproduce it on my machine.

cexbrayat:
Sure, sorry, I forgot to provide the repro.

Create a project with:

    npx -p @vue/cli@next vue create eslint-issue --inlinePreset '{"useConfigFiles": true,"plugins": {"@vue/cli-plugin-typescript": {"classComponent": false},"@vue/cli-plugin-eslint": {"config": "prettier","lintOn": ["save"]}}}'

Change the dependencies to use latest eslint, plugins and presets:

```diff
"devDependencies": {
+  "@typescript-eslint/eslint-plugin": "2.10.0",
+  "@typescript-eslint/parser": "2.10.0",
  "@vue/cli-plugin-e2e-cypress": "^4.1.0",
  "@vue/cli-plugin-eslint": "^4.1.0",
  "@vue/cli-plugin-typescript": "^4.1.0",
  "@vue/cli-service": "^4.1.0",
-  "@vue/eslint-config-prettier": "^5.0.0",
-  "@vue/eslint-config-typescript": "^4.0.0",
-  "eslint": "^5.16.0",
-  "eslint-plugin-prettier": "^3.1.1",
-  "eslint-plugin-vue": "^5.0.0",
+  "@vue/eslint-config-prettier": "6.0.0",
+  "@vue/eslint-config-typescript": "5.0.1",
+  "eslint": "6.7.2",
+  "eslint-plugin-prettier": "3.1.1",
+  "eslint-plugin-vue": "6.0.1",
  "prettier": "^1.19.1",
  "typescript": "~3.5.3",
  "vue-template-compiler": "^2.6.10"
}
```

(It's in fact enough to just bump to the latest `eslint`).

Then install and run the linter:

    yarn lint

Throws:

    error: Definition for rule 'import/no-extraneous-dependencies' was not found (import/no-extraneous-dependencies) at tests/e2e/plugins/index.js

The fix should get rid of the issue for teams running the latest sets of eslint/plugins/config,
and avoid the default CLI project to run into it when upgrading these dependencies.

Issue #2850, opened 2018-10-31
Yarn PnP support


archSeer:
### What problem does this feature solve?
yarn 1.12 supports PnP: https://github.com/yarnpkg/rfcs/blob/master/accepted/0000-plug-an-play.md

### What does the proposed API look like?
I tried to get it to work, but the default project was missing a whole bunch of peer dependencies in the package.lock. Once that was resolved, the vue-cli webpack rules would have to be adjusted with:

https://github.com/arcanis/pnp-webpack-plugin

And for jest:

https://github.com/arcanis/jest-pnp-resolver

vielhuber:
This would be awesome for many usecases: Using vue-cli in Docker, Windows WSL or Vagrant currently is a pain (because of >100.000 mounted and synced files inside node_modules). Without this everything would be gone.

Do you consider this integrating in vue-cli (when choosing "yarn" inside the installation wizard)?

yuu2lee4:
yeah,i would glad see the supporting for pnp,i want know `transpileDependencies` how to defained

bodograumann:
I was trying to setup `pnp` for a basic `vue-cli` application. See https://github.com/bodograumann/yarn-pnp-app.

`pnp` is enabled and `php-webpack-plugin` is installed. This already fixed a lot of errors about module loading.

Now I’m at a point where I have no idea to proceed. Running `yarn build` gives the following error:

```
 ERROR  Failed to compile with 1 errors                                                                                                                                               14.42.06

This relative module was not found:

* ./src/main.js in multi ./src/main.js
 ERROR  Build failed with errors.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

Does anybody have an idea?

sodatea:
@bodograumann 
We currently don't have the capacity to fix all the issues with yarn pnp compatibility. But here's a note from @lbogdan who have successfully made a basic vue cli project work with yarn v2:
https://gist.github.com/lbogdan/b8c652e6fd6f906492e0adf0ee044c82
Hope it helps.

lbogdan:
@sodatea Would you be open to a PR that fixes (most of) the issues I encountered, while not breaking current functionality?

sodatea:
Yeah I would love to see that PR

hufeng:
I look forward to being able to support

hufeng:
@bodograumann Yes, I also encountered the same error

hufeng:
[...]

PR #4953, opened 2019-12-12
feat: adds transparent PnP support to Webpack

arcanis:

- [x] Feature
- [x] Underlying tools

- [x] No

**Other information:**

Followup to #2850 - I've fixed the remaining issues on our side, only remains one thing! 🎉

Webpack currently requires using the [PnP Webpack plugin](https://github.com/arcanis/pnp-webpack-plugin) in order to properly resolve dependencies (it will be optional starting from Webpack 6). It doesn't have [any effect](https://github.com/arcanis/pnp-webpack-plugin/blob/master/index.js#L116) for people that aren't using PnP, so no risks at all 🙂


yarn.lock
```diff
@@ -7045,7 +7045,7 @@ errorhandler@^1.5.1:
     accepts "~1.3.7"
     escape-html "~1.0.3"
 
-es-abstract@^1.12.0, es-abstract@^1.4.3, es-abstract@^1.5.1, es-abstract@^1.7.0:
```
arcanis: Those changes aren't needed by my diff, they just got there when running `yarn add` 🤔

packages/@vue/cli-service/lib/config/base.js
```diff
@@ -37,6 +37,10 @@ module.exports = (api, options) => {
         .publicPath(options.publicPath)
 
     webpackConfig.resolve
+      // This plugin can be removed once we switch to Webpack 6
+      .plugin('pnp')
+        .use({ ...require('pnp-webpack-plugin') })
```
arcanis: The spread is a small trick required for `webpack-chain` (it mutates the plugin instance, but since the PnP plugin is supposed to be the same instance for every config it crashes when webpack-chain tries to override its previous registration).

pksunkara:
Looks good to me. But I would wait for @sodatea to merge it.

pksunkara:
@arcanis Can you rebase this? I would like vue-cli to support the pnp feature. It's really good feature 😄

arcanis:
@pksunkara done! Let me know when it gets released and I'll setup our E2E tests accordingly to confirm everything is good 🙂

pksunkara:
I tested this locally.

PR #4954, opened 2019-12-12
make vue.config.js async, this will let pull param

gocmarek:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Please add this little change to your incredible job, it will make possibility to pull some data from external source to vue.config.js. In my case this will help a lot and I think not only for me.

I think this should not bring any security degradation or something like that.

### DIFF
--- a/packages/@vue/cli-service/lib/Service.js
+++ b/packages/@vue/cli-service/lib/Service.js
@@ -56,7 +56,7 @@ module.exports = class Service {
     }
   }
 
-  init (mode = process.env.VUE_CLI_MODE) {
+ async init (mode = process.env.VUE_CLI_MODE) {
     if (this.initialized) {
       return
     }
@@ -71,7 +71,7 @@ module.exports = class Service {
     this.loadEnv()
 
     // load user config
-    const userOptions = this.loadUserOptions()
+    const userOptions = await this.loadUserOptions()
     this.projectOptions = defaultsDeep(userOptions, defaults())
 
     debug('vue:project-config')(this.projectOptions)
@@ -216,7 +216,7 @@ module.exports = class Service {
     this.setPluginsToSkip(args)
 
     // load env variables, load user config, apply plugins
-    this.init(mode)
+    await this.init(mode)
 
     args._ = args._ || []
     let command = this.commands[name]
@@ -302,7 +302,7 @@ module.exports = class Service {
     return config
   }
 
-  loadUserOptions () {
+  async loadUserOptions () {
     // vue.config.js
     let fileConfig, pkgConfig, resolved, resolvedFrom
     const configPath = (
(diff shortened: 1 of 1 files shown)


gocmarek:
Guys, please do not omit this request ;)

LinusBorg:
This PR breaks all test groups. It doesn't update any tests or add new tests for this feature. We cannot merge this PR like this - a *lot* of work is missing.

gocmarek:
Hi,

Any other idea hot to add some data to project via vue.config.js when data needs to be pulled from database??
Any hints will be appreciated.

Regards,
Marek

simon300000:
One solution might be, to write a different script which save the result on some temporary file; execute this script with something like `child_process.execSync`, and use `readFileSync` to read the result in `vue.config.js`

gocmarek:
Hi, 
Thanks for advice, I already applied some kind of solution like
that.Anyway it would be very helpful to have possibility to inject
something before the app started and for this I wanted to have merged
this what I proposed.
Regards,Marek
-------- Wiadomość oryginalna ----------Od: simon3000 <
notifications@github.com>Odpowiedź do: vuejs/vue-cli <
reply@reply.github.com>Do:
vuejs/vue-cli <vue-cli@noreply.github.com>Kopia: Marek <
goc_marek@ma-al.com>, Author <author@noreply.github.com>Temat: Re:
[vuejs/vue-cli] make vue.config.js async, this will let pull param
(#4954)Data: Thu, 19 Mar 2020 14:03:37 -0700

One solution might be, to write a different script which save the
result on some temporary file; execute this script with something like
child_process.execSync, and use readFileSync to read the result in
vue.config.js

LinusBorg:
Well, this proposal doesn't work. I'll close this PR since it's not mergable in this form.

If we wanted to support an async config, we would have to probnably do an RFC about it and evaluate the impact this would have in various places.

gocmarek:
OK, clear for me.But think about it... 
Anyway, thanks a lot for all effort to make this nice piece ... Vue and
all around.


-------- Wiadomość oryginalna ----------Od: Thorsten Lünborg <
notifications@github.com>Odpowiedź do: vuejs/vue-cli <
reply@reply.github.com>Do:
vuejs/vue-cli <vue-cli@noreply.github.com>Kopia: Marek <
goc_marek@ma-al.com>, Mention <mention@noreply.github.com>Temat: Re:
[vuejs/vue-cli] make vue.config.js async, this will let pull param
(#4954)Data: Fri, 20 Mar 2020 03:00:09 -0700

Well, this proposal doesn't work. I'll close this PR since it's not
mergable in this form.
If we wanted to support an async config, we would have to probnably do
an RFC about it and evaluate the impact this would have in various
places.

PR #4985, opened 2019-12-20
fix(CORS): only allow connections from the designated host

Akryum:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
For security reasons, only allow connection from localhost.


packages/@vue/cli/lib/ui.js
```diff
@@ -36,7 +36,7 @@ async function ui (options = {}, context = process.cwd()) {
     subscriptionsPath: '/graphql',
     enableMocks: false,
     enableEngine: false,
-    cors: '*',
+    cors: host,
```
sodatea: Shouldn't this be a `corsOptions` object?
https://github.com/apollographql/apollo-server/blob/e7949d62c0f2a3f096573f1ee07f146d7aabd8af/packages/apollo-server-express/src/ApolloServer.ts#L27

sodatea:
I'm not sure if I was testing it incorrectly, but as far as I tried out, this change seems does not prevent other domains from accessing the GraphQL interface.

sodatea:
I tested this by running another instance of the `cli-ui` frontend and point the `VUE_APP_CLI_UI_URL` variable to the modified server instance.

sodatea:
Got it.

The `origin` check should be placed **in the WebSocket server**, that is, the `apolloServerOptions.subscriptions.onConnect` method.

It's because
> a CORS policy cannot be inserted into an HTTP Upgrade response

Akryum:
Good catch!

sodatea:
I got it fixed locally, by adding the following lines in [`vue-cli-plugin-apollo/graphql-server/index.js`](https://github.com/Akryum/vue-cli-plugin-apollo/blob/a52696165732381787dafe6b8e694d4b30af4826/graphql-server/index.js#L158):

```js
httpServer.on('upgrade', (req, socket) => {
  const { origin } = req.headers
  if (!origin || !/https?:\/\/localhost/.test(origin)) {
    socket.destroy()
  }
})
```

But since the `httpServer` instance is not exported from the plugin, I can't do the fix on the Vue CLI side.
Could you please add it to the plugin API?

sodatea:
I've migrated to a custom version of the `graphql-server` to work around this issue for now.

Should check it again after the apollo plugin adds the needed API.

PR #5063, opened 2020-01-09
Add optional chaining and nullish coalescing support

afzalsayed96:
Optional Chaining and Nullish Coalescing is currently stage 4. I think it would be nice if Vue CLI supports the modern syntax out of the box until it lands in @babel/preset-env babel/babel#10811.

What are your thoughts?

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [x] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/babel-preset-app/index.js
+++ b/packages/@vue/babel-preset-app/index.js
@@ -170,6 +170,8 @@ module.exports = (context, options = {}) => {
   // too many unstable proposals. Let's be conservative in the defaults here.
   plugins.push(
     require('@babel/plugin-syntax-dynamic-import'),
+    require('@babel/plugin-proposal-nullish-coalescing-operator'),
+    require('@babel/plugin-proposal-optional-chaining'),
     [require('@babel/plugin-proposal-decorators'), {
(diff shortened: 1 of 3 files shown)


sodatea:
I personally am against such additions, because:

1. Babel 7.8 is coming soon anyway. If we merged this PR, we would have to remove these two plugins soon after the upstream update. This is an unnecessary maintenance burden IMO.
2. Packages should have separation of concerns. It's the responsibility of `@babel/preset-env` to support those already standardized features and Vue CLI should not take care of this.
3. The upstream implementation in the mentioned pull request is more complete than this one, with proper browser target support.

Such features could look exciting in the short term and is indeed "nice to have". But I do think in the long run it's not a good practice to undertake the responsibilities of upstream packages unless there're convincing reasons (e.g. the upstream package is poorly maintained, which is not the case for babel).

afzalsayed96:
@sodatea Thank you for weighing in. I agree with the points made above.
Going ahead and closing this request.

PR #5089, opened 2020-01-17
fix: pwa-plugin avoid generating manifest when path is an URL

tkint:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

This PR is a bug fix which occure when you specify an URL inside the config's manifestPath option.

The bug is not blocking the `vue-cli-service build` command, but it throws a misleading error because it attemps to create a directory for the manifest, which is impossible when it is a full URL.

### DIFF
--- a/packages/@vue/cli-plugin-pwa/lib/HtmlPwaPlugin.js
+++ b/packages/@vue/cli-plugin-pwa/lib/HtmlPwaPlugin.js
@@ -151,27 +151,29 @@ module.exports = class HtmlPwaPlugin {
       })
     })
 
-    compiler.hooks.emit.tapAsync(ID, (data, cb) => {
-      const {
-        name,
-        themeColor,
-        manifestPath,
-        manifestOptions
-      } = this.options
-      const publicOptions = {
-        name,
-        short_name: name,
-        theme_color: themeColor
-      }
-      const outputManifest = JSON.stringify(
-        Object.assign(publicOptions, defaultManifest, manifestOptions)
-      )
-      data.assets[manifestPath] = {
-        source: () => outputManifest,
-        size: () => outputManifest.length
-      }
-      cb(null, data)
-    })
+    if (!isHrefAbsoluteUrl(this.options.manifestPath)) {
+      compiler.hooks.emit.tapAsync(ID, (data, cb) => {
+        const {
+          name,
+          themeColor,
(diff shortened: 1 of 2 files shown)


packages/@vue/cli-plugin-pwa/lib/HtmlPwaPlugin.js
```diff
@@ -185,8 +187,12 @@ function makeTag (tagName, attributes, closeTag = false) {
...
   let tagHref = `${href}${assetsVersionStr}`
-  if (!(/(http(s?)):\/\//gi.test(href))) {
+  if (!isHrefAbsoluteUrl(href)) {
     tagHref = `${publicPath}${tagHref}`
   }
   return tagHref
 }
+
+function isHrefAbsoluteUrl (href) {
+  return /(http(s?)):\/\//gi.test(href)
```
pksunkara: Instead of a function like this, why not use a package that does url validation?

sodatea: I don't think it's a good practice to introduce extra dependencies for one-liners

pksunkara: I am more worried about this part not being correct and exhaustive. What about if the path is given as `www.example.com`? Isn't it better to have a package which correctly understands the url and tells us about it?

sodatea: `www.example.com` is a relative path in the context of HTML, and it will fail this test as expected.

Other general-purpose packages may not serve the same interest as the pwa plugin. For example, the widely-used [`is-absolute-url`](https://github.com/sindresorhus/is-absolute-url) package (used by webpack-dev-server, postcss-normalize-url, etc.) But it treats `mainto:` and `file:` links as absolute URLs, too. That's correct according to the URL RFC, but not the case for PWA usage.

So it's better to contain such utility function in our own package, and write our own tests if really needed.

pksunkara: You are right. I stand corrected.

Issue #5022, opened 2020-01-02
Unable to npm run build on new project


ddreier:
### Version
4.1.2

### Environment info
```
Environment Info:

  System:
    OS: Windows 10 10.0.17134
    CPU: (8) x64 Intel(R) Xeon(R) CPU E3-1505M v5 @ 2.80GHz
  Binaries:
    Node: 10.16.2 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 6.13.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 42.17134.1098.0
  npmPackages:
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0
    @vue/babel-plugin-transform-vue-jsx:  1.1.2
    @vue/babel-preset-app:  4.1.2
    @vue/babel-preset-jsx:  1.1.2
    @vue/babel-sugar-functional-vue:  1.1.2
    @vue/babel-sugar-inject-h:  1.1.2
    @vue/babel-sugar-v-model:  1.1.2
    @vue/babel-sugar-v-on:  1.1.2
    @vue/cli-overlay:  4.1.2
    @vue/cli-plugin-babel: ^4.1.0 => 4.1.2
    @vue/cli-plugin-eslint: ^4.1.0 => 4.1.2
    @vue/cli-plugin-router:  4.1.2
    @vue/cli-plugin-vuex:  4.1.2
    @vue/cli-service: ^4.1.0 => 4.1.2
    @vue/cli-shared-utils:  4.1.2
    @vue/component-compiler-utils:  3.1.0
    @vue/preload-webpack-plugin:  1.1.1
    @vue/web-component-wrapper:  1.2.0
    eslint-plugin-vue: ^5.0.0 => 5.2.3
    vue: ^2.6.10 => 2.6.11
    vue-eslint-parser:  5.0.0
    vue-hot-reload-api:  2.3.4
    vue-loader:  15.8.3
    vue-style-loader:  4.1.2
    vue-template-compiler: ^2.6.10 => 2.6.11
    vue-template-es2015-compiler:  1.9.1
  npmGlobalPackages:
    @vue/cli: Not Found
```

### Steps to reproduce
```
PS C:\code> vue create -d repro
PS C:\code> cd repro
PS C:\code\repro> npm run build
```

### What is expected?
`npm run build` works out of the box

### What is actually happening?
```
PS C:\code\repro> npm run build

[...]

Issue #4424, opened 2019-08-12
webpack-dev-server 3.8  broken on windows 10


docteurzoidberg:
### Version
3.10.0

### Reproduction link
[npm run serve](npm run serve)

### Environment info
```
Environment Info:

  System:
    OS: Windows 10
    CPU: (4) x64 Intel(R) Core(TM) i5-7300U CPU @ 2.60GHz
  Binaries:
    Node: 10.16.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.17.3 - C:\Users\DrZoid\AppData\Roaming\npm\yarn.CMD
    npm: 6.10.3 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.18362.1.0
  npmPackages:
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0
    @vue/babel-plugin-transform-vue-jsx:  1.0.0
    @vue/babel-preset-app:  3.10.0
    @vue/babel-preset-jsx:  1.1.0
    @vue/babel-sugar-functional-vue:  1.0.0
    @vue/babel-sugar-inject-h:  1.0.0
    @vue/babel-sugar-v-model:  1.0.0
    @vue/babel-sugar-v-on:  1.1.0
    @vue/cli-overlay:  3.10.0
    @vue/cli-plugin-babel: ^3.10.0 => 3.10.0
    @vue/cli-plugin-eslint: ^3.10.0 => 3.10.0
    @vue/cli-service: ^3.10.0 => 3.10.0
    @vue/cli-shared-utils:  3.10.0
    @vue/component-compiler-utils:  2.6.0 (3.0.0)
    @vue/preload-webpack-plugin:  1.1.1
    @vue/web-component-wrapper:  1.2.0
    eslint-plugin-vue: ^5.0.0 => 5.2.3 (4.7.1)
    vue: ^2.6.10 => 2.6.10
    vue-eslint-parser:  5.0.0 (2.0.3)
    vue-hot-reload-api:  2.3.3
    vue-loader:  15.7.1
    vue-style-loader:  4.1.2
    vue-template-compiler: ^2.6.10 => 2.6.10
    vue-template-es2015-compiler:  1.9.1
  npmGlobalPackages:
    @vue/cli: Not Found
```

### Steps to reproduce
"npm run serve" on newly create project with "vue create" (default options)

### What is expected?
'npm run serve' should work

### What is actually happening?
'npm run serve' give this error: 

[...]

Issue #4174, opened 2019-06-20
After start vue-cli-3 tutorial  project 98% after emitting CopyPlugin


skortabeast:
### Version
3.8.4

### Environment info
```
 System:
    OS: Windows 10
    CPU: (12) x64 Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
  Binaries:
    Node: 10.16.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.16.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.9.0 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 42.17134.1.0
  npmPackages:
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0
    @vue/babel-plugin-transform-vue-jsx:  1.0.0
    @vue/babel-preset-app:  3.8.0
    @vue/babel-preset-jsx:  1.0.0
    @vue/babel-sugar-functional-vue:  1.0.0
    @vue/babel-sugar-inject-h:  1.0.0
    @vue/babel-sugar-v-model:  1.0.0
    @vue/babel-sugar-v-on:  1.0.0
    @vue/cli-overlay:  3.8.0
    @vue/cli-plugin-babel: ^3.8.0 => 3.8.0
    @vue/cli-plugin-eslint: ^3.8.0 => 3.8.0
    @vue/cli-service: ^3.8.0 => 3.8.4
    @vue/cli-shared-utils:  3.8.0
    @vue/component-compiler-utils:  2.6.0
    @vue/preload-webpack-plugin:  1.1.0
    @vue/web-component-wrapper:  1.2.0
    eslint-plugin-vue: ^5.0.0 => 5.2.2 (4.7.1)
    vue: ^2.6.10 => 2.6.10
    vue-eslint-parser:  5.0.0 (2.0.3)
    vue-hot-reload-api:  2.3.3
    vue-loader:  15.7.0
    vue-style-loader:  4.1.2
    vue-template-compiler: ^2.6.10 => 2.6.10
    vue-template-es2015-compiler:  1.9.1
  npmGlobalPackages:
    @vue/cli: Not Found
```

### Steps to reproduce
Download Nodejs version 10.16.0 , install, npm install -g @vue/cli ,  vue create hello-world ,  npm run serve

### What is expected?
project is compiled

### What is actually happening?
INFO  Starting development server...
 98% after emitting CopyPlugin

[...]

Issue #3665, opened 2019-03-18
[SOLVED] Local webfonts not working


MrDouh:
### Version
3.5.1

### Environment info
```
  System:
    OS: Windows 10
    CPU: (8) x64 Intel(R) Core(TM) i7-6700K CPU @ 4.00GHz
  Binaries:
    Node: 10.15.3 - C:\Program Files\nodejs\node.EXE
    Yarn: Not Found
    npm: 6.4.1 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.17763.1.0
  npmPackages:
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0-beta.2
    @vue/babel-plugin-transform-vue-jsx:  1.0.0-beta.2
    @vue/babel-preset-app:  3.5.1
    @vue/babel-preset-jsx:  1.0.0-beta.2
    @vue/babel-sugar-functional-vue:  1.0.0-beta.2
    @vue/babel-sugar-inject-h:  1.0.0-beta.2
    @vue/babel-sugar-v-model:  1.0.0-beta.2
    @vue/babel-sugar-v-on:  1.0.0-beta.2
    @vue/cli-overlay:  3.5.1
    @vue/cli-plugin-babel: ^3.5.0 => 3.5.1
    @vue/cli-plugin-eslint: ^3.5.0 => 3.5.1
    @vue/cli-service: ^3.5.0 => 3.5.1
    @vue/cli-shared-utils:  3.5.1
    @vue/component-compiler-utils:  2.6.0
    @vue/preload-webpack-plugin:  1.1.0
    @vue/web-component-wrapper:  1.2.0
    eslint-plugin-vue: ^5.0.0 => 5.2.2
    vue: ^2.6.6 => 2.6.9
    vue-eslint-parser:  2.0.3
    vue-hot-reload-api:  2.3.3
    vue-loader:  15.7.0
    vue-style-loader:  4.1.2
    vue-template-compiler: ^2.5.21 => 2.6.9
    vue-template-es2015-compiler:  1.9.1
  npmGlobalPackages:
    @vue/cli: Not Found
```

### Steps to reproduce
Hello everyone,

I made a simple Vue Cli project using default configuration.
In src/assets I created both a "css" and a "fonts" folders.

In the "fonts" folder I pasted those font files :
- Roboto-regular.woff2
- Roboto-regular.woff
- Roboto-regular.ttf
- Roboto-black.woff2
[...]

PR #5096, opened 2020-01-20
feat: convert project name to lowercase and warn on creation

### LINKED ISSUE #2547: vue-cli shows "invalid project name" for project names or directories that have uppercase letters
### Version
3.0.3

### Node and OS info
Node 8.12.0 / yarn 1.9.4 / Ubuntu

### Steps to reproduce
The issue can be reproduced by trying to initialize a project that has uppercase letters in its name. `vue create Uppecase` will give this error. 

This is error isn't so obvious when you try to initialize a project with `vue create .` in the current directory which has uppercase letters in its name. In this case, it just says `Invalid project name: "."`.

### What is expected?
vue-cli should allow me to create projects that have uppercase letters in their name.

vue-cli is not letting me create projects that have uppercase letters in their name.

### LINKED ISSUE #5032: vue create . does not allow upper case letters.  Causes headaches when integrating into existing project
When integrating vue into existing web frameworks like Aspnet you cannot install into the root web project unless your aspnet web projects folder is all lower case.

Most guidelines on the web say to make a sub folder called app that you stuff everything into.  That seems like a hack.  I assume the casing restriction is due to npm package.json restrictions.

Allow users to create a package.json file ahead of time and take the name from that instead.  

So steps to add vue to existing project could be

npm init --yes

vue create .

If the naming restriction goes beyond package.json, just add a --lower-name flag or something.

sodatea:
closes #2547
closes #5032

I'm still very hesitant about adding this feature, though.

Aside from the reasons mentioned in https://github.com/facebook/create-react-app/issues/2165, here are my thoughts:

First, this change allows project creation in a folder with uppercase
letters in its name. It is strongly discouraged and may cause many
weird issues all over the ecosystem.

For example, #5022, #4424, #3665, [#4174](https://github.com/vuejs/vue-cli/issues/4174#issuecomment-569709494) are all
caused by case issues. Adding support for uppercase project names will
only worsen this situation.

Secondly, it adds a lot of maintenance burden on us. As noted in the
comments, these prompts are hard to test right now (because
`createTestProject` runs in another process so it's hard to intercept
the prompts). Even if such test utilities are added in the future, it's
still very tedious to take care of all the case issues in the test
suite.

What's worse is that we can affect the project folders created by
@vue/cli by converting the project name to lower case. But for
`vue create .`, we cannot change the current folder's name. So, we'll
have another edge case to test (projectName != folderName).

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli/lib/create.js
+++ b/packages/@vue/cli/lib/create.js
@@ -7,6 +7,7 @@ const { getPromptModules } = require('./util/createTools')
 const { chalk, error, stopSpinner, exit } = require('@vue/cli-shared-utils')
 const validateProjectName = require('validate-npm-package-name')
 
+// TODO: add test cases for prompts in this file
 async function create (projectName, options) {
   if (options.proxy) {
     process.env.HTTP_PROXY = options.proxy
@@ -14,9 +15,11 @@ async function create (projectName, options) {
 
   const cwd = options.cwd || process.cwd()
   const inCurrent = projectName === '.'
-  const name = inCurrent ? path.relative('../', cwd) : projectName
-  const targetDir = path.resolve(cwd, projectName || '.')
 
+  const originalName = inCurrent ? path.relative('../', cwd) : projectName
+  const name = originalName.toLowerCase()
+
+  // TODO: should also implement this logic in the UI
   const result = validateProjectName(name)
   if (!result.validForNewPackages) {
     console.error(chalk.red(`Invalid project name: "${name}"`))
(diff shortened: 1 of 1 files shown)


sodatea:
Another issue caused by uppercase letters is https://github.com/vuejs/vue-cli/issues/4251
It's because though Vue CLI UI already converts project names to uppercase, it does not convert `targetDir` along with that. https://github.com/vuejs/vue-cli/blob/8fcea225b7b4406ff7c03e34d4e962b45d78ef9f/packages/%40vue/cli-ui/apollo-server/connectors/projects.js#L260

It's easily fixable but it's an example of how many edge cases there could have been if we supported uppercase project names.

pksunkara:
I wouldn't actually allow the uppercase situation at all.

PR #5148, opened 2020-02-05
fix(e2e-nightwatch): fix eslint config generation

cexbrayat:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Commit 1b64ff8464c317f8d5616626bf483f582c2e01c3 introduced a nested if in `_eslintrc.js` with the same condition than the wrapping one. I may be missing something really obvious, but I think it was meant to check for TS instead.


sodatea:
😅Ahh you are right. Thanks!

PR #5150, opened 2020-02-06
fix: should infer package manager from config if there's no lockfile in the project

sodatea:
Fixed monorepo use cases.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-shared-utils/lib/env.js
+++ b/packages/@vue/cli-shared-utils/lib/env.js
@@ -140,7 +140,7 @@ exports.hasProjectNpm = (cwd) => {
   }
 
   const lockFile = path.join(cwd, 'package-lock.json')
-  const result = fs.exists(lockFile)
+  const result = fs.existsSync(lockFile)
   _npmProjects.set(cwd, result)
   return result
 }
--- a/packages/@vue/cli-shared-utils/lib/env.js
+++ b/packages/@vue/cli-shared-utils/lib/env.js
@@ -130,6 +130,21 @@ function checkPnpm (result) {
   return result
 }
 
+const _npmProjects = new LRU({
+  max: 10,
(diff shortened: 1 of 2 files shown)


packages/@vue/cli/lib/util/ProjectPackageManager.js
```diff
@@ -88,8 +89,17 @@ class PackageManager {
     if (forcePackageManager) {
```
Akryum: How do we force the package manager?

packages/@vue/cli-shared-utils/lib/env.js
```diff
@@ -130,6 +130,21 @@ function checkPnpm (result) {
...
+const _npmProjects = new LRU({
+  max: 10,
+  maxAge: 1000
+})
+exports.hasProjectNpm = (cwd) => {
+  if (_npmProjects.has(cwd)) {
+    return true
+  }
+
+  const lockFile = path.join(cwd, 'package-lock.json')
```
Akryum: Maybe we should use findup?

packages/@vue/cli/lib/util/ProjectPackageManager.js
sodatea: The `-m / --packageManager` option of `vue create`. Maybe we should add this option for `vue add` and `vue invoke` too?

packages/@vue/cli-shared-utils/lib/env.js
sodatea: Yeah that makes more sense.

PR #5166, opened 2020-02-08
chore: switch over to leven for command suggestion

### LINKED ISSUE #5164: Switch over to leven for command suggestion
As of now [didyoumean.js](https://github.com/dcporter/didyoumean.js/) matching engine is used to suggest matching commands if the user mistypes. But it is not longer actively maintained. Instead I would suggest switching over to [leven](https://github.com/sindresorhus/leven) which is the fastest JS implementation of the Levenshtein distance algorithm. 

`leven` returns the distance between 2 strings which can be made use of to find the exact match.

jamesgeorge007:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

closes #5164 

- [ ] Yes
- [x] No

**Other information:**
N/A

### DIFF
--- a/packages/@vue/cli/bin/vue.js
+++ b/packages/@vue/cli/bin/vue.js
@@ -5,10 +5,7 @@
 
 const { chalk, semver } = require('@vue/cli-shared-utils')
 const requiredVersion = require('../package.json').engines.node
-const didYouMean = require('didyoumean')
-
-// Setting edit distance to 60% of the input string's length
-didYouMean.threshold = 0.6
+const leven = require('leven')
 
 function checkNodeVersion (wanted, id) {
   if (!semver.satisfies(process.version, wanted)) {
@@ -261,7 +258,9 @@ if (!process.argv.slice(2).length) {
 function suggestCommands (unknownCommand) {
   const availableCommands = program.commands.map(cmd => cmd._name)
 
-  const suggestion = didYouMean(unknownCommand, availableCommands)
+  const suggestion = availableCommands.find(cmd => {
+    return leven(cmd, unknownCommand) < 3
+  })
   if (suggestion) {
     console.log(`  ` + chalk.red(`Did you mean ${chalk.yellow(suggestion)}?`))
   }
--- a/packages/@vue/cli/bin/vue.js
+++ b/packages/@vue/cli/bin/vue.js
@@ -261,7 +261,8 @@ function suggestCommands (unknownCommand) {
   let suggestion
 
   availableCommands.forEach(cmd => {
-    if (leven(cmd, unknownCommand) < 3) {
+    const isBestMatch = leven(cmd, unknownCommand) < leven(suggestion || '', unknownCommand)
+    if (leven(cmd, unknownCommand) < 3 && isBestMatch) {
       suggestion = cmd
     }
   })
--- a/packages/@vue/cli/bin/vue.js
+++ b/packages/@vue/cli/bin/vue.js
@@ -257,10 +257,15 @@ if (!process.argv.slice(2).length) {
 
 function suggestCommands (unknownCommand) {
   const availableCommands = program.commands.map(cmd => cmd._name)
-
-  const suggestion = availableCommands.find(cmd => {
-    return leven(cmd, unknownCommand) < 3
+  
+  let suggestion
+  
+  availableCommands.forEach(cmd => {
+    if (leven(cmd, unknownCommand) < 3) {
+      suggestion = cmd
+    }
   })
+
   if (suggestion) {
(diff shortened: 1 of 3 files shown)


packages/@vue/cli/bin/vue.js
```diff
@@ -261,7 +258,9 @@ if (!process.argv.slice(2).length) {
 function suggestCommands (unknownCommand) {
   const availableCommands = program.commands.map(cmd => cmd._name)
 
-  const suggestion = didYouMean(unknownCommand, availableCommands)
+  const suggestion = availableCommands.find(cmd => {
+    return leven(cmd, unknownCommand) < 3
```
sodatea: Better sort and find the closest result.
Otherwise, it might return `init` for the `infi` (should return `info`).

jamesgeorge007: Could you elaborate?

sodatea: Using `find` it will suggest the first command that is less than 3 characters away from the input. It may not be the best guess because there may be a command that is even closer to the user input.

jamesgeorge007: Gotcha.

jamesgeorge007: Do you have any suggestion regarding the implementation? we've to iterate through the entire array of available commands and find the most appropriate match. One approach that I could think of is to filter out the matches and going ahead picking up the right one from them.

sodatea: Just iterate the array and find the one with the shortest distance would suffice.

packages/@vue/cli/bin/vue.js
```diff
@@ -261,7 +258,14 @@ if (!process.argv.slice(2).length) {
 function suggestCommands (unknownCommand) {
   const availableCommands = program.commands.map(cmd => cmd._name)
 
-  const suggestion = didYouMean(unknownCommand, availableCommands)
+  let suggestion
+
+  availableCommands.forEach(cmd => {
+    if (leven(cmd, unknownCommand) < 3) {
+      suggestion = cmd
```
sodatea: You still didn't compare two possible suggestions.

jamesgeorge007:
I'm sorry, didn't get you.

sodatea:
Just try the two commands:
`vue infi` and `vue iniy`
Should return `vue info` and `vue init` respectively.

jamesgeorge007:
> Try the two commands:
> `vue infi` and `vue iniy`
> Should return `vue info` and `vue init` respectively.

```js
    if (leven(cmd, unknownCommand) < 2) {
```

This would do that.

packages/@vue/cli/bin/vue.js
jamesgeorge007: It would be great if you could give me a better picture regarding the implementation strategy that you have in mind.

sodatea:
But it's not the best fix

jamesgeorge007:
Please suggest possible alternatives?

sodatea:
The ideal algorithm is to find the **closest** command, regardless of the distance. (You can set a threshold to stop suggesting when even the closest command is too unrelated, though)

packages/@vue/cli/bin/vue.js
jamesgeorge007:
```js
  availableCommands.forEach(cmd => {
    if (leven(cmd, unknownCommand) < 2) {
      //
   }
})
```
This would find the closest match. To get going ahead can you suggest a possible implementation?

sodatea: "Closest" doesn't mean less than a certain threshold. It means the distance is the smallest. So I have suggested to "compare" and "sort" the distances.

packages/@vue/cli/bin/vue.js
```diff
@@ -261,7 +258,14 @@ if (!process.argv.slice(2).length) {
 function suggestCommands (unknownCommand) {
   const availableCommands = program.commands.map(cmd => cmd._name)
 
-  const suggestion = didYouMean(unknownCommand, availableCommands)
+  let suggestion
+
+  availableCommands.forEach(cmd => {
+    if (leven(cmd, unknownCommand) < 3) {
```
sodatea:
```suggestion
    if (leven(cmd, unknownCommand) < 3 && leven(cmd, unknownCommand) < leven(suggestion || '', unknwonCommand)) {
```

(Of course, you can also memoize the result of `leven(suggestion || '', unknwonCommand)` to a variable to make the code cleaner. But that's the basic idea.

PR #5170, opened 2020-02-09
feat: use @vue/compiler-sfc as a compiler for TS if available

cexbrayat:
Update: I'll let you decide but #5172 is probably a better alternative. See https://github.com/vuejs/vue-cli-plugin-vue-next/issues/5 for more context

The `fork-ts-checker-webpack-plugin` is using `vue-template-compiler` by default, and this compiler is not the correct one to pick for vue-next. This commit tries to load `@vue/compiler-sfc` and falls back to `vue-template-compiler` if it does not find it.

It also bumps the `fork-ts-checker-webpack-plugin` to use v3.1.3

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-plugin-typescript/index.js
+++ b/packages/@vue/cli-plugin-typescript/index.js
@@ -77,12 +77,24 @@ module.exports = (api, projectOptions) => {
     })
 
     if (!process.env.VUE_CLI_TEST) {
+      // try to load `@vue/compiler-sfc` if the project is using Vue 3.
+      // if it is not available, it uses `vue-template-compiler`
+      let compiler = '@vue/compiler-sfc'
+      try {
+        require.resolve(compiler)
+        // use a shim as @vue/compiler-sfc does not offer the `parseComponent` function
+        // but a `parse` function
+        // the shim only delegates to the parse function
+        compiler = '@vue/cli-plugin-typescript/vue-compiler-sfc-shim'
+      } catch (e) {
+        compiler = 'vue-template-compiler'
+      }
       // this plugin does not play well with jest + cypress setup (tsPluginE2e.spec.js) somehow
       // so temporarily disabled for vue-cli tests
       config
         .plugin('fork-ts-checker')
           .use(require('fork-ts-checker-webpack-plugin'), [{
-            vue: true,
+            vue: { enabled: true, compiler },
             tslint: projectOptions.lintOnSave !== false && fs.existsSync(api.resolve('tslint.json')),
             formatter: 'codeframe',
             // https://github.com/TypeStrong/ts-loader#happypackmode-boolean-defaultfalse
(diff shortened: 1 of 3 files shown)


sodatea:
I think we can just add a file that shims the `compiler` API add pass it to the `fork-ts-checker-webpack-plugin`.

sodatea:
All it needs is a path to a file that exports the `parseComponent` API, which can be implemented by wrapping around the `@vue/compiler-sfc` public API.
https://github.com/TypeStrong/fork-ts-checker-webpack-plugin/blob/5008da9f4bd732bf70cfae0e187e4e4ecc95dfd1/src/VueProgram.ts#L303-L323

cexbrayat:
@sodatea I updated the PR based on your idea.
To sum up:
- we need to use the latest `fork-ts-checker-webpack-plugin` to be able to give it another compiler (the one currently used does not)
- the second commit introduces a shim that just delegates to `parse` when `parseComponent` is called by the webpack plugin.
I am open to suggestions for the file name and location.

We should now be able to remove `vue-template-compiler` from the devDependencies when using `vue-next` 🙂

packages/@vue/cli-plugin-typescript/package.json
```diff
@@ -26,7 +26,7 @@
     "@types/webpack-env": "^1.15.1",
     "@vue/cli-shared-utils": "^4.2.2",
     "cache-loader": "^4.1.0",
-    "fork-ts-checker-webpack-plugin": "^1.5.1",
+    "fork-ts-checker-webpack-plugin": "^4.0.3",
```
sodatea: Could you please revert this to v3.x?
I just checked the changelog and the `compiler` option was supported since v3.1.
Meanwhile, the v4 release came with too many breaking changes, such as removing tslint support, we can't ship it in a minor version.

packages/@vue/cli-plugin-typescript/index.js
```diff
@@ -77,12 +77,24 @@ module.exports = (api, projectOptions) => {
...
     if (!process.env.VUE_CLI_TEST) {
+      // try to load `@vue/compiler-sfc` if the project is using Vue 3.
+      // if it is not available, it uses `vue-template-compiler`
+      let compiler = '@vue/compiler-sfc'
+      try {
+        require(compiler)
+        // use a shim as @vue/compiler-sfc does not offer the `parseComponent` function
+        // but a `parse` function
+        // the shim only delegates to the parse function
+        compiler = '@vue/cli-plugin-typescript/vue-compiler-sfc-shim'
```
sodatea: It might be safer to use `require.resolve`?

cexbrayat:
@sodatea I updated the PR with your feedback:
- now bumps to `fork-ts-checker-webpack-plugin@3.1.1`
- now uses `require.resolve` to load the new compiler

PR #5186, opened 2020-02-13
fix: prevent mixing up of lock files

jamesgeorge007:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [x] Underlying tools
- [ ] Other, please describe:

It is likely that people might use `npm` to install dependencies and commit `package-lock.json` while opening a PR. Ignore `package-lock` to avoid mixing up of lock files.

- [ ] Yes
- [x] No

**Other information:**
N/A

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ temp
 .version
 .versions
 .changelog
+package-lock.json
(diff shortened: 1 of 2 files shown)


sodatea:
I don't think we need this.
It's already very clear that the project is developed with `lerna` and `yarn`, and you can only bootstrap the project with `yarn`.

PR #5265, opened 2020-03-12
feat(cli-service): add "VUE_CLI_SERVICE_PARAMS"

zyao89:

- [ ] Bugfix
- [x] Feature
- [ ] Refactor
- [ ] Docs
- [ ] Code style update
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-service/bin/vue-cli-service.js
+++ b/packages/@vue/cli-service/bin/vue-cli-service.js
@@ -12,7 +12,9 @@ if (!semver.satisfies(process.version, requiredVersion)) {
 }
 
 const Service = require('../lib/Service')
-const service = new Service(process.env.VUE_CLI_CONTEXT || process.cwd())
+const getGlobalParams = require('../lib/util/getGlobalParams')
+const context = process.env.VUE_CLI_CONTEXT || process.cwd()
+const service = new Service(context, getGlobalParams(context))
 
 const rawArgv = process.argv.slice(2)
 const args = require('minimist')(rawArgv, {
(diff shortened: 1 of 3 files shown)


sodatea:
What's the use case?

zyao89:
i want to preload some plugins. doesn't only depend on pkg.

eg: 
```js
// ./params.js
module.exports = {
    useBuiltIn: true,
    plugins: [
        ...
    ],
};
```
or
```js
// ./params.js
module.exports = function(){
    return {
        useBuiltIn: true,
        plugins: [
            ...
        ],
    };
}
```

```js
process.env.VUE_CLI_SERVICE_PARAMS = './params.js'
```

> What's the use case?

sodatea:
> doesn't only depend on pkg

Why? Why don't you want it to depend on pkg? Any technical obstacles or personal taste?

zyao89:
just want more choices. 

eg:  custom cli.
```js
module.exports = () => {
    const PRI_DEPS = [ ... ]; // preload
    return {
        useBuiltIn: true,
        pkg,
        plugins: PRI_DEPS.map(id => {
            return {
                id,
                apply: require(id),
            };
        }).concat(projectPlugins),
    };
}
```

> > doesn't only depend on pkg
> 
> Why? Why don't you want it to depend on pkg? Any technical obstacles or personal taste?

sodatea:
> just want more choices.

It's not how we add features. We need a convincing use case. We don't want to increase the maintenance burden unless for good reasons.

> eg: custom cli.

Then why modify the Vue CLI source? You can write your own CLI on top of the Vue CLI API, by requiring `@vue/cli-service/lib/Service` in your own package.

zyao89:
@sodatea thanks! I will close this, but I make a new PR(#5311) to solve it.

PR #5293, opened 2020-03-19
support vue.config.cjs

### LINKED ISSUE #5046: Support "type" field in package.json for ES Modules
Currently the `vue.config.js` must be named exactly that, as far as I can see.

I'm using the `"type": "moduel"` key in `package.json`, as I'm using the `--experimental-modules` feature of Nodejs. However, as vue.config.js is not an ES module, an error is thrown and nothing works.

The suggested solution is to rename the file `vue.config.cjs`, then node accepts the file is in CJS format and it's happy. This change has been implemented in other tools, and, although technically still experimental, it's not likely to be for long. It's already been out from under a flag once, but had to go back because of a lack of ES Lint support (that's been resolved).

It would be helpful, and presumably wouldn't have an impact elsewhere to accept a file called `vue.config.cjs`, which is functionally and formatted identically (simply accept a second file name with a different extension).

Additionally, presumably this will have to be done eventually, unless Vue intends never to support ES modules, which wouldn't be helpful.

I'm happy to write up a PR, if someone can point me in the direction of where to look...

The config file can be called either `vue.config.js` or `vue.config.cjs`. They would be identical, other than the extension.

simon300000:
if `"type": "module"` was set in package.json, the vue service will emit error.
```
 ERROR  Error loading vue.config.js:
 ERROR  Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: .../vue.config.js
require() of ES modules is not supported.
.../vue.config.js from .../node_modules/@vue/cli-service/lib/Service.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
Instead rename vue.config.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from .../package.json.
```

this pr enable use of `vue.config.cjs`, as a simple solution in those environment.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
close #5046 

It would be nice if `import()` is used instead of `require()`, which will support both commonjs and module, but this makes the whole process asynchronous and many code need to be changed.

### DIFF
--- a/packages/@vue/cli-service/__tests__/Service.spec.js
+++ b/packages/@vue/cli-service/__tests__/Service.spec.js
@@ -1,5 +1,6 @@
 jest.mock('fs')
 jest.mock('/vue.config.js', () => ({ lintOnSave: false }), { virtual: true })
+jest.mock('/vue.config.cjs', () => ({ lintOnSave: true }), { virtual: true })
 jest.mock('vue-cli-plugin-foo', () => () => {}, { virtual: true })
 
 const fs = require('fs')
@@ -61,7 +62,7 @@ test('env loading for custom mode', () => {
 test('loading plugins from package.json', () => {
   mockPkg({
     devDependencies: {
-      'bar': '^1.0.0',
+      bar: '^1.0.0',
       '@vue/cli-plugin-babel': '^4.2.0',
       'vue-cli-plugin-foo': '^1.0.0'
     }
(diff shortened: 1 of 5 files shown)


simon300000:
I have no idea why AppVeyor is failed 😢

simon300000:
I merged the dev, the tests also passed, thank you!

simon300000:
I worked out different solution which uses `import(vue.config.js)`,

But I don’t know which one is more suitable.

https://github.com/vuejs/vue-cli/compare/dev...simon300000:import(vue.config.js)

Should I will open a different PR/issue to discuss about it?

sodatea:
Thank you very much for the help!

I'd very like to support the `import('vue.config.js')` form too. But I'm worried about the extra implementation complexity that the `deasync` package brings in. It is a native module that requires the `node-gyp` toolchain to be present in the user environment, thus may cause compatibility issues under some circumstances.

Out of the same concern, we've made the `fibers` dependency optional for the `sass` setup: https://github.com/vuejs/vue-cli/pull/3628

So, as of now, I think we'd go with the implementation in this PR to support `vue.config.js`.
For true ESM support, please open a separate issue to discuss.

IMO we should support async loading of the config before supporting ESM imports, so that we can avoid using `deasync`.

PR #5294, opened 2020-03-19
feat(service): Allow mocha unit tests debugger to be bound to a specified IP and port

darrylkuhn:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
This change allows developers to specify a host and port to bind the node debugger to. This is helpful in my case where I am running my application in a docker container and need to allow the debugger to listen on an IP other than the default localhost

### DIFF
--- a/packages/@vue/cli-plugin-unit-mocha/index.js
+++ b/packages/@vue/cli-plugin-unit-mocha/index.js
@@ -37,8 +37,9 @@ module.exports = api => {
       `https://sysgears.github.io/mochapack/docs/installation/cli-usage.html`
     )
   }, (args, rawArgv) => {
-    const inspectPos = rawArgv.indexOf('--inspect-brk')
     let nodeArgs = []
+
+    const inspectPos = rawArgv.findIndex(arg => arg.startsWith('--inspect-brk'))
     if (inspectPos !== -1) {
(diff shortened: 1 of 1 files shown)


packages/@vue/cli-plugin-unit-mocha/index.js
```diff
@@ -37,9 +37,17 @@ module.exports = api => {
       `https://sysgears.github.io/mochapack/docs/installation/cli-usage.html`
     )
   }, (args, rawArgv) => {
-    const inspectPos = rawArgv.indexOf('--inspect-brk')
```
sodatea, line 40: I think the functionality can be achieved with a one-liner:

```js
const inspectPos = rawArgv.findIndex(arg => arg.startsWith('--inspect-brk'))
```

darrylkuhn: Yes indeed - that is a cleaner implementation. I have pushed a change to reflect your suggestion.

sodatea: Thanks!

PR #5335, opened 2020-03-29
docs: Use `vue add @vue/eslint` instead of `vue add eslint`

zetaraku:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

* `plugins-and-presets.md`: According to **line 29**, it should be `vue add @vue/eslint` instead of `vue add eslint`.
* `cli-service.md `: It's better to use `npx vue-cli-service` instead of `./node_modules/.bin/vue-cli-service` now.

### DIFF
--- a/docs/guide/cli-service.md
+++ b/docs/guide/cli-service.md
@@ -2,7 +2,7 @@
 
 ## Using the Binary
 
-Inside a Vue CLI project, `@vue/cli-service` installs a binary named `vue-cli-service`. You can access the binary directly as `vue-cli-service` in npm scripts, or as `./node_modules/.bin/vue-cli-service` from the terminal.
+Inside a Vue CLI project, `@vue/cli-service` installs a binary named `vue-cli-service`. You can access the binary directly as `vue-cli-service` in npm scripts, or as `npx vue-cli-service` from the terminal.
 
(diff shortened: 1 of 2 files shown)


sodatea:
What's your Vue CLI version?
It should have been supported since v3.11

sodatea:
As for the `npx` part, not everyone has `npx` installed on their machine. Some users only use yarn.

zetaraku:
@sodatea The document on line 29 says "The command resolves @vue/eslint to the full package name @vue/cli-plugin-eslint".
For consistency, although it is supported, shouldn't this line be `@vue/eslint`?

sodatea:
Yeah for consistency the statement at line 29 needs to be changed.
We can add more explanations on this though:

For core plugins, it will resolve to `@vue/cli-plugin-eslint`; for any other 3rd-party plugin, it will resolve to `vue-cli-plugin-foo`.

zetaraku:
Hmm... maybe we need to add another example for adding a plugin with a non-@vue scope now.

PR #5354, opened 2020-04-04
feat: use vue-jest v4

lmiller1990:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No (don't think so)

**Other information:**

Update to the latest vue-jest version. It has some cool new features like support for vue-i18n. I updated some of my more complex Vue apps to use vue-jest v4 and it worked fine.


sodatea:
I'm afraid updating the underlying vue-jest version could mean breaking changes to some users.
We can do this in the next major.

But before that, vue-jest needs to ship a new v4 release so that we can support Jest 26.

lmiller1990:
Alright. I will get the latest v4 out and test with Jest 26.

PR #5391, opened 2020-04-15
Remove ListFilter logic

rcidaleassumpo:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**
I'm currently studying the project to see patterns used in real world projects, and I found this logic that's currently being used, but it seems like there is not much need for it?
I would like if possible your opinion on why you guys decided to use the ListFilter, instead of a cleaner solution? 
Thank you

### DIFF
--- a/packages/@vue/cli-ui/src/components/dashboard/WidgetAddPane.vue
+++ b/packages/@vue/cli-ui/src/components/dashboard/WidgetAddPane.vue
@@ -34,19 +34,12 @@
         />
 
         <template v-else-if="data">
-          <ListFilter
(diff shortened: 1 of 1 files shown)


rcidaleassumpo:
> Thanks!
> I think it is fine to remove `ListFilter`.
> Would you also remove other usages of `ListFilter` in this PR altogether, please?

Yes!

PR #5522, opened 2020-05-25
feat(babel-preset-app): pass full config to @babel/preset-env

### LINKED ISSUE #5208: @vue/babel-preset-app option exclude conflict
### Version
3.12.1

### Reproduction link
[https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/babel-preset-app/index.js#L8](https://github.com/vuejs/vue-cli/blob/dev/packages/%40vue/babel-preset-app/index.js#L8)

### Environment info
```
Chrome: 82
```

### Steps to reproduce
at babel.config.js config `exclude:['es6.promise']`

### What is expected?
build dist code exclude core-js es6.promise

only set exclude:['es6.promise'] then build dist code still include core-js es6.promise

---
require config polyfills:['es6.array.iterator','es6.object.assign'] rewrite defaultPolyfills 

but node_modules/@babel/preset-env/lib/built-in-definitions.js has Promise definitions

```js
const definitions = {
  builtins: {
    //...
    Map: ["es6.map"].concat(CommonIterators),
    Number: "es6.number.constructor",
    Promise: ["es6.object.to-string", "es6.promise"]
    //...
  }
}

```

may be defaultPolyfills can remove es6.promise

lucaswerkmeister:
Pass not just the useBuiltIns and corejs options, but the whole envOptions object, into the @babel/preset-env preset that is used to transform @babel/runtime, just like for the @babel/preset-env that is used for the application source code. This allows users to also specify other options, such as `exclude` and `polyfills`, and have them apply here too.

In particular, this can be used to exclude the Promise polyfill, e. g. if Promise is already polyfilled in some other way. Previously,

```js
exclude: ['es.promise'],
polyfills: ['es.array.iterator', 'es.object.assign'],
```

could be used to configure the preset for the application source code, but the babel runtime would be transformed without those options, and so es.promise would still end up being included.

Closes #5208

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No (I think)

**Other information:**

### DIFF
--- a/packages/@vue/babel-preset-app/__tests__/babel-preset.spec.js
+++ b/packages/@vue/babel-preset-app/__tests__/babel-preset.spec.js
@@ -209,3 +209,18 @@ test('should inject polyfills / helpers using "import" statements for an es modu
   expect(code).toMatch('import "core-js/modules/es.promise"')
   expect(code).not.toMatch('require(')
 })
+
+test('should not inject excluded polyfills', () => {
+  const { code } = babel.transformSync(`
+    new Promise()
+  `.trim(), {
+    babelrc: false,
+    presets: [[preset, {
+      exclude: ['es.promise'],
+      polyfills: ['es.array.iterator', 'es.object.assign']
+    }]],
+    filename: 'test-entry-file.js'
+  })
+
+  expect(code).not.toMatch('es.promise')
+})
--- a/packages/@vue/babel-preset-app/index.js
+++ b/packages/@vue/babel-preset-app/index.js
@@ -246,10 +246,7 @@ module.exports = (context, options = {}) => {
       // https://github.com/babel/babel/issues/9903
       include: [/@babel[\/|\\\\]runtime/],
       presets: [
-        [require('@babel/preset-env'), {
-          useBuiltIns,
-          corejs: useBuiltIns ? require('core-js/package.json').version : false
-        }]
+        [require('@babel/preset-env'), envOptions]
       ]
     }]
   }


lucaswerkmeister:
(On second thought, `polyfills: ['es.object.assign']` might be better than `polyfills: ['es.array.iterator', 'es.object.assign']`, since according to [the corresponding comment](https://github.com/vuejs/vue-cli/blob/34f303b0b6a03f96e8c21d308931219154a7e627/packages/%40vue/babel-preset-app/index.js#L5-L7), `es.array.iterator` is only needed to make `es.promise` work on IE. But I’m not sure.)

lucaswerkmeister:
If anyone is wondering where this PR is coming from or interested in more context, feel free to look at [this task](https://phabricator.wikimedia.org/T248840) and specifically [this comment](https://phabricator.wikimedia.org/T248840#6163535) in our downstream bug tracker :)

sodatea:
Thanks!
Actually I was thinking about dropping the transpilation of `@babel/runtime` entirely considering https://github.com/babel/babel/issues/7597 has been fixed.

But I'm not sure if that's safe to do…
Maybe this PR is a better option before the runtime is sure to be bug-free.

lucaswerkmeister:
@micgro42 I think you were also wondering whether the transpilation was still necessary, did you come to any conclusion?

micgro42:
"Conclusion"? No. My approach would likely just be to remove it, let the tests run, and test it manually on IE11 and Safari. If all that checks out then I'd probably call it a day.

sodatea:
After investigation, there's still one edge case: https://github.com/babel/babel/issues/9277

So I think I'll go for this PR instead of removing transpilation at the moment.

lucaswerkmeister:
Cool, thanks :)

PR #5580, opened 2020-06-13
Fix: stop ignoring --config-file cypress option

ahderman:
Since we always call `removeArg(rawArgs, 'config')`, and since the `matchRE` matches `--config` as well as `--config-file`, we end up ignoring the `--config-file` option if it is provided (e.g. `--config-file my-config.json`).

This commit fixes this by ensuring that calling `removeArg('config')` removes `--config` without removing `--config-file`.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


ahderman:
@sodatea Tests fixed.

PR #5581, opened 2020-06-14
Support ESLint + Coffeescript

helixbass:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No (at least in theory I don't think it should have to, but may need clarification)

**Other information:**

Now that ESLint is legitimately supported for Coffeescript as of Coffeescript v2.5.0 (via [`eslint-plugin-coffee`](https://github.com/helixbass/eslint-plugin-coffee)), people using Coffeescript + Vue should be able to successfully run ESLint against their Coffeescript files

But there have been a few issues encountered (see https://github.com/helixbass/eslint-plugin-coffee/issues/27)

Specifically here, I needed to adjust the Vue CLI ESLint config to run against `.coffee` files

These changes (plus https://github.com/mysticatea/vue-eslint-parser/pull/62) got ESLint working with Coffeescript in an example Vue project: https://github.com/helixbass/vue-coffee-eslint-example

But I'm not sure if further refinement is needed here for this to be flexible enough eg for projects that include `.coffee` files but don't have ESLint configured for Coffeescript, or for "hybrid" projects that include both Coffeescript + JS?

### DIFF
--- a/packages/@vue/cli-plugin-eslint/index.js
+++ b/packages/@vue/cli-plugin-eslint/index.js
@@ -41,7 +41,7 @@ module.exports = (api, options) => {
             .add(/node_modules/)
             .add(path.dirname(require.resolve('@vue/cli-service')))
             .end()
-          .test(/\.(vue|(j|t)sx?)$/)
+          .test(/\.(vue|(j|t)sx?|coffee)$/)
           .use('eslint-loader')
             .loader(require.resolve('eslint-loader'))
             .options({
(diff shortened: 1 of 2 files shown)


packages/@vue/cli-plugin-eslint/index.js
```diff
@@ -41,7 +41,7 @@ module.exports = (api, options) => {
             .add(/node_modules/)
             .add(path.dirname(require.resolve('@vue/cli-service')))
             .end()
-          .test(/\.(vue|(j|t)sx?)$/)
+          .test(/\.(vue|(j|t)sx?|coffee)$/)
```
sodatea, line 44: This should also be the responsibility of the `coffee` plugin.
Can be achieved with this statement in `chainWebpack`:

```js
config.module.rule('eslint').set('test', /\.(vue|(j|t)sx?|coffee)$/)
```

Issue #5411, opened 2020-04-21
vue cli fails to get packages when npm is using a custom nexus npm registry


Ealdst:
### Version
4.3.1

### Environment info
```
System:
    OS: Windows 10 10.0.18362
    CPU: (8) x64 Intel(R) Xeon(R) CPU E3-1505M v6 @ 3.00GHz
  Binaries:
    Node: 12.16.1 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.16.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.14.4 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: 44.18362.449.0
  npmPackages:
    @fortawesome/vue-fontawesome: ^0.1.9 => 0.1.9
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0
    @vue/babel-plugin-transform-vue-jsx:  1.1.2
    @vue/babel-preset-app:  4.3.1
    @vue/babel-preset-jsx:  1.1.2
    @vue/babel-sugar-functional-vue:  1.1.2
    @vue/babel-sugar-inject-h:  1.1.2
    @vue/babel-sugar-v-model:  1.1.2
    @vue/babel-sugar-v-on:  1.1.2
    @vue/cli-overlay:  4.3.1
    @vue/cli-plugin-babel: ^4.3.1 => 4.3.1
    @vue/cli-plugin-eslint: ^4.3.1 => 4.3.1
    @vue/cli-plugin-pwa: ^4.3.1 => 4.3.1
    @vue/cli-plugin-router:  4.3.1
    @vue/cli-plugin-unit-jest: ^4.3.1 => 4.3.1
    @vue/cli-plugin-vuex:  4.3.1
    @vue/cli-service: ^4.3.1 => 4.3.1
    @vue/cli-shared-utils:  4.3.1
    @vue/component-compiler-utils:  3.1.2
    @vue/eslint-config-airbnb: ^5.0.2 => 5.0.2
    @vue/preload-webpack-plugin:  1.1.1
    @vue/test-utils: ^1.0.0-beta.32 => 1.0.0-beta.32
    @vue/web-component-wrapper:  1.2.0
    eslint-plugin-vue: ^6.2.2 => 6.2.2
    jest-serializer-vue:  2.0.2
    svg-to-vue:  0.6.0
    vue: ^2.6.11 => 2.6.11
[...]

PR #5609, opened 2020-06-23
fix: fix support for some legacy registry servers

sodatea:
Some registry servers do not recognize `application/vnd.npm.install-v1+json`
in the Accept header, so here we add `application/json` as fallback.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli/lib/util/ProjectPackageManager.js
+++ b/packages/@vue/cli/lib/util/ProjectPackageManager.js
@@ -248,7 +248,7 @@ class PackageManager {
 
     const headers = {}
     if (!full) {
-      headers.Accept = 'application/vnd.npm.install-v1+json'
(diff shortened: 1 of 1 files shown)


sodatea:
I've tested with [verdaccio](https://github.com/verdaccio/verdaccio) and the latest version of [Sonatype Nexus3](https://github.com/sonatype/docker-nexus3), both recognized the `application/vnd.npm.install-v1+json` header. But there may be some legacy versions doesn't support it, as reported in this issue: https://github.com/vuejs/vue-cli/issues/5411#issuecomment-647586698

PR #5629, opened 2020-06-28
refactor: replace jscodeshift with vue-codemod

sodatea:
By removing the dependency on `vue-jscodeshift-adapter`, the global `vue-template-compiler` peer dependency error should be fully fixed.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


sodatea:
TODO: vue-codemod needs to support Node.js v8.9+

screetBloom:
When is V3 expected to be supported?

@sodatea

sodatea:
What do you mean? Vue 3 is already supported

screetBloom:
> What do you mean? Vue 3 is already supported

@vue/cli 3.x  still relies on the **vue-jscodeshift-adapter.**
When will this package be replaced with @vue/codemod?

@sodatea

sodatea:
I don't plan to add new features to 3.x. So it won't.

screetBloom:
> I don't plan to add new features to 3.x. So it won't.

Brother, why(大哥，不是吧)? 😂 

I created a tool that relies on both 3.x and 4.x. And the vue-jscodeshift-adapter may cause a mismatch between the vue and vue-template-compiler. 
So would it be possible to support that in version 3，please

@sodatea

sodatea:
But `vue-template-compiler` is required in a Vue 2.x project. How does it become a problem?

screetBloom:
Okay, let me describe the problem we're having.

- Generally, we will select a fixed vue/vue-template-compiler version number. For example, our team chose 2.6.11
- @vue/cli 3.x still relies on the vue-jscodeshift-adapter. And vue-jscodeshift-adapter has a vue-template-compiler@^2.5.13 installed by itself.
- That means vue-jscodeshift-adapter will try to install the latest version of vue-template-compiler when it is available. Such as 2.6.12
- This may result in an error: version mismatch

Personally, I wish @vue/cli would get rid of the third-party dependency on the vue-template-compiler@^2.5.13

@sodatea

sodatea:
1. Please use the `resolutions` field. As far as I tested, npm can correctly dedupe `2.6.11` and `^2.5.13`, so only Yarn has this issue, and Yarn supports [selective version resolutions](https://classic.yarnpkg.com/en/docs/selective-version-resolutions).
2. Please consider update Vue to 2.6.12, it's basically the same as 2.6.11

PR #5694, opened 2020-07-18
[Fix] common misspelling errors

Necmttn:
For reference: https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines

### DIFF
--- a/packages/@vue/cli-plugin-eslint/generator/index.js
+++ b/packages/@vue/cli-plugin-eslint/generator/index.js
@@ -43,7 +43,7 @@ module.exports = (api, { config, lintOn = [] }, _, invoking) => {
     pkg.gitHooks = {
       'pre-commit': 'lint-staged'
     }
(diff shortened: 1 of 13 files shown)


packages/@vue/cli-plugin-eslint/generator/index.js
```diff
@@ -43,7 +43,7 @@ module.exports = (api, { config, lintOn = [] }, _, invoking) => {
     pkg.gitHooks = {
       'pre-commit': 'lint-staged'
     }
-    const extentions = require('../eslintOptions').extensions(api)
+    const extensions = require('../eslintOptions').extensions(api)
       .map(ext => ext.replace(/^\./, ''))  // remove the leading `.`
     pkg['lint-staged'] = {
       [`*.{${extentions.join(',')}}`]: ['vue-cli-service lint', 'git add']
```
sodatea, line 49: Thanks! But you need to fix this line too.

Issue #5695, opened 2020-07-18
E2E tests fails on projects generated with typescript and vue-router.


TETRA2000:
### Version
4.4.6

### Reproduction link
[https://github.com/TETRA2000/vue-ts-app](https://github.com/TETRA2000/vue-ts-app)

### Environment info
```
Environment Info:

  System:
    OS: macOS 10.15.6
    CPU: (8) x64 Intel(R) Core(TM) i5-8257U CPU @ 1.40GHz
  Binaries:
    Node: 12.16.3 - ~/.nodenv/versions/12.16.3/bin/node
    Yarn: 1.22.4 - /usr/local/bin/yarn
    npm: 6.14.4 - ~/.nodenv/versions/12.16.3/bin/npm
  Browsers:
    Chrome: 84.0.4147.89
    Edge: 84.0.522.40
    Firefox: 78.0.2
    Safari: 13.1.2
  npmPackages:
    @vue/cli-overlay:  4.4.6
    @vue/cli-plugin-e2e-cypress: ~4.4.0 => 4.4.6
    @vue/cli-plugin-router: ~4.4.0 => 4.4.6
    @vue/cli-plugin-typescript: ~4.4.0 => 4.4.6
    @vue/cli-plugin-unit-mocha: ~4.4.0 => 4.4.6
    @vue/cli-plugin-vuex:  4.4.6
    @vue/cli-service: ~4.4.0 => 4.4.6
    @vue/cli-shared-utils:  4.4.6
    @vue/component-compiler-utils:  3.1.2
    @vue/preload-webpack-plugin:  1.1.1
    @vue/test-utils: ^1.0.3 => 1.0.3
    @vue/web-component-wrapper:  1.2.0
    typescript: ~3.9.3 => 3.9.7
    vue: ^2.6.11 => 2.6.11
    vue-class-component: ^7.2.3 => 7.2.3
    vue-hot-reload-api:  2.3.4
    vue-loader:  15.9.3
    vue-property-decorator: ^8.4.2 => 8.5.1
    vue-router: ^3.2.0 => 3.3.4
    vue-style-loader:  4.1.2
    vue-template-compiler: ^2.6.11 => 2.6.11
    vue-template-es2015-compiler:  1.9.1
  npmGlobalPackages:
    @vue/cli: 4.4.6
```

### Steps to reproduce
1.Create a new project with these configuration.
```
 ◉ TypeScript
 ◉ Router
 ◉ Unit Testing
[...]

PR #5637, opened 2020-07-01
feat: allow choosing vue version on creation (and in presets)

sodatea:
This PR:
- adds a `vueVersion` field in the preset schema (accepts a string, defaults to `'2'`, and can be `'3'`).
- adds a new "Default (Vue 3 Preview)" preset
- allow choosing Vue version on `vue create` manual mode

Progress:
- [x] new preset field
- [x] `vue create` prompt
- [x] `@vue/cli-service`
- [x] ESLint plugin
- [x] Router plugin
- [x] Vuex plugin
- [x] TypeScript
- [x] Babel JSX support (merged #5665)
- [x] Jest plugin
- [x] Mocha plugin
- [x] Cypress plugin
- [x] NightWatch plugin
- [x] Tests for these templates
- [x] Prompts in UI
- [x] Test for prompts
- [x] <http://link.vuejs.org/feature-flags>

Manually tested (should write unit tests later):
- [x] `vue invoke` in a Vue 3 project
- [x] JSX & TSX

Maybe later:
- [ ] Wrap object exports with `defineComponent` calls in `.vue` files

**What kind of change does this PR introduce?** (check at least one)

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

**Other information:**

PR #5697, opened 2020-07-19
[RFC] fix: Generate TypeScript templete when using Router

TETRA2000:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [x] Yes
- [ ] No

**Other information:**

References #5695

I found that `@vue/cli-plugin-typescript` won't generate typescript vue templete while using `@vue/cli-plugin-router`.
Its seems overridden by router's template. 
(Please take a look at the issue for detail.)

I made small patch to modify plugin order before their execution.

This is just a PoC.
I don't think this is the best way to solve the problem.

I'd like to ask you to give me some advice.
Maybe it's better to add some methods to GeneratorAPI to manage execution order.

---

FYI: 

I tried to solve this by adding options to `sortObject` in `Creator.resolvePlugins` like this.
```javascript
  // { id: options } => [{ id, apply, options }]
  async resolvePlugins (rawPlugins, pkg) {
    // ensure cli-service is invoked first
    rawPlugins = sortObject(rawPlugins, ['@vue/cli-service', '@vue/cli-plugin-router', '@vue/cli-plugin-typescript'], true)
```

But somehow it didn't work as I intended. It generated same JS template.

I suspect that there is another dependencies on execution order in `generator` functions.

### DIFF
--- a/packages/@vue/cli/lib/Generator.js
+++ b/packages/@vue/cli/lib/Generator.js
@@ -3,6 +3,7 @@ const debug = require('debug')
 const GeneratorAPI = require('./GeneratorAPI')
 const PackageManager = require('./util/ProjectPackageManager')
 const sortObject = require('./util/sortObject')
+const sortArrayPartially = require('./util/sortArrayPartially')
 const writeFileTree = require('./util/writeFileTree')
 const inferRootOptions = require('./util/inferRootOptions')
 const normalizeFilePaths = require('./util/normalizeFilePaths')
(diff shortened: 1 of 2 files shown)


sodatea:
FYI it's been fixed as part of PR #5637 
[`df93f46` (#5637)](https://github.com/vuejs/vue-cli/pull/5637/commits/df93f46600cdc1013f85f99484a957cb04406566)

It's certainly not the best way to solve the problem but the plugin ordering API is not a priority at the moment so I think we'll live with the hack for a while.

TETRA2000:
@sodatea  Thank you for notifying me about that.
And I also found that my PR will generate wrong `App.vue` (which doesn't use `router-link` )

I'm gonna close this one.

Issue #5698, opened 2020-07-19
Update version of eslint-loader used in cli-plugin-eslint


Amndeep7:
### What problem does this feature solve?
vue cli (cli-plugin-eslint) 4.4.6 uses eslint-loader 2.2.1.  eslint-loader 2.2.1 does not support the 7.x release of eslint, which is the most recent version of eslint.  The latest version of eslint-loader is 4.0.2 which does support the 7.x release of eslint.  Can the version of eslint-loader used by cli-plugin-eslint be updated to the most recent version?  Word of caution that doing so will drop support for versions of eslint less than 6.x.  

### What does the proposed API look like?
No api change, just version change in package.json for cli-plugin-eslint.

zhangzhao666:
good

lianzhao:
Well, seems that `eslint-loader` [has been deprecated](https://github.com/webpack-contrib/eslint-loader/issues/337).

Maybe vue-cli v5 should migrate to [eslint-webpack-plugin](https://github.com/webpack-contrib/eslint-webpack-plugin)?

Amndeep7:
I'm for it, but it seems like eslint-webpack-plugin only supports eslint v7 so it will be even more of a breaking change.  It's also not a drop-in replacement so some changes to configuration will need to be made but the docs have a little bit of advice on how to migrate between them at least.

Amndeep7:
Should probably tag @sodatea as well to hear from them about when they think this should occur.

sodatea:
Done in [v5.0.0-alpha.0](https://github.com/vuejs/vue-cli/releases/tag/v5.0.0-alpha.0)

PR #5699, opened 2020-07-20
Update version of eslint-loader

Amndeep7:
Update version of eslint-loader to the most recent release so as to be able to use the most recent release of eslint.  See #5698

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [x] Yes
- [ ] No

**Other information:**
Will no longer support versions of eslint below 5.x.

### DIFF
--- a/packages/@vue/cli-plugin-eslint/package.json
+++ b/packages/@vue/cli-plugin-eslint/package.json
@@ -24,7 +24,7 @@
   },
   "dependencies": {
     "@vue/cli-shared-utils": "^4.4.6",
-    "eslint-loader": "^2.2.1",
+    "eslint-loader": "^4.0.2",
     "globby": "^9.2.0",
     "inquirer": "^7.1.0",
     "webpack": "^4.0.0",


sodatea:
I'm sorry but we are only able to update to eslint-loader v3 in this major, due to breaking changes.

Amndeep7:
3.0.4 of eslint-loader only supports the 5.x and 6.x branches of eslint.  This is still a change from the behavior for the version of eslint-loader currently being used (2.2.1) which is 1.6 through the 6.x branches of eslint.  What's the difference between dropping support for the 1.6-4.x branches of eslint and dropping support for the 1.6-5.x branches of eslint in terms of being a breaking change?

sodatea:
Oh sorry, I thought we've already dropped ESLint v4 support. I just double-checked and it's still supported. Then we can't update the loader to either version.

Amndeep7:
Thanks for the clarification @sodatea.  Is there any expectation for when the next major version update will be happening for this library?

PR #5776, opened 2020-08-11
refactor(wip): replace node-ipc with a cleanroom IPC implementation

### LINKED ISSUE #2621: Remove the four npm dependencies of which use software licenses that violate the tenets of open source and are incompatible with parent MIT licensed modules.
The following four deeply nested dependencies use DBAD licenses, which are incompatible with MIT Licenses, due to the fact that they state that a pint is owed to the maintainer of these modules if significant amount of wealth is made off of software that uses these modules.

Whether or not this is what the maintainer intended, the license implies that not sharing a pint with the maintainer is a direct violation of the license.

https://github.com/RIAEvangelist/node-ipc

https://github.com/RIAEvangelist/js-queue

https://github.com/RIAEvangelist/js-message

https://github.com/RIAEvangelist/easy-stack

I would like to recommend to the entire Vue.js community, that we take a stance similar to the one that Apache takes on nonsensical licenses:

Nonsensical licenses
These licenses while amusing to their creators are legally problematic. They often include subjective Field of use restrictions e.g. “Don’t be evil” with no arbiter for that subjective restriction defined. In some cases they may not even grant sufficient rights to conform to the OSI open source definition. Since we do not wish to surprise our downstream consumers we forbid the use of such licenses.

I am aware that there are a large amount of libraries used by the node.js community that have these modules as dependencies - however, I think this is due to unawareness not by choice. I am not suggesting we convince the entire node.js community to discontinue usage of these libraries, however I would like to encourage at least the Vue community look into finding out what options we have.

Everything I am saying is out of pure individual interest. I am not an informed lawyer. I do not represent the views of a company.

Licenses such as DBAD will cause useful libraries to be avoided by entire companies so as to steer clear of what is implied by the DBAD license terms.

If we can come up with a solution, we can benefit the community in two main ways:
1) Increase the amount of adoption of the Vue.js framework and all of its associated tools (For the companies that would only consider seriously adopting the framework if they also adopted usage of the vue-cli)
2) Remove the risk of legal implications and public embarrassment of lawsuits that could possibly occur over the terms that are implied in the DBAD license text.

Again- I do not represent the views of a company nor of a lawyer.

My main goal is to increase the adoption and usage of Vue.js and the Vue-cli.

If there is a better forum to bring this issue to the attention of the community, can someone help me do so? I have never had to do something like this before.

There is no API for this feature request. I was not sure if this should be opened as a Bug Report or Feature Request since those are the only two options on this form.

Regarding the approaches we could take to mitigate this problem:

1) Ask the maintainer to change the license of these four modules himself
     I know of one individual who asked already, and the maintainer didn't seem to want to drop the DBAD license completely. I'm not sure if the maintainer will listen if the entire community chimes in.

2) Ask the maintainer to dual license. If the maintainer wants to be able to make a statement with the DBAD license, he will still be able to do so - however adding a second license like the MIT license will remove the risk that is currently imposed on the rest of the community

3) Get the Vue-cli community to investigate and develop different solutions to our problems. If it comes to it, the only way to get away from using the software in these modules may be to remove the current usage of those for modules with the DBAD license. Other solutions will have to be considered. 

The downside of this last approach, is that finding alternative solutions may in fact be difficult.

https://dbad-license.org/

sodatea:
The packet format is compatible with the `node-ipc` default.

After this commit, no core plugins will rely on `node-ipc` to do anything.
But the UI still needs it to communicate with plugins.

TODOs:
- Fix UI build task
- Finish the IPC server implementation
- Refactor `@vue/cli-ui` to get rid of `node-ipc`, too
- Fully refactor the IpcMessenger module to address cross-platform
issues, etc.

Fixes #2621

It is a cleanroom implementation because I've never read a line of the `node-ipc` code. I've only read [the package's README](https://www.npmjs.com/package/node-ipc) before implementing this PR.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -4,7 +4,7 @@ const {
   hasProjectYarn,
   hasProjectPnpm,
   openBrowser,
-  IpcMessenger
+  NewIpcMessenger
 } = require('@vue/cli-shared-utils')
 
 const defaults = {
@@ -301,7 +301,7 @@ module.exports = (api, options) => {
 
           // Send final app URL
           if (args.dashboard) {
-            const ipc = new IpcMessenger()
+            const ipc = new NewIpcMessenger()
             ipc.send({
               vueServe: {
                 url: localUrlForBrowser
--- a/packages/@vue/cli-service/lib/webpack/DashboardPlugin.js
+++ b/packages/@vue/cli-service/lib/webpack/DashboardPlugin.js
@@ -7,14 +7,14 @@
 const path = require('path')
 const fs = require('fs-extra')
 const webpack = require('webpack')
-const { IpcMessenger } = require('@vue/cli-shared-utils')
+const { NewIpcMessenger } = require('@vue/cli-shared-utils')
 const { analyzeBundle } = require('./analyzeBundle')
 
 const ID = 'vue-cli-dashboard-plugin'
 const ONE_SECOND = 1000
 const FILENAME_QUERY_REGEXP = /\?.*$/
 
-const ipc = new IpcMessenger()
+const ipc = new NewIpcMessenger()
 
 function getTimeMessage (timer) {
   let time = Date.now() - timer
--- a/packages/@vue/cli-shared-utils/index.js
+++ b/packages/@vue/cli-shared-utils/index.js
@@ -2,6 +2,7 @@
   'env',
   'exit',
   'ipc',
+  'NewIpcMessenger',
   'logger',
   'module',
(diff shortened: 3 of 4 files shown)


bkhatkov:
Hi @sodatea. Is there any projection on when is this going to be merged? It is so important :)

RIAEvangelist:
node-ipc updated to MIT and latest buffer implementation.

Issue #5135, opened 2020-02-03
Yarn2 adaptation (vue cli failed to create project using yarn2)


Jancat:
### Version
4.1.2

### Environment info
```
Environment Info:

  System:
    OS: macOS 10.15.2
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
  Binaries:
    Node: 13.7.0 - /usr/local/bin/node
    Yarn: 2.0.0-rc.27 - /usr/local/bin/yarn
    npm: 6.13.7 - /usr/local/bin/npm
  Browsers:
    Chrome: 79.0.3945.130
    Firefox: Not Found
    Safari: 13.0.4
  npmGlobalPackages:
    @vue/cli: 4.1.2
```

### Steps to reproduce
1. Run the `vue create demo` command, using the following presets:

```json
"demo-preset": {
      "useConfigFiles": true,
      "plugins": {
        "@vue/cli-plugin-babel": {},
        "@vue/cli-plugin-typescript": {
          "classComponent": true,
          "useTsWithBabel": true
        },
        "@vue/cli-plugin-pwa": {},
        "@vue/cli-plugin-router": {
          "historyMode": true
        },
        "@vue/cli-plugin-vuex": {},
        "@vue/cli-plugin-eslint": {
          "config": "prettier",
          "lintOn": [
            "save",
            "commit"
          ]
        },
        "@vue/cli-plugin-unit-mocha": {}
      },
      "cssPreprocessor": "dart-sass"
    }
```

### What is expected?
Create the project and run `yarn serve` successfully.

### Version
4.1.2

### Environment info
```
Environment Info:

  System:
    OS: macOS 10.15.2
    CPU: (12) x64 Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
  Binaries:
    Node: 13.7.0 - /usr/local/bin/node
    Yarn: 2.0.0-rc.27 - /usr/local/bin/yarn
    npm: 6.13.7 - /usr/local/bin/npm
  Browsers:
    Chrome: 79.0.3945.130
    Firefox: Not Found
    Safari: 13.0.4
  npmGlobalPackages:
    @vue/cli: 4.1.2
```

### Steps to reproduce
1. Run the `vue create demo` command, using the following presets:

```json
"demo-preset": {
      "useConfigFiles": true,
      "plugins": {
        "@vue/cli-plugin-babel": {},
        "@vue/cli-plugin-typescript": {
[...]

PR #5794, opened 2020-08-16
fix(cli): resolve plugins relative to the package context

merceyz:
`@vue/cli-service` is trying to resolve plugins on behalf of itself instead of the `pkgContext` which makes it rely on hoisting to bring it to an accessible location which is not guaranteed, especially in monorepos

https://yarnpkg.com/advanced/rulebook#packages-should-only-ever-require-what-they-formally-list-in-their-dependencies

Ref https://github.com/yarnpkg/berry/issues/964
Ref https://github.com/vuejs/vue-cli/issues/5135

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

### DIFF
--- a/packages/@vue/cli-service/lib/Service.js
+++ b/packages/@vue/cli-service/lib/Service.js
@@ -176,11 +176,10 @@ module.exports = class Service {
             this.pkg.optionalDependencies &&
             id in this.pkg.optionalDependencies
           ) {
-            let apply = () => {}
-            try {
-              apply = loadModule(id, this.pkgContext)
-            } catch (e) {
+            let apply = loadModule(id, this.pkgContext)
+            if (!apply) {
               warn(`Optional dependency ${id} is not installed.`)
+              apply = () => {}
             }
 
             return { id, apply }
--- a/packages/@vue/cli-service/lib/Service.js
+++ b/packages/@vue/cli-service/lib/Service.js
@@ -143,9 +143,9 @@ module.exports = class Service {
   }
 
   resolvePlugins (inlinePlugins, useBuiltIn) {
-    const idToPlugin = id => ({
+    const idToPlugin = (id, absolutePath) => ({
       id: id.replace(/^.\//, 'built-in:'),
-      apply: require(id)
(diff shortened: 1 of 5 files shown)


yoyo930021:
#5135

merceyz:
Not sure how to deal with the failing tests, @sodatea could I get some assistance please?

merceyz:
Tests finally pass :tada:

packages/@vue/cli-service/lib/Service.js
```diff
@@ -178,14 +178,14 @@ module.exports = class Service {
           ) {
             let apply = () => {}
             try {
-              apply = require(id)
+              apply = require(resolveModule(id, this.pkgContext))
```
fangbinwei, line 181: Is it better to use `loadModule` directly?  `require(resolveModule(..` looks a little redundant to me

packages/@vue/cli-service/lib/Service.js
```diff
@@ -178,14 +178,14 @@ module.exports = class Service {
...
             } catch (e) {
               warn(`Optional dependency ${id} is not installed.`)
             }
 
             return { id, apply }
           } else {
-            return idToPlugin(id)
+            return idToPlugin(id, resolveModule(id, this.pkgContext))
```
fangbinwei, line 188: `idToPlugin(resolveModule(id, this.pkgContext))` is better for me. And we don't need to modify the `idToPlugin` function.

merceyz: I can't do that as it would change the `id` in the object

fangbinwei: I think I ignored something about 'id', will check it later

merceyz:
@sodatea Is there a chance this could be backported to v4?

scinos:
Happy to work in the backport, if we are willing to accept it.

PR #5847, opened 2020-09-04
Updated netlify docs for vue/cli-plugin-pwa

sarbikbetal:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Added a small fix related to PWAs deployed on Netlify. If you don't exclude the `_redirects` file, the service worker will try to cache it but it won't be able to find the file in the netlify servers and as a result the service worker installation will fail, which will break all the PWA functionalities, so we need to exclude the file through workbox exclude option.

### DIFF
--- a/docs/guide/deployment.md
+++ b/docs/guide/deployment.md
@@ -145,7 +145,7 @@ Typically, your static website will be hosted on https://yourUserName.gitlab.io/
 
 
 ```javascript
-// vue.config.js file to be place in the root of your repository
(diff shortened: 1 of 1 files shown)


gustawdaniel (approved):

sodatea (approved):
Thanks!

PR #5852, opened 2020-09-07
fix: fix duplicate id="app" in Vue 3 project template

### LINKED ISSUE #5813: Duplicate root app (id="app") when creating vue3 project
### Version
4.5.4

### Reproduction link
[https://github.com/kirianguiller/vue3-issue](https://github.com/kirianguiller/vue3-issue)

### Environment info
```
  System:
    OS: Linux 5.4 Ubuntu 20.04 LTS (Focal Fossa)
    CPU: (8) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
  Binaries:
    Node: 10.19.0 - /usr/bin/node
    Yarn: Not Found
    npm: 6.14.7 - /usr/local/bin/npm
  Browsers:
    Chrome: Not Found
    Firefox: 79.0
  npmGlobalPackages:
    @vue/cli: 4.5.4

```

### Steps to reproduce
- Create a vue3 app with vue CLI 4.5.4
- Run 'npm run serve'

### What is expected?
Only one "id='app'" html element

Two "id='app'" html element

---
These duplicates are an issue i if we add some CSS styling for padding (for instance) as they will be duplicated and the padding will be summed. 

I don't know if the problem come from the vue@cli default setup or from the vuejs3 mounting method.

PS : This duplication is not occuring for vue2 app.

sodatea:
closes #5813

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-plugin-router/__tests__/routerGenerator.spec.js
+++ b/packages/@vue/cli-plugin-router/__tests__/routerGenerator.spec.js
@@ -85,6 +85,8 @@ test('use with Vue 3', async () => {
 
   expect(files['src/main.js']).toMatch('.use(router)')
 
+  expect(files['src/App.vue']).not.toMatch('<div id="app">')
+
   expect(pkg.dependencies).toHaveProperty('vue-router')
   expect(pkg.dependencies['vue-router']).toMatch('^4')
 })
--- a/packages/@vue/cli-plugin-router/generator/template-vue3/src/App.vue
+++ b/packages/@vue/cli-plugin-router/generator/template-vue3/src/App.vue
@@ -1,0 +1,63 @@
+---
+extend: '@vue/cli-service/generator/template/src/App.vue'
+replace:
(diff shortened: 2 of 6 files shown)


9mm:
Just wondering, does this remove one of the #app divs from index.html or App.vue? Or, does it cause the App.vue node to replace the index.html node like in vue 2?

I tried to install vue-cli from master to check myself, but it complains about version mismatch.

sodatea:
From `App.vue`, the on in the HTML still needs to exist to be served as a mounting point.

HackyDev:
The new version of `App.vue` has styles for the `#app` container declared in the `index.html`. It doesn't feel right, because the component applies styles to an element declared outside its template.

titusdecali:
I agree with @HackyDev, this is very strange and should be reduced to a single #app.

PR #5918, opened 2020-10-01
fixed #git-command-not-execute-in-script-issue

Azamat-Alibaev:
to enable execute git in script

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**


sodatea:
Could you elaborate on this line? What exact problem does it solve? I don't get it.

Also, please target the PR to the `master` branch.

Azamat-Alibaev:
It enables running git commands from script. Without this line git commands
in deploy.sh script just skipped on my linux machine.

пт, 9 окт. 2020 г., 11:08 Haoqun Jiang <notifications@github.com>:

Azamat-Alibaev:
I found solution for simular issue here:

https://stackoverflow.com/questions/6122932/execute-git-command-inside-bash-script

Thanks for patience.

сб, 10 окт. 2020 г. в 15:09, Нурия Алибаева <alibaevan3@gmail.com>:

sodatea:
How does that work? If I understand correctly, `env -i` only works in the way with a command as its argument, such as `env -i git pull ...`. It does not have any effect as a standalone command.

sodatea:
I think the documentation is not the best place to put such a workaround solution.

The actual problem is the unexpected environment variables in your shell.
`env -i git` is only a workaround, not a solution.
And if we put that line on, all other code examples that could possibly be affected by unexpected environment variables would also have to put such prefixes on. This would confuse users.

BTW, such problems won't occur if the app's deployed via TravisCI, because they would have set up a clean environment.

Azamat-Alibaev:
Ok, thank you.

сб, 10 окт. 2020 г., 12:06 Haoqun Jiang <notifications@github.com>:

PR #5972, opened 2020-10-16
fix: prioritize node_modules of current project

pedrothome1:
This PR changes the order of searching for node_modules, putting the relative path as the last one. It avoids, after all, duplication of dependencies.

Related Stackoverflow answer: https://stackoverflow.com/a/63992540/14397787
Webpack reference: https://v4.webpack.js.org/configuration/resolve/#resolvemodules

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
None

### DIFF
--- a/packages/@vue/cli-service/lib/config/base.js
+++ b/packages/@vue/cli-service/lib/config/base.js
@@ -57,9 +57,9 @@ module.exports = (api, options) => {
         .merge(['.mjs', '.js', '.jsx', '.vue', '.json', '.wasm'])
         .end()
       .modules
-        .add('node_modules')
         .add(api.resolve('node_modules'))
         .add(resolveLocal('node_modules'))
+        .add('node_modules')
         .end()
       .alias
         .set('@', api.resolve('src'))


merceyz:
It shouldn't be necessary to add these at all, webpack normally resolves X from Y which your package manager is responsible for ensuring is the right version

pedrothome1:
Why would it be necessary or useful to have two versions of vue or some other package? Shouldn't it prioritize the closest package.json since it's in the same directory of the application?

merceyz:
`vue`, much like `react`, is a special case in that it should be a singleton but for other packages no it shouldn't, it should follow the `node_modules` resolution algorithm and use the closest `node_modules` folder to the file. 

Packages having the same name does not mean it's the same version, for example take this dependency tree:
```
.
├── node_modules/
│   ├── lodash@4
│   └── some-lib/
│       ├── index.js
│       └── node_modules/
│           └── lodash@3
└── main.js
```
With your changes both `main.js` and `index.js` would get `lodash@4` which is incorrect

pedrothome1:
Ok, I got it. Thank you.

PR #5998, opened 2020-10-23
fix: add hash to filename in other mode

alinwyr:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
Hash exist in "NODE_ENV=production" only. It make other modes, ".env.testing" or ".env.study" run "vue-cli-service build --mode testing" or "vue-cli-service build --mode study" and get the file without hash key.


sodatea:
It's by design.
`isProd` is calculated from `process.env.NODE_ENV`. You can set `NODE_ENV` to `production` under any mode.

PR #6000, opened 2020-10-24
fix: prevent snoretoast shortcut, set notif title (#2720)

### LINKED ISSUE #2720: SnoreToast is added to the start menu when notifying on Windows
### Version
3.0.4

### Node and OS info
Node v9.4.0, npm v5.6.0, Windows 10

### Steps to reproduce
- Install `@vue/cli` on Windows
- Run `vue ui` on a project
- Install/update/etc a dependency in the UI
  - Note that no notification was triggered and that `%APPDATA%\Microsoft\Windows\Start Menu\Programs` now has a SnoreToast shortcut
- Install/update/etc another dependency
  - Note that a notification was triggered successfully

### What is expected?
SnoreToast would not have a shortcut since the user may not like its addition to their start menu without indication of its purpose.

Additionally, the notification should trigger the first time, but I'm guessing this is caused by SnoreToast failing to notify without an app ID in the start menu to use for the notification.

The shortcut is installed without an opt-in, notification fails the first time

---
The UI's notification API uses [`node-notifier`](https://github.com/mikaelbr/node-notifier) which in turn uses [`SnoreToast`](https://github.com/KDE/snoretoast) for native Windows notifications.

[SnoreToast seems to indicate that a program in the start menu is necessary for notifications](https://github.com/KDE/snoretoast/blob/master/README.md), though I can't find anything to back up that it needs to be in the start menu in case SnoreToast can be used from the `node_modules` folder.

If the CLI was an Electron app, the intended use of `node-notifier` would be to use the shortcut for the CLI instead (which would fix also fix the issue with it not firing and it saying the wrong app name when notifying). Since there isn't a shortcut for the CLI, it sounds like a workaround would need to be found to keep the implementation the same.

The issue with it not firing the first time is most likely related to and could be fixed by "installing" the shortcut before the first notification.

tony19:
This prevents SnoreToast from installing a Start Menu shortcut by specifying an `appID`. The `appID` is also used to set the
notification's title bar text in Windows 10. That text is now set to "Vue UI", which replaces the default value of "SnoreToast".

fixes #2720

*Screenshot of desktop notification on Windows 10:*
desktop-notif

*Screenshot of notification bar on Windows 10:*
notif-bar

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-ui/apollo-server/util/notification.js
+++ b/packages/@vue/cli-ui/apollo-server/util/notification.js
@@ -6,8 +6,16 @@ const builtinIcons = {
   error: path.resolve(__dirname, '../../src/assets/error.png')
 }
 
+// https://github.com/mikaelbr/node-notifier/issues/154
+// Specify appID to prevent SnoreToast shortcut installation.
+// SnoreToast actually uses it as the string in the notification's
+// title bar (different from title heading inside notification).
+// This only has an effect in Windows.
+const snoreToastOptions = notifier.Notification === notifier.WindowsToaster && { appID: 'Vue UI' }
+
 exports.notify = ({ title, message, icon }) => {
   notifier.notify({
+    ...snoreToastOptions,
     title,
     message,
     icon: builtinIcons[icon] || icon


packages/@vue/cli-ui/apollo-server/util/notification.js
```diff
@@ -6,8 +6,16 @@ const builtinIcons = {
   error: path.resolve(__dirname, '../../src/assets/error.png')
 }
 
+// https://github.com/mikaelbr/node-notifier/issues/154
+// Specify appID to prevent SnoreToast shortcut installation.
+// SnoreToast actually uses it as the string in the notification's
+// title bar (different from title heading inside notification).
+// This only has an effect in Windows.
+const snoreToastOptions = notifier.Notification === notifier.WindowsToaster && { appID: 'Vue UI' }
```
tony19, line 14: [`notifier` is set to `WindowsToaster`](https://github.com/mikaelbr/node-notifier/blob/29b99f2855f09ed3ee573e0410eb4ea4553e10f0/index.js#L30), which [uses SnoreToast](https://github.com/mikaelbr/node-notifier/blob/29b99f2855f09ed3ee573e0410eb4ea4553e10f0/notifiers/toaster.js#L5)


I only set `appID` for SnoreToast to avoid any effects on other notifiers that might also have an `appID` parameter.

tony19:
Once this fix lands, users may want to manually undo the SnoreToast changes from their system, as described below.

## Remove SnoreToast shortcut from *Start Menu*

#### Option A: File Explorer

 1. Right-click **_Start Menu_ > _SnoreToast_**, and choose **_More_ > _Open File Location_**, which opens *File Explorer*. 
 2. In *File Explorer*, delete the root folder of *SnoreToast*.

#### Option B: Windows Command Line

 1. Open [`cmd.exe`](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd)
 2. Run this [`rmdir`](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/rmdir) command:

         rmdir /s /q "%APPDATA%\Microsoft\Windows\Start Menu\Programs\SnoreToast"

## Remove SnoreToast's registry key

*Note:* [SnoreToast sets a registry key](https://github.com/KDE/snoretoast/blob/8b3dc33a81126381a333af099cfb098d571f39e9/src/linkhelper.cpp#L91) in order to register its notification center. The UUID seen in the key is for [SnoreToast v0.7.0](https://github.com/KDE/snoretoast/blob/5aee2959cc107b69acbaedb3068de9d85b54666e/CMakeLists.txt#L5).

#### Option A: Registry Editor

 1. Open [`regedit.exe`](https://ss64.com/nt/regedit.html)
 2. In the location textbox at the top right below the menu bar, enter: `HKCU\Software\Classes\CLSID\{eb1fdd5b-8f70-4b5a-b230-998a2dc19303}` to jump to the key in the registry.
 3. Right-click the `{eb1fdd5b-8f70-4b5a-b230-998a2dc19303}` folder, and choose `Delete`.

#### Option B: Windows Command Line

 1. Open `cmd.exe`
 2. Run this [`reg`](https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/reg) command:

         reg delete /f \Software\Classes\CLSID\{eb1fdd5b-8f70-4b5a-b230-998a2dc19303}

mellester:
This is a good thing to have

sodatea (approved):
Thank you so much for the fix and the detailed guidance!

PR #6001, opened 2020-10-24
feat: open browser when toast clicked

tony19:
Currently, clicking the toast notification from Vue UI has no effect other than closing the notification. This PR opens a browser to the originating Vue UI upon notification click.

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

`node-notifier` 5/6 has a [bug in Windows that does not forward notification-click events](https://github.com/mikaelbr/node-notifier/issues/291), so this change will only work on other platforms until that gets fixed.


packages/@vue/cli-ui/apollo-server/util/notification.js
```diff
@@ -6,10 +6,15 @@ const builtinIcons = {
   error: path.resolve(__dirname, '../../src/assets/error.png')
 }
 
+let notifCallback = null
```
tony19, line 9: `notifCallback` can be `null`, which is safe because [`node-notifier` defaults it to a no-op](https://github.com/mikaelbr/node-notifier/blob/29b99f2855f09ed3ee573e0410eb4ea4553e10f0/notifiers/toaster.js#L54).

tony19:
Actually, this change has a bug in that it re-opens the browser when the notification is automatically dismissed from a timeout, which could be annoying. I'll look into it.

sodatea:
Thanks! This PR looks good. Could you please fix the merge conflicts here so that I can merge it?

tony19:
@sodatea Ok, resolved merged conflicts. Thanks!

sodatea (approved):
Thanks!

Issue #6021, opened 2020-10-28
Prettier gives error when using unsafe `any`


jeneser:
### Version
4.5.8

### Reproduction link
[https://github.com/jeneser/vue-cli-prettier-issue](https://github.com/jeneser/vue-cli-prettier-issue)

### Environment info
```
Environment Info:

  System:
    OS: Linux 4.14 Ubuntu 20.04 LTS (Focal Fossa)
    CPU: (16) x64 Intel(R) Xeon(R) CPU @ 2.20GHz
  Binaries:
    Node: 12.19.0 - ~/.nvm/versions/node/v12.19.0/bin/node
    Yarn: 1.22.10 - ~/.nvm/versions/node/v12.19.0/bin/yarn
    npm: 6.14.8 - ~/.nvm/versions/node/v12.19.0/bin/npm
  Browsers:
    Chrome: Not Found
    Firefox: Not Found
  npmPackages:
    @vue/babel-helper-vue-jsx-merge-props:  1.2.1 
    @vue/babel-helper-vue-transform-on:  1.0.0-rc.2 
    @vue/babel-plugin-jsx:  1.0.0-rc.3 
    @vue/babel-plugin-transform-vue-jsx:  1.2.1 
    @vue/babel-preset-app:  4.5.8 
    @vue/babel-preset-jsx:  1.2.4 
    @vue/babel-sugar-composition-api-inject-h:  1.2.1 
    @vue/babel-sugar-composition-api-render-instance:  1.2.4 
    @vue/babel-sugar-functional-vue:  1.2.2 
    @vue/babel-sugar-inject-h:  1.2.2 
    @vue/babel-sugar-v-model:  1.2.3 
    @vue/babel-sugar-v-on:  1.2.3 
    @vue/cli:  4.5.8 
    @vue/cli-overlay:  4.5.8 
    @vue/cli-plugin-babel:  4.5.8 
    @vue/cli-plugin-eslint:  4.5.8 
    @vue/cli-plugin-router:  4.5.8 
    @vue/cli-plugin-vuex:  4.5.8 
    @vue/cli-service:  4.5.8 
    @vue/cli-service-global:  4.5.8 
    @vue/cli-shared-utils:  4.5.8 
    @vue/cli-ui:  4.5.8 
    @vue/cli-ui-addon-webpack:  4.5.8 
    @vue/cli-ui-addon-widgets:  4.5.8 
    @vue/compiler-core:  3.0.2 
[...]

PR #6023, opened 2020-10-28
fix: update default vue 3 typescript shim

jeneser:
#6021  eslint warns about "unsafe any" in default shim when formatting.

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-plugin-typescript/codemods/__testfixtures__/shims-vue.output.ts
+++ b/packages/@vue/cli-plugin-typescript/codemods/__testfixtures__/shims-vue.output.ts
@@ -1,5 +1,5 @@
 declare module '*.vue' {
   import { DefineComponent } from 'vue';
-  const component: DefineComponent<{}, {}, any>;
+  const component: DefineComponent<{}, {}, unknown>;
   export default component;
 }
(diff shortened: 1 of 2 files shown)


sodatea:
@cexbrayat @pikax I'm no expert in TypeScript so I'm not sure which is the better way to shim `.vue` type, using `unknown` or `any` with an `eslint-disable-line` comment. Do you have any opinion on this?

From what I understand, `unknown` is not assignable to any other type, so seems using the `unknown` type may break some real-world use cases? I haven't encountered any but I'm concerned about that.

cexbrayat:
I think `unknown` is going to be painful, as `wrapper.vm.something` will be typed as `never` in unit tests, forcing developers to cast `vm` themselves every time. I do agree that `any` is not ideal, but it allows to write unit tests without too much trouble. I sadly don't think we can do better right now.

2 possible ways to deal with the issue with eslint:
- add a simple `eslint-disable-next-line no-any` in the shim
- or exclude `d.ts` files from eslint (my favorite, as it makes no real sense for eslint to analyze the shim)

bodograumann:
The rule could also be `@typescript-eslint/no-explicit-any`.

jubairsaidi (commented):
consider changing the shim to:
```typescript
declare module '*.vue' {
  import type { DefineComponent } from 'vue';
  const component: DefineComponent<
    Record<string, unknown>,
    Record<string, unknown>,
    unknown
  >;
  export default component;
}
```

gbumanzordev (approved):

PR #6044, opened 2020-11-09
fix: NODE_ENV cannot be correctly set to "development"

cpg0525:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
When the environment is "dev", NODE_ENV cannot be correctly set to "development"

### DIFF
--- a/packages/@vue/cli-service/lib/Service.js
+++ b/packages/@vue/cli-service/lib/Service.js
@@ -121,7 +121,8 @@ module.exports = class Service {
       const defaultNodeEnv = (mode === 'production' || mode === 'test')
         ? mode
         : 'development'
-      if (shouldForceDefaultEnv || process.env.NODE_ENV == null) {
+        // fix: When the environment is "dev", NODE_ENV cannot be correctly set to "development"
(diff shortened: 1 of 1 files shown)


fangbinwei:
Why check if NODE_ENV is'dev'?

cpg0525:
> Why check if NODE_ENV is'dev'?

``

> Why check if NODE_ENV is'dev'?

Ensure that process.env.NODE_ENV is always development when the node environment is dev and config.get can read the development.json configuration correctly.

cpg0525:
> Why check if NODE_ENV is'dev'?

Ensure that process.env.NODE_ENV is always development when the node environment is dev and config.get can read the development.json configuration correctly.Therefore, I think it is necessary to do an operation compatible with process.env.node _ env = development.

sodatea:
Setting `NODE_ENV` to `'dev'` is not a good practice.
Doing an implicit conversion is not a good practice either.

Users would be extremely confused when debugging problems involving this conversion.

Better warn the user than silently convert the environment variable.

PR #6057, opened 2020-11-14
fix(cypress): make cypress a peer dependency

elevatebart:
Cypress has a very stable CLI API.
Every 12 weeks Cypress releases a new Major version with
breaking changes to the JS tests and config.
But almost no changes to the CLI API.
This rhythm of Major updates does not match the slower pace of
the vue CLI.

To allow more nimble updates by end users, we add
cypress as a true devDependency to the user package.json
and as a peerDependency to send a warning if we remove it.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [x] Yes
- [ ] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-plugin-e2e-cypress/generator/index.js
+++ b/packages/@vue/cli-plugin-e2e-cypress/generator/index.js
@@ -7,6 +7,9 @@ module.exports = api => {
   api.extendPackage({
     scripts: {
       'test:e2e': 'vue-cli-service test:e2e'
+    },
+    devDependencies: {
+      'cypress': '^5.0.0'
     }
   })
 }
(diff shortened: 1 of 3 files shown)


elevatebart:
The breaking change is that user with the plugin currently installed will have to install the plugin manually if they update.
It is dangerous.

Since we now have 3 Vue people at cypress, we were thinking that we could pull our weight and create a plugin that we maintain.

I would call it vue-cli-plugin-cypress. this way users can install the plugin using `vue add cypress`, more simple than `@vue/cypress`

PR #6066, opened 2020-11-18
fix(cli-service): pass --public host to devserver

### LINKED ISSUE #3220: Set allowedHosts option when serving with --public option
background:

1. @vue/cli-service has depd "webpack-dev-server": "'^3.1.10".
2. any new created project will use webpack-dev-server to 3.1.14 (current latest match ^3.1.10)
3. webpack-dev-server 3.1.14 added a new option: [allowedHosts](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L97)

Please add the `allowedHosts: ${hostname of public url}` option to webpack-dev-server when user start vue dev server with parameter --public

Or set `disableHostCheck: true` (not recommended)

otherwise HMR will fails

### Impacts

All projects created by `vue create` and using `--public` parameter starts from now on

jonaskuske:
close #3220

Prevent "Invalid Host" error by passing public host through to webpack-dev-server when `--public` is used

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

When you use the `public` option in webpack-dev-server [it'll automatically allow requests from the specified host](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L926-L935), without needing the additional `allowedHosts[]` or `disableHostCheck` options. However when the `--public` option in Vue CLI was used, the value was not passed through to webpack-dev-server (as HMR client injection is handled separately) so you still got the `Invalid host` error in that case.

This PR extracts the host from the `--public` arg and passes it through to the dev server to prevent this error.

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -112,6 +112,7 @@ module.exports = (api, options) => {
         ? rawPublicUrl
         : `${protocol}://${rawPublicUrl}`
       : null
+    const publicHost = publicUrl ? /^[a-zA-Z]+:\/\/([^/?#]+)/.exec(publicUrl)[1] : undefined
 
     const urls = prepareURLs(
       protocol,
@@ -181,6 +182,7 @@ module.exports = (api, options) => {
       hot: !isProduction,
       injectClient: false,
       compress: isProduction,
+      public: publicHost,
       publicPath: options.publicPath,
       overlay: isProduction // TODO disable this
         ? false
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -182,7 +182,6 @@ module.exports = (api, options) => {
       hot: !isProduction,
       injectClient: false,
       compress: isProduction,
-      public: publicHost,
       publicPath: options.publicPath,
       overlay: isProduction // TODO disable this
         ? false
@@ -190,6 +189,7 @@ module.exports = (api, options) => {
     }, projectDevServerOptions, {
       https: useHttps,
       proxy: proxySettings,
+      public: publicHost,
       // eslint-disable-next-line no-shadow
       before (app, server) {
         // launch editor support.


jonaskuske:
Done as suggested, `--public` overrides `devServer.public` now. :)

___

However, note that `devServer.public` isn't 100% equivalent to `--public` atm.
Even though [the DevServer docs unfortunately don't really mention this](https://webpack.js.org/configuration/dev-server/#devserverpublic), `devServer.public` needs to be a host, not a URL – [it'll strip the port when it determines the hostname to allow, but not the protocol etc.](https://github.com/webpack/webpack-dev-server/blob/master/lib/Server.js#L928-L930)

At the same time, `--public` is documented as "specify the public network **URL** for the HMR client" – which makes sense, as Vue CLI also uses it as the page to `--open`, so the protocol might be necessary if you want Vue CLI to correctly open the https version in your browser. (DevServer accepts an additional option called `openPage` for this purpose, but of course Vue CLI force-disables webpack's built-in open functionality entirely)

So, long story short:

1. `--public http://myhost.test` will work because Vue CLI accepts URLs and will determine the host, but `devServer.public: 'http://myhost.test'` will still give you an `Invalid Host` error

2. Vue CLI does not allow you to configure the HMR network URL separately from the URL you want to open on launch, e.g. if you run your site through `my-host.test` but your server doesn't handle websockets so you still want the HMR client to connect directly to `127.0.0.1:8080` – is this worth opening an issue?

fangbinwei:
> --public `http://myhost.test` will work because Vue CLI accepts URLs and will determine the host, but devServer.public: `'http://myhost.test'` will still give you an Invalid Host error

Your codes fix this problem, so we can handle `devServer.public: 'http://myhost.test'` . Since the priority of `publicHost` is higher than `devServer.public` now.

> When you use the public option in webpack-dev-server it'll automatically allow requests from the specified host, without needing the additional allowedHosts[] or disableHostCheck options

Your codes make CLI behave the same as webpack-dev-server in this respect.

> Vue CLI does not allow you to configure the HMR network URL separately from the URL you want to open on launch,

I think it's a very corner case.

Moreover, there may be some historical reasons that I don’t know make CLI and webpack-dev-server behave inconsistently

jonaskuske:
> Your codes make CLI behave the same as webpack-dev-server in this respect.

Ah, you're right of course, somehow totally missed that my code that extracts the host will be run both for `--public` _and_ the dev server option. 😅

Well, guess this PR is ready then?

fangbinwei (approved):
Thanks!

PR #6079, opened 2020-11-22
chore(cli): return with an exit code of 1 on supplying an unknown command

jamesgeorge007:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [x] Other, please describe: Supplying an unknown command returns an exit code of 1.

- [ ] Yes
- [x] No

**Other information:**

N/A

### DIFF
--- a/packages/@vue/cli/__tests__/args.spec.js
+++ b/packages/@vue/cli/__tests__/args.spec.js
@@ -3,9 +3,12 @@ const execa = require('execa')
 
 const CLI_PATH = path.resolve(__dirname, '..', 'bin', 'vue.js')
 
-const runAsync = async args => await execa(CLI_PATH, args)
+const runAsync = (args, options) => execa(CLI_PATH, args, options)
 
 test('suggests matching command', async () => {
-  const { stdout } = await runAsync(['confgi'])
+  const { code, stdout } = await runAsync(['confgi'], { reject: false })
+
+  // Assertions
+  expect(code).toBe(1)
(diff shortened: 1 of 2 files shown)


packages/@vue/cli/__tests__/__e2e__/args.spec.js
```diff
@@ -0,0 +1,14 @@
...
+
+const runAsync = async (args, options) => execa(CLI_PATH, args, options)
+
+test('suggests matching command', async () => {
+  const { code, stdout } = await runAsync(['confgi'], { reject: false })
+
+  // Assertions
+  expect(code).toBe(1)
+  expect(stdout).toContain('Did you mean config?')
+})
```
fangbinwei, line 14: Why create a new file?

jamesgeorge007: It is always better to have an abstraction layer between `e2e` and unit tests.

fangbinwei: I agree. If so, it's better to refactor all project tests folder. Only create a `__e2e` for this `args.spec.js` will make directory structure more complicated. I think consistency first.

jamesgeorge007: In that case, let's keep the directory structure as it is for now.

fangbinwei (approved):

PR #6134, opened 2020-12-09
fix: eslint formatter path error

githoniel:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

ESlint 7 has move code formatter to `eslint/lib/cli-engine/formatters/codeframe` but we have not change the path

### DIFF
--- a/packages/@vue/cli-plugin-eslint/index.js
+++ b/packages/@vue/cli-plugin-eslint/index.js
@@ -52,7 +52,7 @@ module.exports = (api, options) => {
           resolveModule('eslint/package.json', cwd) ||
             resolveModule('eslint/package.json', __dirname)
         ),
-        formatter: loadModule('eslint/lib/formatters/codeframe', cwd, true)
+        formatter: loadModule('eslint/lib/cli-engine/formatters/codeframe', cwd, true)
       }
       webpackConfig.plugin('eslint').use(eslintWebpackPlugin, [eslintWebpackPluginOptions])
     })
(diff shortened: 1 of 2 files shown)


packages/@vue/cli-plugin-eslint/index.js
```diff
@@ -52,7 +52,7 @@ module.exports = (api, options) => {
           resolveModule('eslint/package.json', cwd) ||
             resolveModule('eslint/package.json', __dirname)
         ),
-        formatter: loadModule('eslint/lib/formatters/codeframe', cwd, true)
+        formatter: loadModule('eslint/lib/cli-engine/formatters/codeframe', cwd, true)
```
fangbinwei, line 55:
```suggestion
        formatter: 'codeframe'
```

Since we have specified the path of `eslint` via `eslintPath`,  using [`eslint.loadFormatter('codeframe')`](https://github.com/webpack-contrib/eslint-webpack-plugin/blob/2e6c8eb62d50c3bdca7a6bd33b4acefacdccf402/src/linter.js#L238) can get the correct formatter, and more robust. What do you think about it?

fangbinwei (commented):
It seems that eslint@6 has already changed the directory structure. After using eslint@6, `loadModule('eslint/lib/cli-engine/formatters/codeframe', cwd, true)` always return `undefined`, then default value `stylish` is used

packages/@vue/cli-plugin-eslint/index.js
githoniel: I think `codeframe` is more nature as eslint offical document provide this. using function it look like we are using custom formatter

fangbinwei:
I added a test for formatter.

fangbinwei (approved):
Thanks a lot!

PR #6163, opened 2020-12-22
refactor: faster levenshtein

ka-weihe:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ x] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x ] No

**Other information:**
[leven](https://github.com/sindresorhus/leven) is one of the slowest levenshtein packages out there and the author is trying to deceive people with shitty benchmarks where he is only comparing it slower packages. Please consider using  [fastest-levenshtein]( https://github.com/ka-weihe/fastest-levenshtein) instead which is 10x faster and much higher quality.


sodatea:
Thanks for the PR. But due to the same reason that Babel doesn't swap out `levenary` https://github.com/babel/babel/pull/11911#issuecomment-668727746, I don't think Vue CLI needs a faster Levenshtein implementation either.

ka-weihe:
Babel swapped out levenary though?

sodatea:
They implemented the algorithm themselves.
We can do this in Vue CLI too someday. After all, we don't need a fast implementation, but a concise and easy-to-maintain one. `leven` does its job well.

PR #6199, opened 2021-01-09
Fix documentation on default linted files

Robin-Hoodie:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [X] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

The documentation mentions all files in the `src` and `tests` directories are linted by default. Upon inspection of the [sourcecode](https://github.com/vuejs/vue-cli/blob/3eaef4d388c52ea1d698991e7e96497164f90a04/packages/%40vue/cli-plugin-eslint/lint.js#L55) however, it seems that also Javascript files in the root directory are being linted by default.

I stumbled upon this when trying to figure out why my `.eslintrc.js` file itself was being linted by running `vue-cli-service lint` without passing any args.

### DIFF
--- a/docs/core-plugins/eslint.md
+++ b/docs/core-plugins/eslint.md
@@ -17,7 +17,7 @@
     --max-warnings       specify number of warnings to make build failed (default: Infinity)
   ```
 
-  Lints and fixes files. If no specific files are given, it lints all files in `src` and `tests`.
+  Lints and fixes files. If no specific files are given, it lints all files in `src` and `tests`, as well as all Javascript files in the root directory (these are most often config files such as `babel.config.js` or `.eslintrc.js`).
 
   Other [ESLint CLI options](https://eslint.org/docs/user-guide/command-line-interface#options) are also supported.
 
--- a/docs/core-plugins/eslint.md
+++ b/docs/core-plugins/eslint.md
@@ -17,9 +17,14 @@
     --max-warnings       specify number of warnings to make build failed (default: Infinity)
   ```
 
-  Lints and fixes files. If no specific files are given, it lints all files in `src` and `tests`, as well as all Javascript files in the root directory (these are most often config files such as `babel.config.js` or `.eslintrc.js`).
+Lints and fixes files. If no specific files are given, it lints all files in `src` and `tests`.
 
-  Other [ESLint CLI options](https://eslint.org/docs/user-guide/command-line-interface#options) are also supported.
+Other [ESLint CLI options](https://eslint.org/docs/user-guide/command-line-interface#options) are also supported.
+  
+::: tip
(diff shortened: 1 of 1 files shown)


fangbinwei (commented):
`.*.js` like `.eslintrc.js` will be linted only if  there is no `.eslintignore` in project according to https://github.com/vuejs/vue-cli/blob/3eaef4d388c52ea1d698991e7e96497164f90a04/packages/%40vue/cli-plugin-eslint/lint.js#L39 .Maybe it's not worth mentioning ?

> If `.eslintignore` exists and we want to lint `.eslintrc.js`,  `!.*.js` should be added to `.eslintignore`

Robin-Hoodie:
I wasn't aware dot-files were not being linted by default by ESLint. I also wasn't aware that the default behaviour of `vue-cli-service lint` is to override that default behavior of ESLint.

Perhaps it's best to mention both behaviours then?

Something along the lines of:

"`vue-cli-service lint` will by default lint hidden dot-files if there is no `.eslintignore` file or `ignorePatterns` property in your ESLint config. This behaviour is in contrast to ESLint's default behaviour, which will not lint dot-files by default. If you want to follow ESLint's default behaviour instead, consider adding an (empty) `.eslintignore` file or any `ignorePatterns` property to your ESLint config file"

docs/core-plugins/eslint.md
```diff
@@ -17,7 +17,7 @@
     --max-warnings       specify number of warnings to make build failed (default: Infinity)
   ```
 
-  Lints and fixes files. If no specific files are given, it lints all files in `src` and `tests`.
+  Lints and fixes files. If no specific files are given, it lints all files in `src` and `tests`, as well as all Javascript files in the root directory (these are most often config files such as `babel.config.js` or `.eslintrc.js`).
```
fangbinwei, line 20: Javascript => JavaScript

fangbinwei (changes requested):
I think a simple tip is enough. We prefer not to describe the default behavior of ESlint, which may change.  [ESlint@7](https://eslint.org/docs/user-guide/migrating-to-7.0.0#default-ignore-patterns-have-changed) no longer ignores `.eslintrc.js`, 
```
:::tip
`vue-cli-service lint` will lint dotfiles `.*.js` by default. If you want to follow ESLint's default behavior instead, consider adding a `.eslintignore` file in your project.
:::
```

docs/core-plugins/eslint.md
fangbinwei: The sentence you added is good, why did you delete it? 🤣  @Robin-Hoodie

Robin-Hoodie: My bad, I misinterpreted https://github.com/vuejs/vue-cli/pull/6199#pullrequestreview-566003817 to mean that my original sentence should also be removed

fangbinwei (approved):
Thanks a lot

PR #6220, opened 2021-01-20
feat: transpile dependencies in node_modules

sodatea:
TODOs:
- [x] ~~A simplified `dependencies` babel preset for `node_modules`, to reduce the performance overhead~~ I don't think there will be much overhead because the default preset only adds very few custom features. It would also be against the users' intention if they choose a different preset for their project
- [x] ~~Decide whether to opt-in or opt-out of this feature~~ Should be opt-out. And we should continue improving the cache configuration to speed up incremental builds.
- [ ] Test cases
- [ ] Revise the `transpileDependencies` documentation
- [ ] Mention this feature in the migration documentation

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/docs/guide/cli-service.md
+++ b/docs/guide/cli-service.md
@@ -52,6 +52,7 @@ Options:
   --https        use https (default: false)
   --public       specify the public network URL for the HMR client
   --skip-plugins comma-separated list of plugin names to skip for this run
+  --no-transpile-all: don't transpile all dependencies with Babel; only those specified by 'transpileDependencies' will be processed
 ```
 
 ::: tip --copy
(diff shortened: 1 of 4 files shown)


sodatea:
🤔 I think it might make more sense to reuse the existing `transpileDependencies` option than to add a new command-line flag.

I'll work on a new PR to just accept `transpileDependencies: true` as an option to transpile all dependencies.

PR #6257, opened 2021-02-02
fix resolve lib public path

ppossanzini:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

when i'm building a library and i've correctly configured vue.config.js publicPath, the resolver must use configured path instead of '' for two reasons: 
1) it's not good to change a user configuration (it's responsability of the user to correctly se publicPath according to his needs)
2) i'm developing a big application using microfrontend development approach and i need to address specific module separation also in folder structure because my application is composed at runtime of different umd modules. so is not correct to put all modules in / root folder or in a global javascript folder of main app. if you reset publicPath to '' i'm forced to put all modules in root folder and i cannot address module separation or module versioning by folder structure that is important during microfrontend development and ci/cd process. 

in the picture you can see an example of a deployment configuration i cannot address with publicPath resetted to '' for libs. 

image

### DIFF
--- a/packages/@vue/cli-service/lib/commands/build/resolveLibConfig.js
+++ b/packages/@vue/cli-service/lib/commands/build/resolveLibConfig.js
@@ -123,10 +123,7 @@ module.exports = (api, { entry, name, formats, filename, 'inline-vue': inlineVue
     }, rawConfig.output, {
       filename: `${entryName}.js`,
       chunkFilename: `${entryName}.[name].js`,
-      // use dynamic publicPath so this can be deployed anywhere
-      // the actual path will be determined at runtime by checking
-      // document.currentScript.src.
-      publicPath: ''
+
     })
 
     return rawConfig


fangbinwei (commented):
> The development branch is dev. And there are several version branches for archiving old versions of Vue CLI, such as v2, v3.
Pull requests that touches the code should be forked from dev, unless it's only targeting an old version.

You should create PR from `dev` or `v4` branch

> it's not good to change a user configuration

I agree with you.  We can enhance it

> if you reset publicPath to '' i'm forced to put all modules in root folder 

I don't think `publicPaht: '' ` is exactly the same as the root folder. CLI set publicPath to '', and overwrite `__webpack_public_path__`  at runtime by `setPublicPath` https://github.com/vuejs/vue-cli/blob/b41ed76c5ce54c618587f24b9d14c35cf31a96d4/packages/%40vue/cli-service/lib/commands/build/entry-lib.js#L1

Sorry I'm not quite sure what problems you'll run into if you don't set the `publicPath`  by yourself.

ppossanzini:
I fangbinwei, thanks for your comment, i'll request the PR from dev. 
I have a group of microfronend applications, each one build as lib in a separate build process. 
My CI/CD scripts place every build in a different server path, so i've change the publicPath in vue.config of every lib to match my publish path.
The main app load libs and use it correctly loading every lib at runtime via a json config from the correct subpath, but if a lib build has produced an additional chunk it fails if PRODUCTION env, because of publicPah has been resetted and load of libs chunks is from main app path and not from lib path;  if i build in DEVELOPMENT mode it works, because in DEVELOPMENT env the cli does not reset the publicPath of lib . 

i'll repost this comment in the next pull request

PR #6259, opened 2021-02-03
fix publicPath reset if build in PRODUCTION mode

ppossanzini:

- [X] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [X] No

**Other information:**

I have a group of microfronend applications, each one build as lib in a separate build process. 
My CI/CD scripts place every build in a different server path, so i've change the publicPath in vue.config of each lib to match my publish path.
The main app load libs and use it correctly loading every lib at runtime via a json config from the correct subpath, but if a lib build has produced an additional chunk it fails if PRODUCTION env, because of publicPath has been resetted to '' and load of libs chunks happens from main app path and not from lib path;  if i build in DEVELOPMENT mode it works, because in DEVELOPMENT env the cli does not reset the publicPath of lib .

### DIFF
--- a/packages/@vue/cli-service/lib/commands/build/resolveLibConfig.js
+++ b/packages/@vue/cli-service/lib/commands/build/resolveLibConfig.js
@@ -122,8 +122,7 @@ module.exports = (api, { entry, name, formats, filename, 'inline-vue': inlineVue
       globalObject: `(typeof self !== 'undefined' ? self : this)`
     }, rawConfig.output, {
       filename: `${entryName}.js`,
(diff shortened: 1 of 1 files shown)


nurlansu (commented):
Linking the previous PR: https://github.com/vuejs/vue-cli/pull/6257.

@ppossanzini you could simply change the target branch and rebase from `dev`, instead of creating a new PR.

PR #6323, opened 2021-03-01
feat(wip): add filesystem cache config for webpack 5

sodatea:
New:
- Turn on `filesystem` cache by default in webpack 5
- An `addBuildDependencies` plugin API for plugins to add additional
dependencies to the cache config

To Fix:
- Need to implement a `--no-cache` option for both `serve` and `build`
command
- Need to decide how to deal with `.js` config files when a random
environment variable change may affect its exports
- `lintOnSave` option needs an overhaul to work better with webpack 5
and the new cache mechanism
- Not sure how to test the cache implementation

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

Due to the may-be-buggy cache configuration, it should be considered as a breaking change.

- [x] Yes
- [ ] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-service/lib/PluginAPI.js
+++ b/packages/@vue/cli-service/lib/PluginAPI.js
@@ -212,6 +212,28 @@ class PluginAPI {
     const cacheIdentifier = hash(variables)
     return { cacheDirectory, cacheIdentifier }
   }
+
+  /**
+   * Vue CLI enables the filesystem cache in webpack 5 by default.
+   * If your plugin depends on an additional config file (e.g. `stylelint.config.js`),
+   * the default cache config does not take that file into account,
+   * thus causing unexpected cache persistence.
+   * To avoid such problems, you should add that file to the `buildDependencies`
+   * of webpack by calling this function.
+   * (e.g. `api.addBuildDependencies('stylelint.config.js')`)
+   * @param {string[]|Object} deps if an array of strings is passed, it will be
+   * appended to the cache.buildDependencies.config field; if an object is passed,
+   * it will be merged into the cache.buildDependencies filed
+   */
+  addBuildDependencies (deps, { shouldEvaluate }) {
+    if (Array.isArray(deps)) {
+      this.configureWebpack(() => ({ cache: { buildDependencies: { config: deps } } }))
+    } else {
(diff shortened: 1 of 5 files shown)


fangbinwei:
> lintOnSave option needs an overhaul to work better with webpack 5 and the new cache mechanism

Some information about `lintOnSave` needs to be taken into account

When we set `lintOnSave: true`, CLI will pass `emitWarning: true, emitError: false` to eslint-webpack-plugin.

In `eslint-webpack-plugin@2.4.3`, it will emit all eslint error as compilation warning. 

But after `2.5.0`https://github.com/webpack-contrib/eslint-webpack-plugin/pull/46 , it will only emit eslint warning as compilation warning, and doesn't report eslint error. We couldn't see eslint rule error message in console totally, the behavior of `emitError` and `emitWarning` is just like its name.

And I believe we can use the default value `true` of both `emitError` and `emitWarning`, and using `failOnError`, `failOnWarning` instead according to  https://github.com/webpack-contrib/eslint-webpack-plugin/pull/72  https://github.com/webpack-contrib/eslint-webpack-plugin/pull/85

The version of eslint-webpack-plugin is pinned to 2.4.3 in https://github.com/vuejs/vue-cli/pull/6291, Is there more information about this?

sodatea:
> And I believe we can use the default value `true` of both `emitError` and `emitWarning`, and using `failOnError`, `failOnWarning` instead according to [webpack-contrib/eslint-webpack-plugin#72](https://github.com/webpack-contrib/eslint-webpack-plugin/pull/72) [webpack-contrib/eslint-webpack-plugin#85](https://github.com/webpack-contrib/eslint-webpack-plugin/pull/85)

Yeah, I think so.

> The version of eslint-webpack-plugin is pinned to 2.4.3 in #6291, Is there more information about this?

Somehow the newest version won't re-lint when I run `yarn serve` the second time in my test project. Haven't figure out the reason yet. But I guess it's related to this PR: https://github.com/webpack-contrib/eslint-webpack-plugin/pull/67

yoyo837:
Any update on this?

PR #6347, opened 2021-03-12
refactor!: move vue-jest and ts-jest to peer dependencies

sodatea:
vue-jest now requires several different peer deps for vue 2 and vue 3;
ts-jest now requires typescript as peer dep, which should be optional
for unit-jest plugin users because they might choose not to use
typescript.

Both are causing annoying peer dep warnings.

So it only makes sense to move them to the project level, to keep things
clean.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [x] Yes
- [ ] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-plugin-unit-jest/migrator/index.js
+++ b/packages/@vue/cli-plugin-unit-jest/migrator/index.js
@@ -1,0 +1,24 @@
+/** @param {import('@vue/cli/lib/MigratorAPI')} api MigratorAPI */
+module.exports = (api) => {
+  api.extendPackage(pkg => {
+    const toMerge = {}
+
+    const allDeps = {
+      ...pkg.dependencies,
+      ...pkg.devDependencies,
+      ...pkg.optionalDependencies
+    }
+
+    if (!allDeps['vue-jest']) {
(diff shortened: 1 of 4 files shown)


packages/@vue/cli-plugin-unit-jest/migrator/index.js
```diff
@@ -0,0 +1,24 @@
...
+      // Likely a Vue 2 project, and uses the builtin preset.
+      // Because we used to add `vue-jest` v5 to dev deps for Vue 3 projects.
+      toMerge['vue-jest'] = '^4.0.1'
+    }
+
+    if (allDeps['@vue/cli-plugin-typescript'] && !allDeps['ts-jest']) {
+      toMerge['ts-jest'] = '^26.5.3'
+    }
+
+    return toMerge
```
PiDelport, line 22: Likely bug: These lines end up adding values to the root of `package.json` itself, instead of under the `dependencies` / `devDependencies` keys?

sodatea: Oops, thanks for noting this!

PR #6358, opened 2021-03-17
fix: work around npm6/postcss8 hoisting issue

sodatea:
Closes #6342

**What kind of change does this PR introduce?** (check at least one)

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

**Does this PR introduce a breaking change?** (check one)

- [ ] Yes
- [x] No

**Other information:**

PR #6372, opened 2021-03-24
fix: check hoisted postcss version

sodatea:
Reverts vuejs/vue-cli#6358


packages/@vue/cli-service/lib/config/css.js
```diff
@@ -58,6 +59,19 @@ module.exports = (api, rootOptions) => {
...
+      const autoprefixerDirectory = path.dirname(require.resolve('autoprefixer/package.json'))
+      const postcssPkg = loadModule('postcss/package.json', autoprefixerDirectory)
+      const postcssVersion = postcssPkg.version
+      if (!semver.satisfies(postcssVersion, '8.x')) {
+        throw new Error(
+          `The package manager has hoisted a wrong version of ${chalk.cyan('postcss')}, ` +
+          `please run ${chalk.cyan('npm i postcss@8 -D')} to fix it.`
+        )
+      }
+
```
fangbinwei, line 74: What if `hasPostCSSConfig` is `true`?

sodatea: That means the user is not likely using the built-in `autoprefixer` plugin.
In that case, they should take care of the peer dependency requirements of their own installed plugins.

sodatea: Projects created by Vue CLI < 4.1.0 may have a PostCSS configuration (which only adds `autoprefixer`) but no direct dependency installed. That's a bad practice. We can implement a migrator to help remove these configurations.

Issue #6251, opened 2021-01-31
Add the ability to reorder plugin execution


hl037:
### What problem does this feature solve?
Currently, plugins are executed in the order of the package.json dependency list. Since these dependencies are automatically alphabetically sorted by `yarn install`, One plugin can suddenly break when renamed if it expect some other plugin to be executed before ( see https://stackoverflow.com/questions/65977135/how-to-change-chainwebpack-call-order-with-vue-js )
...This would also bring some determism to when each plugin is called

### What does the proposed API look like?
An API could for example bring some kind of dependency graph across the plugins, like this :

First, the main export would be either a callback as it is currently, either an object.
if it is an object, it would look like this :
```
{
  before? : array<string>,
  after? : array<string>,
  name? : string,
  callback : (api, options) => void),
  metaPlugin? : boolean
}
```
`metaPlugin` defaults to False. I tell if callback will contains only calls to api.addPlugin() (defined below). If it is set, before, after and name should be left undefined/null/empty.

`callback` is equivalent to the current exported function.

If metaPlugin is not set :

`name`, if set is the name of group this plugin is part of. if not set, it defaults to the package name.

`before` and `after` are the requirement lists of the plugin in the dependency graph. (other plugin's name)

Then an `addPlugin()`
```
api.addPlugin({
  before? : array<string>,
  after? : array<string>,
  name? : string,
  callback : (api, options) => void),
  metaPlugin? : boolean
})
```
The fields have the same semantic as the exported object.

The plugin would be executed like this :

First, gather all (meta)plugin. Execute all meta-plugins "recursively" (a meta plugin can add another meta-plugin in its callback).

Then, solve the dep-graph. If not possible, raise an error.

Example : 

```js
const dts = require('dts-bundle');
const path = require('path');

module.exports = {
  metaPlugin : true
  callback = (api, options) => {
    api.addPlugin({
      after : ["ts-loader"],
      name : "ts-bundler",
      callback : (api, options) => {

        api.chainWebpack(config => {
          //Disable thread-loader, cache-loader
          const tsRule = config.module.rule('ts').test(/\.ts$/);
          const tsxRule = config.module.rule('tsx').test(/\.tsx$/);

          debugger
          tsRule.uses.delete('cache-loader');
          tsxRule.uses.delete('cache-loader');
          tsRule.uses.delete('thread-loader');
          tsxRule.uses.delete('thread-loader');

          //Enable the generating of declaration files 
          tsRule
[...]

Issue #5106, opened 2020-01-24
App.vue cannot be overwritten or customized using vue-cli presets


tbl0605:
### Version
4.1.2

### Reproduction link
[https://github.com/tbl0605/vue-cli-plugin-preset-bug-demo](https://github.com/tbl0605/vue-cli-plugin-preset-bug-demo)

### Environment info
```
  System:
    OS: Windows 10 10.0.18363
    CPU: (8) x64 Intel(R) Core(TM) i7-8550U CPU @ 1.80GHz
  Binaries:
    Node: 12.14.1 - E:\tools\nodejs\node.EXE
    Yarn: 1.16.0 - C:\Program Files (x86)\Yarn\bin\yarn.CMD
    npm: 6.13.4 - E:\tools\nodejs\npm.CMD   
  Browsers:
    Edge: 44.18362.449.0
  npmPackages:
    @fortawesome/vue-fontawesome: * => 0.1.9 
    @vue/babel-helper-vue-jsx-merge-props:  1.0.0 
    @vue/babel-plugin-transform-vue-jsx:  1.1.2 
    @vue/babel-preset-app:  4.1.2 
    @vue/babel-preset-jsx:  1.1.2 
    @vue/babel-sugar-functional-vue:  1.1.2 
    @vue/babel-sugar-inject-h:  1.1.2 
    @vue/babel-sugar-v-model:  1.1.2 
    @vue/babel-sugar-v-on:  1.1.2
    @vue/cli-overlay:  4.1.2
    @vue/cli-plugin-babel: ^4.1.0 => 4.1.2
    @vue/cli-plugin-eslint: ^4.1.0 => 4.1.2
    @vue/cli-plugin-router: ^4.1.0 => 4.1.2
    @vue/cli-plugin-vuex: ^4.1.0 => 4.1.2
    @vue/cli-service: ^4.1.0 => 4.1.2
    @vue/cli-shared-utils:  4.1.2
    @vue/component-compiler-utils:  3.1.1
    @vue/eslint-config-prettier: ^5.0.0 => 5.1.0
    @vue/preload-webpack-plugin:  1.1.1
    @vue/web-component-wrapper:  1.2.0
    eslint-plugin-vue: ^5.2.3 => 5.2.3
    vue: ^2.6.10 => 2.6.11
    vue-cli-webpack:  1.0.0
    vue-eslint-parser:  undefined (5.0.0)
    vue-hot-reload-api:  2.3.4
[...]

PR #6411, opened 2021-04-09
feat: implement plugin execution order

fangbinwei:
- [x] test
- [ ] doc

If there is nothing wrong with the code and design, I can try to add documentation

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

By default, the plugins are executed in the same order as before I believe

**Other information:**

related to #5106, #6251

### interface
```ts
interface Plugin {
  (): any
  after?: string|Array<string>
}
```

### Example

**vue-cli-plugin-foo/generator/index.js**
```js
module.exports = api => {
  api.render('./template')
}
```

**vue-cli-plugin-bar/generator/index.js**
```js
module.exports = api => {
  api.render('./template')
}
module.exports.after = 'vue-cli-plugin-bar'
```

 the 'after' can be used to fine-tune the execution order

### Implement

Topological Sort Algorithm sorts the plugins by 'after'. see [leetcode 210](https://leetcode.com/problems/course-schedule-ii/)

### DIFF
--- a/packages/@vue/cli-plugin-typescript/__tests__/tsGenerator.spec.js
+++ b/packages/@vue/cli-plugin-typescript/__tests__/tsGenerator.spec.js
@@ -65,19 +65,34 @@ test('use with Babel', async () => {
 })
 
 test('use with router', async () => {
+  const tsApply = require('../generator')
+
+  expect(tsApply.after).toBe('@vue/cli-plugin-router')
+
   const { files } = await generateWithPlugin([
+    {
+      id: '@vue/cli-service',
+      apply: require('@vue/cli-service/generator'),
+      options: {
+        plugins: {
+          '@vue/cli-service': {},
+          '@vue/cli-plugin-router': {},
+          '@vue/cli-plugin-typescript': {}
+        }
+      }
+    },
+    {
+      id: '@vue/cli-plugin-typescript',
+      apply: tsApply,
+      options: {}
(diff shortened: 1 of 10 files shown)


sodatea:
I like this idea! `after` will be very useful.

But I'm not so sure about `stage`. To actually use it we need to export quite a lot of stage constants and then the internal workflow of Vue CLI becomes somehow public. I'm not sure it's maintainable.

fangbinwei:
`stage`  is just to ensure that the generator of  `@vue/cli-service` is  invoked firstly, and it's a substitution  of `sortObject`.  Even without 'stage', 'after' can work with the current code.  Or we can just mention 'after' in the documentation?

sodatea:
Yeah, I think it can be kept an implementation detail until we figure out a better way to use it.

In the current state, I feel it's like `z-index` in CSS, and `z-index` is very hard to manage unless there's a design system to put some constraint on it…

sodatea (approved):
The implementation looks great to me.
Let's ship it first and add documentation in another PR later.

leookun:
hey，wonderful realization！
But I can't seem to put some plugin ahead of built-in:commands/serve, built-in:commands/serve is the first

PR #6426, opened 2021-04-19
code-style-update

iysf:

- [ ] Bugfix
- [ ] Feature
- [x] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/packages/@vue/cli-plugin-router/generator/index.js
+++ b/packages/@vue/cli-plugin-router/generator/index.js
@@ -19,7 +19,7 @@ module.exports = (api, options = {}, rootOptions = {}) => {
       }
     })
   }
-  
+
   api.render(`./template${isVue3 ? '-vue3' : ''}`, {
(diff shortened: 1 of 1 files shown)


sodatea:
`template-vue3` needs to be rendered **after** the normal templates, because it does not include all the necessary files, only those that differ from the Vue 2 template.

iysf:
> `template-vue3` needs to be rendered **after** the normal templates, because it does not include all the necessary files, only those that differ from the Vue 2 template.

Yes, I don't think there's any problem with your logic

But I want you to read the code carefully and think about the difference between the two

Compared with the previous function, there is no change, but the code will become more concise.

sodatea:
> Compared with the previous function, there is no change

The new version doesn't call `api.render('./template')` when `isVue3` is `true`.

PR #6428, opened 2021-04-21
feat(plugin-typescript): add all recommended tsconfig

IndexXuan:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
there are 11 [recommended tsconfig options](https://www.typescriptlang.org/tsconfig#forceConsistentCasingInFileNames), we should add the missing one **forceConsistentCasingInFileNames**.
Also **importHelpers** is useless if target `esnext` and if **importHelpers** is true, we should add **tslib** in deps(not impl it in this PR).


sodatea (approved):
Thanks!

PR #6603, opened 2021-07-25
Add text-align center to the template CSS

JIBSIL:

- [x] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [x] Other, please describe: Quality improvement: originally a bug, I think.

- [ ] Yes
- [x] No

**Other information:**
Not really a breaking change, just a quality update. After creating a project and adding the router plugin, anything you code after that will have the text-align: center CSS trait. This patch removes the `text-align: center trait` from the #app and shifts it to the #nav section, making following code useful and keeping the current navbar functionality.


JIBSIL:
I'm not super familiar with the regex used for substitution: if anyone could help me out on that, it would be nice.

PR #6630, opened 2021-08-10
Update installation.md to warn user about using CLI v5 with Vue v3

adube:
Vue CLI 4.x.x does not create Vue 3 compatible demos.  I had to install Vue CLI 5 (next) to make it work.

Having a notice about this in the documentation would help future users the same mistake I did.

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [x] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

### DIFF
--- a/docs/guide/installation.md
+++ b/docs/guide/installation.md
@@ -5,6 +5,10 @@ The package name changed from `vue-cli` to `@vue/cli`.
 If you have the previous `vue-cli` (1.x or 2.x) package installed globally, you need to uninstall it first with `npm uninstall vue-cli -g` or `yarn global remove vue-cli`.
 :::
 
+::: warning Warning regarding Next Version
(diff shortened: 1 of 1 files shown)


tony19:
I disagree with this docs update. 😄  I think a more helpful solution is to backport [the fix from 5.x](https://github.com/vuejs/vue-cli/commit/20355435e7b0ca59c1584d249cfad1abbac396c8) to 4.x.

adube:
Hey there @tony19 !  Seeing you "here" as well made me smile :)

> I think a more helpful solution is to backport the fix from 5.x to 4.x.

That's a good idea.

My original intent was: "I fell in a trap, wasted a few hours trying to figure out what I was doing wrong and would be happy if no one else fell in it too".

I'll close this PR.  Thanks again.

PR #6634, opened 2021-08-13
Remove deprecated cache-loader

nVitius:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
The cache-loader library is deprecated. It still runs on webpack 5, but it has a peerDependency on webpack 4 which breaks the installation of vue-cli with npm@7 (npm installs peerDependencies by default now).

This PR removes the dependency on cache-loader and adds webpack's new filesystem cache instead.


nVitius:
I started testing this locally today and haven't run into any issues.

nVitius:
Not sure why the `core` and `plugins` tests are failing. It looks like some regex isn't matching on the built files. I don't think anything I changed would have affected that though.

The `cli-ui` and `e2e` tests are failing because [vue-cli-plugin-apollo](https://github.com/Akryum/vue-cli-plugin-apollo) uses `cache-loader`. I'll open an issue over there, but it doesn't look like the repo is being actively maintained.

nVitius:
@sodatea 
Pinging you as you're the OP of the CLI v5 issue.

Can I get your feedback on this PR?

packages/@vue/cli-service/lib/config/base.js
```diff
@@ -18,6 +18,9 @@ module.exports = (api, options) => {
         .rule('esm')
           .test(/\.m?jsx?$/)
           .resolve.set('fullySpecified', false)
+
+      webpackConfig
+        .cache({ type: 'filesystem' })
```
sodatea, line 23: Don't do this. It's not safe. I'm pretty sure it'll fail at least one of these test cases:
https://github.com/sodatea/vue-cli/blob/e3cfcf57a30bb58ccc7895fe1d3242fcef797501/packages/%40vue/cli-service/__tests__/cache.spec.js#L32-L61

sodatea (changes requested):

packages/@vue/cli-service/lib/config/base.js
nVitius: Thanks for the feedback! I appreciate it.

It's not explicitly stated in the documentation, but the spec for caching in webpack 5 states: 
> Each build with a different webpack configuration should store a unique copy of its cache versus another webpack configuration. E.g. A development build and a production build must have distinct caches due to them having different options set.

The spec for it is here: https://github.com/webpack/webpack/issues/6527

I'm not sure how the `--mode` option works in vue-cli, but it does look like those specific tests did not fail for this PR. You can see the reports for those here: https://app.circleci.com/pipelines/github/vuejs/vue-cli/1124/workflows/10aeafe8-27a2-486a-9609-2111481209be/jobs/27240

The only two tests that failed do not seem to be related to the changes I made.

sodatea: The link points to another PR that I'm working on.
The tests are not written yet, therefore it's not in the pipeline.

sodatea: > The only two tests that failed do not seem to be related to the changes I made.

I'm afraid they're related. Looks like the build process for modern mode reused the cache from the legacy mode.

719media:
Is this issue going to be solved prior to v5 final release? Given that it's currently in RC status (v5.0.0-rc.1), I hope it is something that gets resolved prior to release.

Icacls:
any update on this issiue?

yoyo837:
Any update on this issue? Again.

bodograumann:
vue-cli is now in maintenance mode, so don't hold your breath.

PR #6744, opened 2021-10-12
chore: Introduce serve warning to prevent insecure misuse

lloydjatkinson:

- [ ] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [x] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

This PR introduces a warning for the `vue-cli-service serve` command, specifically below the output that says `App running at:`

The reasoning behind this is that I've personally encountered many setups where instead of an actual production web server like Apache, Nginx, etc being used they have setup a host and simply run Vue CLI app on the host and have it running via `pm2` or another insecure and hacky solution. After discussing the setup it became clear that it was a misunderstanding of Vue CLI's purpose, and that the docs don't have any remarks on it. Afterall, Vue CLI simply uses Webpacks dev server.

This topic often regularly crops up on the Vue Discord and people need to keep repeating warnings of why this is not a secure or viable setup and why it won't scale like a real web server. Additionally, not running an accepted web server can be enough to fail an independent security assessment of developers applications/infrastructure.

Vue CLI does not warn users to not use Vue CLI in production. Other framework's CLI tools do warn the user, such as Angular's CLI. So to solve this I suggest a warning to guide users to a better approach. I hope the wording and colours are OK but of course I don't mind, the real goal here is to help users.

Here is what it looks like:

image

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -286,6 +286,12 @@ module.exports = (api, options) => {
         console.log()
         console.log(`  App running at:`)
         console.log(`  - Local:   ${chalk.cyan(urls.localUrlForTerminal)} ${copied}`)
+        
+        console.log(chalk.red.bold(`  This is a simple server for use in testing or debugging Vue CLI applications locally.`));
+        console.log(chalk.white.bgRed.bold(`  Don't use this in production!`));
+        console.log();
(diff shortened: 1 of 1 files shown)


lloydjatkinson:
ping.

lloydjatkinson:
Hi @sodatea can you look into this please?

lloydjatkinson:
Assuming this will never be implemented.

PR #6752, opened 2021-10-14
Add a top-level `terser` option to allow users to customize the minifier

Viggo95:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [x] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

#### background

- Form [`terser-webpack-plugin@5.2`](https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/CHANGELOG.md#520-2021-08-31) we can custom minify
- So Vue CLI 5 can support user to custom terser-webpack-plugin minify as `esbuild`, `swc` or `uglifyJs` 👇

<br>

#### Some data
``` shell
// delete .cache before each build
yarn build --no-module
```

##### Vue CLI 5 demo

- Generally, projects with more files will be more effective
- I just tested the speed briefly,  we have used it in the company’s project, and the build time can be reduced by about **20%**

> no cache,  minify with default options

<br>
<br>
<br>

### DIFF
--- a/packages/@vue/cli-service/lib/config/terserOptions.js
+++ b/packages/@vue/cli-service/lib/config/terserOptions.js
@@ -96,10 +96,10 @@ const uglifyJsMinify = (options) => ({
 
 // Currently we do not allow custom minify function
 const getMinify = (options) => {
-  const { minify = 'default' } = options.terser || {}
+  const { minify = 'terser' } = options.terser || {}
 
   const minifyMap = {
-    default: terserMinify,
+    terser: terserMinify,
     esbuild: esbuildMinify,
     swc: swcMinify,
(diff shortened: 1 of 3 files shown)


.prettierrc
```diff
@@ -0,0 +1,8 @@
+{
```
sodatea, line 1: Please remove this file.
The Vue CLI codebase was not formatted by prettier. Introducing prettier in an irrelevant PR could bring in unrelated diffs.

It is also the cause of the failing TypeScript test.

screetBloom: PR updated

PR #7031, opened 2022-03-07
feat: add Playwright for end-to-end testing

mxschmitt:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Hello, 👋 
[Playwright](https://playwright.dev/) is getting more and more traction over the last years and we should provide users an easy integration with the Vue ecosystem so users benefit from it.
This Pull Request adds Playwright as an end-to-end testing solution like it's done for Cypress. It might make sense to throw an error if the user selects both variants for e2e testing.

The CI is broken because Playwright requires Ubuntu as a linux distribution, probably migrating to github actions might be the easiest.

Also in this PR I'm not using a custom command, I tried it but it feels like it makes things more complex, since when the user would do: `vue-cli-service test:e2e --help` would lead to vue-cli-server's help text and not forward it to Playwright CLI. Since Playwright has running a webserver integrated, it should be fine, wdyt?

Thank you

cc @sodatea


haoqunjiang:
Hi Max, thanks for the PR! Playwright is a really cool project and I'm excited to see it becoming more and more popular.

But as stated in the project README, Vue CLI is in maintenance mode now.
We don't have the capacity to review, add, and keep maintaining such a big new feature.

So rather than integrating the plugin into the `vue-cli` repo, I think it would be better for the plugin to be shipped as a standalone 3rd-party Vue CLI plugin. A separate repository would be much more maintainable.

I'll dedicate some time to review the corresponding PR in `create-vue` later though.

PR #7076, opened 2022-03-25
feat: Refactor the ipc module.

ygj6:

- [ ] Bugfix
- [x] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
* ~~The dependencies of @achrinza/node-ipc has security risks.~~
  * ~~The owner of  [js-message](https://www.npmjs.com/package/js-message) and [easy-stack](https://www.npmjs.com/package/easy-stack) remains `Brandon Nozaki Miller`.~~
  * ~~`@achrinza/node-ipc` does not accept new features.~~
* Rewrite the iPc module to eliminate dependencies.
* More Robust Custom Communication Protocols.
* Remove emitevent to reduce performance overhead.

### DIFF
--- a/packages/@vue/cli-shared-utils/lib/env.js
+++ b/packages/@vue/cli-shared-utils/lib/env.js
@@ -15,7 +15,7 @@ const _gitProjects = new LRU({
   max: 10,
   maxAge: 1000
 })
-const CRLF = '\r\n'
+const DELIMITER = '\f'
 
 // env detection
 exports.hasYarn = () => {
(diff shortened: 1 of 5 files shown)


ygj6:
Local is passed. let me see.:sweat_smile:

haoqunjiang:
> * The owner of  [js-message](https://www.npmjs.com/package/js-message) and [easy-stack](https://www.npmjs.com/package/easy-stack) remains `Brandon Nozaki Miller`.

Their versions are fixed, npm does not allow overwriting existing versions, so no risk.

> * `@achrinza/node-ipc` does not accept new features.

We don't need new features, either. As long as there are no vulnerabilities we are fine.

RedNuttyGuy:
Hi,

Is there an estimate on when this change will be released?

`@achrinza/node-ipc` recently updated fixing an issue with non-lts versions of `node` and `yarn` caused by the `engines.node` config.

If it's going to be a while away I can create a PR to update the dependency.

haoqunjiang (changes requested):
My main concern with this PR is that it makes CLI UI not able to communicate with projects running older versions of `@vue/cli-service` via IPC.

Is there anyway to keep that compatibility?

ygj6:
To be compatible with older versions of @vue/cli-service, we can use the same `pipe path` as node-ipc. I'll verify it further.

ygj6:
> Is there anyway to keep that compatibility?

It can be compatible with older versions after using the same ipc path and message format. I have tested it on the `dashboard` page.

PR #7102, opened 2022-04-08
fix: the match fails when the compile dependency option is regular

yzydeveloper:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**
When the `transpileDependencies` option is regular, it will be written like this according to the developer's convention: `/@scope\/external-dep/` . However, the matching fails due to the windows style path, and must be `/@scope\\external-dep/` to be matched. This commit fixes this bug and escapes the regex again.


cgaof (commented):
ok

PR #7202, opened 2022-06-19
fix: optimize the judgment on whether HTTPS has been set in options

backrunner:

- [x] Bugfix
- [ ] Feature
- [ ] Code style update
- [ ] Refactor
- [ ] Docs
- [ ] Underlying tools
- [ ] Other, please describe:

- [ ] Yes
- [x] No

**Other information:**

Webpack also provides two configuration items named `http2` and `devServer.server`, if `http2` is set to `true`, the dev server will also start with HTTPS, and it's the same for `devServer.server` is set to `https` or `spdy`.

The current code only judgement whether `projectDevServerOptions.https` is a truthy value, this is an inaccurate judgement.

This patch contains changes that cover more cases than the existing code, allowing users to get correct URLs after executing `vue-cli-service serve`.

### DIFF
--- a/packages/@vue/cli-service/lib/commands/serve.js
+++ b/packages/@vue/cli-service/lib/commands/serve.js
@@ -103,8 +103,8 @@ module.exports = (api, options) => {
     // resolve server options
     const modesUseHttps = ['https', 'http2']
     const serversUseHttps = ['https', 'spdy']
-    const optionsUseHttps = modesUseHttps.some(modeName => !!projectDevServerOptions[modeName])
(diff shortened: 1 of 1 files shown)


haoqunjiang (approved):
Thanks!

backrunner:
@sodatea Your commit [558dea2](https://github.com/vuejs/vue-cli/commit/558dea2) suddenly reminded me that my fix is not perfect. `devServer.server.type` is not included in the judgement, I'll create another patch to cover that case.
If you have time now, you can complete this judgement first. Sorry for this fix isn't perfect.

Issue #6594, opened 2021-07-19
Wrong url for safari-nomodule-fix if publicPath is absolute url (V5)


7iomka:
### Version
5.0.0-beta.2

### Environment info
```
  System:
    OS: macOS 10.15.7
    CPU: (6) x64 Intel(R) Core(TM) i5-8500B CPU @ 3.00GHz
  Binaries:
    Node: 12.16.3 - /usr/local/bin/node
    Yarn: 1.17.3 - ~/.npm-global/bin/yarn
    npm: 6.14.8 - ~/.npm-global/bin/npm
  Browsers:
    Chrome: 92.0.4515.93
    Edge: Not Found
    Firefox: 90.0
    Safari: 14.1
  npmPackages:
    @vue/babel-helper-vue-jsx-merge-props:  1.2.1 
    @vue/babel-helper-vue-transform-on:  1.0.2 
    @vue/babel-plugin-jsx:  1.0.6 
    @vue/babel-plugin-transform-vue-jsx:  1.2.1 
    @vue/babel-preset-app:  5.0.0-beta.2 
    @vue/babel-preset-jsx:  1.2.4 
    @vue/babel-sugar-composition-api-inject-h:  1.2.1 
    @vue/babel-sugar-composition-api-render-instance:  1.2.4 
    @vue/babel-sugar-functional-vue:  1.2.2 
    @vue/babel-sugar-inject-h:  1.2.2 
    @vue/babel-sugar-v-model:  1.2.3 
    @vue/babel-sugar-v-on:  1.2.3 
    @vue/cli-overlay:  5.0.0-beta.2 
    @vue/cli-plugin-babel: ~5.0.0-beta.2 => 5.0.0-beta.2 
    @vue/cli-plugin-eslint: ^5.0.0-beta.2 => 5.0.0-beta.2 
    @vue/cli-plugin-router: ^5.0.0-beta.2 => 5.0.0-beta.2 
    @vue/cli-plugin-vuex: ^5.0.0-beta.2 => 5.0.0-beta.2 
    @vue/cli-service: ~5.0.0-beta.2 => 5.0.0-beta.2 
    @vue/cli-shared-utils:  5.0.0-beta.2 
    @vue/compiler-core:  3.1.5 
    @vue/compiler-dom:  3.1.5 
    @vue/compiler-sfc:  3.1.5 
    @vue/compiler-ssr:  3.1.5 
    @vue/component-compiler-utils:  3.2.2 
[...]

PR #7232, opened 2022-07-12
fix: SafariNomoduleFixPlugin.js fullSafariFixPath path error

Kyoloro:

- [x] Bugfix

- [ ] Yes
- [x] No

**Other information:**
issues reply:
#6594


Kyoloro:
#6594

yilihjy:
In your fix
If publicPath is set to `/relative`
The end result will be
`//relative/js/safari-nomodule-fix.js`
Causes a browser request
`http://relative/js/safari-nomodule-fix.js`
So I think it would be a better way to use "://" to judge publicPath, such as https://github.com/vuejs/vue-cli/pull/7247

